PR #2, opened 2020-01-06
Test use @pingcap-incubator/pd-client-js lib

baurine:
- install and use @pingcap-incubator/pd-client-js from GitHub Packages
- update build scripts and related docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,10 @@ swagger_spec:
 
 yarn_dependencies:
 ifneq ($(SKIP_YARN_INSTALL), 1)
+	# Skip post-install scripts by `--ignore-scripts` here,
+	# as a malicious script could steal NODE_AUTH_TOKEN.
 	cd ui &&\
-	yarn install --frozen-lockfile
+	yarn install --ignore-scripts
 endif
 
 swagger_client: swagger_spec yarn_dependencies
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,8 @@ DASHBOARD_API ?=
 
 BUILD_TAGS ?=
 
+SKIP_YARN_INSTALL ?=
+
 ifeq ($(SWAGGER),1)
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -20,10 +20,10 @@ swagger:
 	scripts/generate_swagger.sh
 
 swagger_client: swagger
-	cd ui && yarn && npm run build_api_client
+	cd ui && yarn install --update-checksums && npm run build_api_client
```
breeswish: Do we need `--frozen-lockfile` instead?

ui/src/index.js
```diff
@@ -3,4 +3,12 @@ import ReactDOM from 'react-dom';
 import './index.css';
 import App from './App';
 
+import PDClient from '@pingcap-incubator/pd-client-js';
+
+// TODO
+const pdClient = new PDClient({
```
breeswish: You can put it in `src/utils/client/index.js`, change it to expose two APIs:

```js
export default {
  dashboard: dashboardClient,
  pd: pdClient,
};
```

Makefile
baurine: Nope, adding `--update-checksums` is used to fix ["Integrity checked failed error"](https://github.com/yarnpkg/yarn/issues/7552) for yarn

ui/src/index.js
baurine: Ok, will fix it soon.

Makefile
baurine: It seems `--frozen-lockfile` option works fine as well.

breeswish:
Nice. Looks like you need to update TiDB Dashboard's GitHub action file in order to make it work.

baurine:
Yep, will fix it tmw, it seems we need to add a more step to create a `.npmrc` file for authentication.

baurine:
Fix the github actions, now it works fine.

PR #3, opened 2020-01-07
backend: add `/logs` API for searching logs

Deardrops:
This PR add a new API `/api/logs` for Dashboard backend.
For all available APIs, please run 
```
SWAGGER=1 make server && ./bin/tidb-dashboard
```
and visit API documents generated by swagger:
http://127.0.0.1:12333/dashboard/api/swagger/index.html

Changes:
- [x] Add [gorm](gorm.io/) package
- [x] Add comments for swagger


pkg/apiserver/apiserver.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/gin-contrib/cors"
 	"github.com/gin-gonic/gin"
 	"github.com/pingcap-incubator/tidb-dashboard/pkg/apiserver/foo"
+	"github.com/pingcap-incubator/tidb-dashboard/pkg/apiserver/logs"
```
baurine: It seems the logs.go file is missed.

Deardrops: yes... thanks for mention

lonng:
Do you mean the client should fetch logs of different components or instances individually? and sort them by time in the client?

breeswish:
@lonng This is a POC implementation. I have sent you a formal design.

.gitignore
```diff
@@ -54,3 +54,5 @@ fabric.properties
 
 # Android studio 3.1+ serialized cache file
 .idea/caches/build_file_checksums.ser
+
+sqlite.db
```
breeswish: Prefer to call it `dashboard.sqlite.db` so that it can be recognized easily when integrated into PD.

pkg/apiserver/logs/handler.go
```diff
@@ -0,0 +1,161 @@
...
+	err := controller.stopTaskGroup(taskGroupID, false)
+	if err != nil {
+		c.String(http.StatusBadRequest, err.Error())
+		return
+	}
+	c.String(http.StatusOK, "")
+}
+
+func (s *Service) cancelTaskHandler(c *gin.Context) {
+	taskID := c.Query("taskID")
```
breeswish: Prefer to follow RESTful APIs.

pkg/apiserver/logs/handler.go
```diff
@@ -0,0 +1,161 @@
...
+	taskGroupID := controller.AddTaskGroup(args)
+	err := controller.RunTaskGroup(taskGroupID)
+	if err != nil {
+		c.String(http.StatusInternalServerError, err.Error())
+		return
+	}
+	c.String(http.StatusOK, taskGroupID)
+}
+
+func (s *Service) listAllTasksHandler(c *gin.Context) {
```
breeswish: Please make sure to add necessary comments according to https://github.com/swaggo/swag so that all APIs are available in our Swagger UI (so that it will be also available in our JavaScript side).

pkg/apiserver/logs/controller.go
```diff
@@ -0,0 +1,184 @@
...
+	"io"
+	"net/http"
+	"os"
+	"path"
+)
+
+var controller *Controller
+
+func init() {
+	controller = NewController("./sqlite.db")
```
breeswish: - we may need this sqlite instance across multiple controllers. please move it to global. (maybe consider introducing some standard MVC framework?)
- prefer to put it behind our `config.DataDir`

pkg/apiserver/logs/handler.go
```diff
@@ -0,0 +1,161 @@
...
+
+type Service struct {
+	config *config.Config
+}
+
+func NewService(config *config.Config) *Service {
+	return &Service{config: config}
+}
+
+func (s *Service) Register(r *gin.RouterGroup) {
```
mapleFU: Would you mind following the Rest Api? Using "GET" everywhere is confused...

pkg/apiserver/logs/db.go
```diff
@@ -0,0 +1,157 @@
...
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logs
+
```
mapleFU: Would you mind change SQL keyword to upper case?

pkg/apiserver/logsearch/minheap.go
```diff
@@ -0,0 +1,102 @@
+package logsearch
```
wjhuang2016: License？

Deardrops: I forgot it, thanks~

scripts/generate_swagger_spec.sh
```diff
@@ -19,3 +19,5 @@ go install github.com/swaggo/swag/cmd/swag
 
 echo "+ Generate swagger spec"
 swag init -g cmd/tidb-dashboard/main.go
+go mod vendor
+swag init --parseVendor -g cmd/tidb-dashboard/main.go
```
Deardrops: In issue swaggo/swag#567, We cannot parse struct correctly in external packages without `--parseVendor` flag.
So we change the commands here:
1. Swag generates `docs/docs.go` file, but without the JSON field related to external packages.
2. Enable `go vendor` for this package
3. Swag regenerates `docs/docs.go` file.

breeswish: Can we remove line 21?

pkg/apiserver/logsearch/models.go
```diff
@@ -0,0 +1,141 @@
...
+	gorm.Model `json:"-"`
+	TaskID     string                    `json:"task_id"`
+	Message    *diagnosticspb.LogMessage `json:"message" gorm:"embedded"`
+}
+
+type DBClient struct {
+	db *store.DB
+}
+
+var dbClient DBClient
```
breeswish: prefer not to use a global db client.

pkg/apiserver/logsearch/models.go
```diff
@@ -0,0 +1,141 @@
...
+// limitations under the License.
+
+package logsearch
+
+import (
+	"database/sql/driver"
+	"encoding/json"
+
+	"github.com/jinzhu/gorm"
+	_ "github.com/jinzhu/gorm/dialects/sqlite"
```
breeswish: this can be removed now

pkg/apiserver/logsearch/models.go
```diff
@@ -0,0 +1,141 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logsearch
+
+import (
+	"database/sql/driver"
```
breeswish: maybe not useful anymore?

pkg/apiserver/logsearch/models.go
```diff
@@ -0,0 +1,141 @@
...
+	d.db.Create(&preview)
+}
+
+func (d *DBClient) cleanPreview(taskID string) error {
+	return d.db.Delete(PreviewModel{}, "task_id = ?", taskID).Error
+}
+
+func (d *DBClient) cleanTask(taskID string) error {
+	return d.db.Delete(TaskModel{}, "task_id = ?", taskID).Error
+}
```
breeswish: I prefer to simply call inner methods directly, i.e. without introducing this abstraction, since the inner methods are already readable and simple enough.

pkg/apiserver/logsearch/error.go
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logsearch
+
+import "github.com/gin-gonic/gin"
+
+// NewError example
+func NewError(ctx *gin.Context, status int, err error) {
```
breeswish: Maybe better to put it outside `logsearch` so that other services can use it as well.

pkg/apiserver/apiserver.go
```diff
@@ -32,13 +34,16 @@ func Handler(apiPrefix string, config *config.Config) http.Handler {
 		gin.SetMode(gin.ReleaseMode)
 	})
 
+	db := store.NewDB(config)
```
breeswish: Forgot to merge master?

pkg/apiserver/logsearch/models.go
Deardrops: No it's used by [a function](https://github.com/pingcap-incubator/tidb-dashboard/pull/3/files#diff-9d74faa65ca1eeddd99f6ff22525a913R40).

scripts/generate_swagger_spec.sh
Deardrops: No, it will cause the 3rd step panic, this panic is a bug inner swaggo.

pkg/apiserver/logsearch/task.go
```diff
@@ -0,0 +1,222 @@
...
+		if err != nil {
+			t.Error = err.Error()
+			return
+		}
+	}
+}
+
+func toLine(msg *diagnosticspb.LogMessage) string {
+	timeStr := time.Unix(0, msg.Time*int64(time.Millisecond)).Format(sysutil.TimeStampLayout)
+	return fmt.Sprintf("[%s] [%s] %s\n", timeStr, diagnosticspb.LogLevel_name[int32(msg.Level)], msg.Message)
```
lonng: It's better to use `msg.Level.String()` to replace the `diagnosticspb.LogLevel_name[int32(msg.Level)]`.

pkg/apiserver/logsearch/router.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+	var components = []*Component{
+		{
+			ServerType: "tidb",
+			IP:         "127.0.0.1",
+			Port:       "4000",
+			StatusPort: "10080",
+		},
+	}
+	taskGroupID := scheduler.addTasks(components, searchLogReq)
```
lonng: Does a single search request contain multiple tasks?

Deardrops: Yes, a request will start a `taskGroup`, and it will start multiple tasks (one component per task).

pkg/apiserver/logsearch/router.go
```diff
@@ -0,0 +1,353 @@
...
+// @Summary Create a task group
+// @Description create and run a task group
+// @Produce json
+// @Success 200 {object} httputil.HTTPSuccess
+// @Failure 400 {object} httputil.HTTPError
+// @Failure 500 {object} httputil.HTTPError
+// @Router /logs/taskgroups [post]
+func (s *Service) TaskGroupCreate(c *gin.Context) {
+	// TODO: using parameters provided by client
+	endTime := time.Now().UnixNano() / int64(time.Millisecond)
```
breeswish: need to use actual request parameters now. Maybe you can simply accept a json object.

pkg/apiserver/logsearch/router.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+	var components = []*Component{
+		{
+			ServerType: "tidb",
+			IP:         "127.0.0.1",
+			Port:       "4000",
+			StatusPort: "10080",
+		},
+	}
+	if s.scheduler.taskGroup != nil && s.scheduler.taskGroup.State == StateRunning {
```
breeswish: why do we need to limit this? there might be multiple pages starting a search task.

pkg/apiserver/logsearch/router.go
```diff
@@ -0,0 +1,353 @@
...
+	endpoint.DELETE("/taskgroups/:id", s.TaskGroupDelete)
+}
+
+// @Summary List all tasks
+// @Description list all log search tasks
+// @Produce json
+// @Success 200 {array} TaskModel
+// @Failure 400 {object} httputil.HTTPError
+// @Router /logs/tasks [get]
+func (s *Service) TaskGetList(c *gin.Context) {
```
breeswish: Prefer to put verb first, like `ListTasks`.

pkg/apiserver/logsearch/router.go
```diff
@@ -0,0 +1,353 @@
...
+// @Description preview fetched logs in a task by providing task ID
+// @Produce json
+// @Param id path string true "task id"
+// @Success 200 {array} PreviewModel
+// @Failure 400 {object} httputil.HTTPError
+// @Router /logs/tasks/{id}/preview [get]
+func (s *Service) TaskPreview(c *gin.Context) {
+	taskID := c.Param("id")
+	var lines []PreviewModel
+	err := s.db.Where("task_id = ?", taskID).Find(&lines).Error
```
breeswish: Maybe need to add order by?

pkg/apiserver/logsearch/models.go
```diff
@@ -0,0 +1,117 @@
...
+	State   TaskState         `json:"state"`
+}
+
+func (TaskGroupModel) TableName() string {
+	return "log_search_task_groups"
+}
+
+type PreviewModel struct {
+	ID          uint                   `json:"-" grom:"primary_key"`
+	TaskID      string                 `json:"task_id" gorm:"index;type:char(36)"`
```
breeswish: how about using simple auto increment int?

pkg/apiserver/logsearch/router.go
```diff
@@ -0,0 +1,353 @@
...
+		return
+	}
+	_, err = io.Copy(c.Writer, f)
+	if err != nil {
+		httputil.NewError(c, http.StatusInternalServerError, err)
+		return
+	}
+}
+
+func (s *Service) queryTaskByID(taskID string) (task TaskModel, err error) {
```
breeswish: If you use int ID then no need for this function

pkg/apiserver/logsearch/router.go
```diff
@@ -0,0 +1,353 @@
...
+	if err != nil {
+		httputil.NewError(c, http.StatusBadRequest, err)
+		return
+	}
+	f, err := os.Open(task.SavedPath)
+	if err != nil {
+		httputil.NewError(c, http.StatusInternalServerError, err)
+		return
+	}
+	_, err = io.Copy(c.Writer, f)
```
breeswish: Prefer to follow https://github.com/gin-gonic/gin#serving-data-from-reader . Also we'd better specify content-type and file name so that user is more easy to download.

pkg/apiserver/logsearch/router.go
```diff
@@ -0,0 +1,353 @@
...
+	tasks := make([]*TaskModel, 0, len(ids))
+	for _, taskID := range ids {
+		task, err := s.queryTaskByID(taskID)
+		if err != nil {
+			httputil.NewError(c, http.StatusBadRequest, err)
+			return
+		}
+		tasks = append(tasks, &task)
+	}
+	err := dumpLogs(tasks, c.Writer)
```
breeswish: maybe call it packLogsAsTarball is better

pkg/apiserver/logsearch/router.go
```diff
@@ -0,0 +1,353 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, err = io.Copy(tw, f)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
breeswish: Prefer to put inside another file.

pkg/apiserver/logsearch/router.go
```diff
@@ -0,0 +1,353 @@
...
+
+// @Summary Retry failed tasks
+// @Description retry tasks that has been failed
+// @Produce json
+// @Success 200 {object} httputil.HTTPSuccess
+// @Failure 500 {object} httputil.HTTPError
+// @Router /logs/tasks/retry [post]
+func (s *Service) TaskRetry(c *gin.Context) {
+	err := s.scheduler.runTaskGroup(true)
```
breeswish: TaskGroupId should be specified in order to correctly work in multi-user env later.

pkg/apiserver/logsearch/models.go
Deardrops: It will increase code complexity

PR #12, opened 2020-01-16
refactor: replace semantic-ui-react components by antd

baurine:
Final result:

WX20200116-163843@2x

### DIFF
--- a/ui/src/apps/keyvis/RootComponent.less
+++ b/ui/src/apps/keyvis/RootComponent.less
@@ -1,3 +1,5 @@
+@import '~antd/es/style/themes/default.less';
+
 .PD-KeyVis {
   // font-family: 'Poppins';
   font-size: 14px;
@@ -44,13 +46,6 @@
 
     .space {
       width: 12px;
-    }
-
-    .select-icon {
-      border: 1px solid #ccc;
-      height: 32px;
-      width: 32px;
-      padding: 4px 8px;
     }
   }
 
@@ -151,7 +146,7 @@
   }
 }
 
-#brightness-overlay {
+#PD-KeyVis-Brightness-Overlay {
   background-color: white;
   padding: 8px;
   border: 1px solid #ccc;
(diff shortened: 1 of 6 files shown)


breeswish:
Good job! In addition, I think following ant-design's own style is better (since it is a comprehensive and consistent design language) and there is no need to reproduce semantic's style here. For example, maybe a standard ant-design drop down with an icon prefix is fine.

baurine:
Got it! will update it tmw.

baurine:
hi @breeswish , updated it by antd components, PTAL.

WX20200117-175756@2x

breeswish:
Cool!

ui/src/apps/keyvis/RootComponent.less
```diff
@@ -135,3 +150,16 @@
...
+  background-color: white;
+  padding: 8px;
+  border: 1px solid #ccc;
+  border-radius: 4px;
+
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+
+  box-shadow: 0 100px 80px rgba(0, 0, 0, 0.12);
```
breeswish: Prefer to use antd variables like `box-shadow: @box-shadow-base`:

https://github.com/ant-design/ant-design/blob/master/components/menu/style/index.less#L17

You can import these antd variables by adding `@import '~antd/es/style/themes/default.less';`.

ui/src/apps/keyvis/ToolBar.tsx
```diff
@@ -18,133 +17,147 @@ export interface IKeyVisToolBarProps {
...
+          icon="sync"
+          type={isAutoFetch ? 'primary' : 'default'}
+        >
+          自动刷新
+        </Button>
+
+        <div className="space" />
+
+        <Input.Group compact style={{ width: 142 }}>
+          <div className="select-icon">
```
breeswish: I prefer to avoid duplicating ant-design's style as much as possible. As a solution for this, how about: https://github.com/ant-design/ant-design/issues/5744#issuecomment-550399711

baurine:
Ok, thank you for your patient review! I will try it soon.

ui/src/apps/keyvis/ToolBar.tsx
baurine: Awesome, works great!

ui/src/apps/keyvis/RootComponent.less
baurine: Done!

baurine:
New version:

WX20200119-112302@2x

PR #15, opened 2020-02-05
Implement statement detail page

baurine:
- Add horizontal visual color bars for table columns
- Implement the statement detail components

WX20200205-191859@2x

WX20200205-191938@2x

PTAL

### DIFF
--- a/ui/.yarnrc
+++ b/ui/.yarnrc
@@ -1,1 +1,2 @@
 "@pingcap-incubator:registry" "https://npm.pkg.github.com"
+"@baurine:registry" "https://npm.pkg.github.com"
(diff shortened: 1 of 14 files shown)


baurine:
- Format the values by different units





hey @breeswish , I tried but didn't use `pretty-ms` and `pretty-bytes`, because they just support one kind unit, we need more, for example, convert 1000 to `1 K`, and `pretty-bytes` is 1000-base, but maybe we need 1024-base, so I still use the code I extracted from grafana, it can convert most of the common units (all you can see in the grafana diagrams), I made it as a lib and published it to GitHub Packages likes pd-client-js.

PR #24, opened 2020-02-09
keyvisual: migrate backend code from pd, update frontend

HunDunDM:
* migrate backend code from pd
* update front-end api
* fix missing color bug


cmd/tidb-dashboard/main.go
```diff
@@ -61,25 +62,40 @@ func NewCLIConfig() *DashboardCLIConfig {
 	}()
 
 	cfg := &DashboardCLIConfig{
-		CoreConfig: &config.Config{
-			Ctx:     ctx,
-			Version: utils.ReleaseVersion,
-		},
+		CoreConfig: config.NewConfig(ctx, utils.ReleaseVersion),
```
breeswish: Is this actually creating a default config?

HunDunDM: I want this effect, but it was not used later.

HunDunDM:
PTAL @breeswish

PR #25, opened 2020-02-09
Introduce storage layer

breeswish:
Extracted from https://github.com/pingcap-incubator/tidb-dashboard/pull/3


HunDunDM:
Can we put the store in CoreConfig?

breeswish:
@HunDunDM It looks really wired to put more and more config unrelated things into CoreConfig.. It should be really called something like context honestly :)

PR #26, opened 2020-02-10
Add lints

breeswish:
This PR adds lints to routine development as well as CI check.

You can build with lint using the same way as other TiDB projects:

```sh
make dev
```


.golangci.yml
```diff
@@ -0,0 +1,33 @@
+run:
+  sip-dirs:
```
HunDunDM: skip-dirs?

Makefile
```diff
@@ -24,8 +24,10 @@ LDFLAGS += -X "$(DASHBOARD_PKG)/pkg/utils.GitBranch=$(shell git rev-parse --abbr
 default:
 	SWAGGER=1 make server
 
-tidy:
-	go mod tidy
+lint:
```
HunDunDM: Need swagger_spec as a target?

breeswish: this may make it impossible to lint without swagger (swagger dependencies are always installed)?

HunDunDM: Sorry, I didn't see `docs/placeholder.go` before

PR #27, opened 2020-02-10
*: use retool to manage tools

rleungx:
This PR uses retool to manage tools


breeswish:
Actually go mod is more preferred over retools since no 3rd-party tool is needed 🤣 See https://stackoverflow.com/questions/56636580/replace-retool-with-tools-go-for-multi-developer-and-ci-environments-using-go-mo

rleungx:
> Actually go mod is more preferred over retools since no 3rd-party tool is needed 🤣 See https://stackoverflow.com/questions/56636580/replace-retool-with-tools-go-for-multi-developer-and-ci-environments-using-go-mo

Yes. But we may need some other tools which are not go-based in the future?

nolouch:
More like go mod to management tools.🤣which tool may not go-based but we need it?

PR #34, opened 2020-02-17
Statement API and integration

baurine:
Implement the statement backend API, use original SQL instead of ORM

Related API design docs: https://docs.google.com/document/d/1KjgbQh4Xv5X2_z_p3pL90ArkR1mESZdG4bgyRoEaku4/edit

Implement following APIs:

image

Almost done, remain works and some questions:

1. TiDB connection info should come from config instead of hard-code.
1. What's the good time to connect to TiDB (aka the time to run OpenTiDB()), what's the best way to handle the situation if connect to TiDB failed (now I assume it will always be success), and when should we close the TiDB connection?

Hi @breeswish , please help take a look, thank you!


baurine:
Connect the backend and frontend:

baurine:
Known issues:

1. some schemas name we get by `show databases` are uppercase, but in the statement table's schema_name column, they are stored by lowercase. so when we select an uppercase schema in the dropdown selector, we get empty matched records.
1. some schema_name column values are empty (I think they should be "mysql")

baurine:
Some refines:

- fix the issue that will get empty records when selecting uppercase schema
- fix the issue that will get empty records when selecting "mysql" schema
- sort the schemas names
- hide the cluster selector
- enable sort for most of all columns in StatementNodesTable
- update styles
- swap the time range selector and schema selector position
- rename some components
- others

BTW, I limited the "SQL Category" column width to 400px instead of ellipse its content, I think it is better to show the full content here. I also tried the monospace typography, but it needs to take more vertical space, so I rolled it back (but of course we can change this later) 

Final effects:



Overview:



Detail:



The effect when using monospace font:

企业微信20200219032743

baurine:
Support save the search options by context, it will keep the previous search options when jumping back to overview page from detail page.

baurine:
Integrate with auth, and support i18n:



@breeswish PTAL

PR #35, opened 2020-02-18
Add I18N support

breeswish:
This PR adds I18N support to Dashboard, as well as translation files for KeyVis.


ui/.github_release_version
```diff
@@ -1,3 +1,3 @@
 # This file contains a version number which will be used to release assets to
 # GitHub. To trigger a new asset release, simply increase this version number.
-20200214_1
+20200218_1
```
HunDunDM: Will we need a special process to modify this line in the future?

breeswish: Nop, it's a specially designed as a manual process so that release is only run on demand.

PR #50, opened 2020-02-19
pkg: add profiling section

rleungx:
This PR adds the profiling section to support `/api/profile`


pkg/apiserver/profile/handler.go
```diff
@@ -0,0 +1,248 @@
...
+// @Produce json
+// @Param tikv query string false "tikv"
+// @Param tidb query string false "tidb"
+// @Param pd query string false "pd"
+// @Success 200 {object} httputil.HTTPSuccess
+// @Router /profile/group/start [get]
+func (s *Service) startHandler(c *gin.Context) {
+	tg := NewTaskGroup()
+	addrs := make(map[string][]string)
+	addrs[tidb] = c.QueryArray(tidb)
```
breeswish: A better way is to:

1. `type StartRequest struct { ... }`
2. Use `c.ShouldBindJSON` to bind the request body to this handler.

pkg/apiserver/profile/handler.go
```diff
@@ -0,0 +1,248 @@
...
+// @Router /profile/group/cancel/{groupId} [post]
+func (s *Service) cancelGroupHandler(c *gin.Context) {
+	taskGroupID := c.Param("groupId")
+	taskGroup := TaskGroupModel{}
+	err := s.db.Where("id = ?", taskGroupID).First(&taskGroup).Error
+	if err != nil {
+		httputil.NewError(c, http.StatusBadRequest, err)
+		return
+	}
+	s.tasks.Range(func(key, value interface{}) bool {
```
breeswish: I see that s.tasks was never deleted? This could cause memory leaks.

pkg/apiserver/httputil/error.go
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package httputil
+
+import "github.com/gin-gonic/gin"
+
+// NewError example
+func NewError(ctx *gin.Context, status int, err error) {
```
breeswish: In latest master we have a better error framework now. You can remove it.

Errors can be passed by:

```
_ = c.Error(xxx)
return
```

pkg/apiserver/profile/model.go
```diff
@@ -0,0 +1,180 @@
...
+	Cancel TaskState = "cancel"
+
+	// TaskGroup can only have these two states.
+	Running TaskState = "running"
+	Finish  TaskState = "finish"
+)
+
+// TaskModel is the model definition of task.
+type TaskModel struct {
+	ID          string    `json:"task_id"`
```
breeswish: `gorm:"primary_key"`?

breeswish: Prefer uint key

pkg/apiserver/profile/model.go
```diff
@@ -0,0 +1,180 @@
...
+	FilePath    string    `json:"file_path" gorm:"type:text"`
+	Component   string    `json:"component"`
+	CreateTime  int64     `json:"create_time"`
+	FinishTime  int64     `json:"finish_time"`
+	Error       string    `json:"error" gorm:"type:text"`
+}
+
+// TaskGroupModel is the model definition of task group.
+type TaskGroupModel struct {
+	ID           string    `json:"task_group_id" `
```
breeswish: Prefer uint id

pkg/apiserver/profile/model.go
rleungx: Oh, I just keep consistent with the definition of the log part.

go.mod
```diff
@@ -9,7 +9,11 @@ require (
 	github.com/gin-contrib/gzip v0.0.1
 	github.com/gin-gonic/gin v1.5.0
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
+	github.com/go-openapi/spec v0.19.5 // indirect
```
breeswish: :) Do you need to clean up go mod?

pkg/apiserver/profile/handler.go
```diff
@@ -0,0 +1,315 @@
...
+)
+
+// Built-in component type
+const (
+	tikv = "tikv"
+	tidb = "tidb"
+	pd   = "pd"
+)
+
+// ProfilingRequest defines the
```
breeswish: defines the ...?

pkg/apiserver/profile/model.go
```diff
@@ -0,0 +1,180 @@
...
+
+// TrackTickInterval is the interval to check the tasks status.
+const TrackTickInterval = 2 * time.Second
+
+// TaskState is used to represent the task/task group state.
+type TaskState string
+
+// Built-in task state
+const (
+	Create TaskState = "create"
```
breeswish: TaskStateCreated

pkg/apiserver/profile/model.go
```diff
@@ -0,0 +1,180 @@
...
+	"go.uber.org/zap"
+
+	"github.com/pingcap-incubator/tidb-dashboard/pkg/dbstore"
+)
+
+// TrackTickInterval is the interval to check the tasks status.
+const TrackTickInterval = 2 * time.Second
+
+// TaskState is used to represent the task/task group state.
+type TaskState string
```
breeswish: type TaskState int

pkg/apiserver/profile/model.go
```diff
@@ -0,0 +1,180 @@
...
+	Create TaskState = "create"
+	Error  TaskState = "error"
+	Cancel TaskState = "cancel"
+
+	// TaskGroup can only have these two states.
+	Running TaskState = "running"
+	Finish  TaskState = "finish"
+)
+
+// TaskModel is the model definition of task.
```
breeswish: These templated comments can be just removed. Dashboard project does not require us to write these `Foo is Foo` comments :)

pkg/apiserver/profile/model.go
```diff
@@ -0,0 +1,180 @@
...
+type TaskModel struct {
+	ID          string    `json:"task_id" gorm:"type:char(36);primary_key"`
```
breeswish: You can take a look at latest log PR, and change to use int as well: https://github.com/pingcap-incubator/tidb-dashboard/blob/ff18b92cedd13d50be20a7a5c0528b701cd893a4/pkg/apiserver/logsearch/models.go#L73

pkg/apiserver/profile/model.go
```diff
@@ -0,0 +1,180 @@
...
+	TaskGroupID string    `json:"task_group_id" gorm:"type:char(36)"`
+	State       TaskState `json:"state"`
+	Addr        string    `json:"address"`
+	FilePath    string    `json:"file_path" gorm:"type:text"`
+	Component   string    `json:"component"`
+	CreateTime  int64     `json:"create_time"`
+	FinishTime  int64     `json:"finish_time"`
+	Error       string    `json:"error" gorm:"type:text"`
+}
```
breeswish: You need to assign table names, otherwise the table name is `TaskModel` which will be confusing (what task it is?). Reference: https://github.com/pingcap-incubator/tidb-dashboard/blob/ff18b92cedd13d50be20a7a5c0528b701cd893a4/pkg/apiserver/logsearch/models.go#L81

pkg/apiserver/profile/model.go
```diff
@@ -0,0 +1,180 @@
...
+	Running TaskState = "running"
+	Finish  TaskState = "finish"
+)
+
+// TaskModel is the model definition of task.
+type TaskModel struct {
+	ID          string    `json:"task_id" gorm:"type:char(36);primary_key"`
+	TaskGroupID string    `json:"task_group_id" gorm:"type:char(36)"`
+	State       TaskState `json:"state"`
+	Addr        string    `json:"address"`
```
breeswish: Do we need to store host and port in separate?

pkg/apiserver/profile/model.go
```diff
@@ -0,0 +1,180 @@
...
+	FilePath    string    `json:"file_path" gorm:"type:text"`
+	Component   string    `json:"component"`
```
breeswish: Is this a component name?

pkg/apiserver/profile/model.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// TaskGroupModel is the model definition of task group.
+type TaskGroupModel struct {
+	ID           string    `json:"task_group_id" gorm:"type:char(36);primary_key"`
+	RunningTasks int       `json:"running_tasks"`
+	State        TaskState `json:"state"`
+}
+
+func autoMigrate(db *dbstore.DB) {
```
breeswish: You can follow https://github.com/pingcap-incubator/tidb-dashboard/blob/ff18b92cedd13d50be20a7a5c0528b701cd893a4/pkg/apiserver/logsearch/models.go#L127

pkg/apiserver/profile/model.go
```diff
@@ -0,0 +1,180 @@
...
+	Addr        string    `json:"address"`
+	FilePath    string    `json:"file_path" gorm:"type:text"`
+	Component   string    `json:"component"`
+	CreateTime  int64     `json:"create_time"`
+	FinishTime  int64     `json:"finish_time"`
+	Error       string    `json:"error" gorm:"type:text"`
+}
+
+// TaskGroupModel is the model definition of task group.
+type TaskGroupModel struct {
```
breeswish: Table name is needed.

pkg/apiserver/profile/model.go
```diff
@@ -0,0 +1,180 @@
...
+	ID           string    `json:"task_group_id" gorm:"type:char(36);primary_key"`
+	RunningTasks int       `json:"running_tasks"`
```
breeswish: Are you sure want to store the number of running tasks..

pkg/apiserver/profile/profile.go
```diff
@@ -0,0 +1,105 @@
...
+	"os"
+	"os/exec"
+
+	"github.com/pkg/errors"
+)
+
+// the default time interval of profiling.
+const grabInterval = 30
+
+func fetchSvg(ctx context.Context, t, addr, filePrefix string) (string, error) {
```
breeswish: maybe better to call it `fetchFlameGraphSVG`.

pkg/apiserver/profile/router.go
```diff
@@ -0,0 +1,48 @@
...
+	tasks  sync.Map
+}
+
+// NewService creates a new service.
+func NewService(config *config.Config, db *dbstore.DB) *Service {
+	autoMigrate(db)
+	return &Service{config: config, db: db, tasks: sync.Map{}}
+}
+
+// Register register the handlers to the service.
```
breeswish: Looks like you can put router together with the handler

pkg/apiserver/profile/profile.go
rleungx: go based components are not the FlameGraph instead

pkg/apiserver/profile/router.go
rleungx: What's the benefit?

pkg/apiserver/profile/model.go
rleungx: What's the benefit?

pkg/apiserver/profile/profile.go
breeswish: oh. yes. It is a dot graph..

pkg/apiserver/profile/router.go
breeswish: So that route path is not maintained in two separated files..

pkg/apiserver/profile/model.go
rleungx: got

pkg/apiserver/profile/model.go
rleungx: Should I change to `componentType` or something else?

pkg/apiserver/profile/model.go
rleungx: for better tracing the running tasks

pkg/apiserver/profile/model.go
rleungx: got it

pkg/apiserver/profile/model.go
```diff
@@ -0,0 +1,121 @@
...
+			Component:   component,
+			CreateTime:  time.Now().Unix(),
+		},
+		db: db,
+	}
+}
+
+func (t *Task) run() {
+	ctx, cancel := context.WithCancel(context.Background())
+	t.cancel = cancel
```
breeswish: `ctx` and `cancel` need to be created and stored before run, otherwise `t.cancel` and `t.stop()` may have race condition (`t` is created, inserted into sync map, called `stop()` by other one before it `run()`)

pkg/apiserver/profile/router.go
```diff
@@ -0,0 +1,358 @@
...
+	if err != nil {
+		log.Fatal("Failed to initialize database", zap.Error(err))
+	}
+	return &Service{config: config, db: db, tasks: sync.Map{}}
+}
+
+// Register register the handlers to the service.
+func (s *Service) Register(r *gin.RouterGroup) {
+	endpoint := r.Group("/profile")
+	endpoint.GET("/group/start", s.startHandler)
```
breeswish: It should be `POST` since it changes things.

pkg/apiserver/profile/router.go
```diff
@@ -0,0 +1,358 @@
...
+// @Summary Create a task group
+// @Description create and run a task group
+// @Produce json
+// @Param pr query profile.ProfilingRequest true "profiling request"
+// @Success 200 {uint} uint "group ID"
+// @Failure 400 {object} utils.APIError
+// @Router /profile/group/start [post]
+func (s *Service) startHandler(c *gin.Context) {
+	var pr ProfilingRequest
+	if err := c.ShouldBind(&pr); err != nil {
```
breeswish: If you use POST, then this can be ShouldBindJSON and `form:"tidb"` in the field tag can also be removed.

pkg/apiserver/profile/router.go
```diff
@@ -0,0 +1,358 @@
...
+	for component, addrs := range addrs {
+		for _, addr := range addrs {
+			t := NewTask(s.db, taskGroup.ID, component, addr)
+			s.db.Create(t.TaskModel)
+			s.tasks.Store(t.ID, t)
+		}
+	}
+	go func() {
+		var wg sync.WaitGroup
+		s.tasks.Range(func(key, value interface{}) bool {
```
breeswish: how about storing tasks in an array in Line 98? Then you don't need to iterate and filter this sync map any more.

breeswish:
@all-contributors please add @rleungx  for code

PR #70, opened 2020-02-22
Support TiDB Authentication

breeswish:
Full I18N is also supported.

Sign in page looks like:

image

If there is no TiDB:
image

If incorrect password is provided:
image

Chinese:
image

After sign in success:
image

User will be redirected to sign in page if the token is expired:
image

Update:

Language select in sign in page:

image

Sign out:

image


ui/src/apps/keyvis/index.js
```diff
@@ -5,8 +5,5 @@ module.exports = {
   icon: 'eye',
   menuTitle: 'Key Visualizer', // TODO: I18N
   isDefaultRouter: true,
-  translations: {
-    en: require('./translations/en.yaml'),
-    zh_CN: require('./translations/zh_CN.yaml'),
-  },
+  translations: require.context('./translations/', false, /\.yaml$/),
```
baurine: Nice!

ui/src/components/LanguageDropdown/index.js
```diff
@@ -1,23 +1,26 @@
 import React from 'react';
-import { Menu, Icon, Dropdown } from 'antd';
+import { Menu, Dropdown } from 'antd';
 import _ from 'lodash';
 import { withTranslation } from 'react-i18next';
-import NavAction from './NavAction';
 
-class LangDropdown extends React.PureComponent {
+@withTranslation()
```
baurine: We can try the function component so we can use `useTranslation()` react hook instead of HOC, I used in https://github.com/pingcap-incubator/tidb-dashboard/commit/97c96ad10cecf342b33dc4d70a2f6a918e0a8ca5

ui/src/utils/i18n.js
```diff
@@ -4,36 +4,28 @@ import LanguageDetector from 'i18next-browser-languagedetector';
 import _ from 'lodash';
 
 const resources = {};
-const languages = ['en', 'zh_CN'];
 
-function loadResource(res) {
-  languages.forEach(lang => {
-    _.merge(resources, { [lang]: { translation: res[lang] } });
+export function loadResourceFromRequireContext(requireContext) {
```
baurine: I removed this function in https://github.com/pingcap-incubator/tidb-dashboard/commit/97c96ad10cecf342b33dc4d70a2f6a918e0a8ca5, replace d it by i18n.addResourceBundle() after init() to support i18n for left nav menus.

breeswish: Would you mind modifying this PR directly for the changes you propose? Since looks like there will be some conflicts.

baurine: Yep, let me update it tmw.

breeswish: I have helped you merge the branch, PTAL 

https://github.com/pingcap-incubator/tidb-dashboard/pull/70/commits/2472f2828f5005dd7617752564c8c53fa405d32f

PR #74, opened 2020-02-24
Simplify statement queries

breeswish:
This PR simplifies the statement query.


baurine:
Awesome, so clean and simplified.

PR #135, opened 2020-03-02
Add error table to make err happend in report's generating process imported to a table

reafans:
before: the error happend in generating report would stop the generating process.
this pr: these error would be imported to a table.

change a table's return err for test: 
image

### DIFF
--- a/pkg/apiserver/diagnose/report.go
+++ b/pkg/apiserver/diagnose/report.go
@@ -134,7 +134,7 @@ func GetReportTables(startTime, endTime string, db *gorm.DB) []*TableDef {
 		GetPDCurrentConfig,
 		GetTiKVCurrentConfig,
 	}
-	
+
 	tables := make([]*TableDef, 0, len(funcs))
 	var errRows []TableRowDef
 	for _, f := range funcs {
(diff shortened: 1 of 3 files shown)


pkg/apiserver/diagnose/report.go
```diff
@@ -60,10 +60,11 @@ const (
 	CategoryPD       = "PD"
 	CategoryTiKV     = "TiKV"
 	CategoryConfig   = "config"
+	CategoryError    = "err"
```
crazycs520: s/err/error

pkg/apiserver/diagnose/report.go
```diff
@@ -134,17 +135,35 @@ func GetReportTables(startTime, endTime string, db *gorm.DB) ([]*TableDef, []err
...
 	tables := make([]*TableDef, 0, len(funcs))
-	errs := make([]error, 0, len(funcs))
+	var errRows []TableRowDef
 	for _, f := range funcs {
 		tbl, err := f(startTime, endTime, db)
 		if err != nil {
-			errs = append(errs, err)
+			category := ""
+			if tbl.Category != nil {
+				category = tbl.Category[0]
```
crazycs520:
```go
category = strings.join(tbl.Category,",")
```

PR #144, opened 2020-03-03
profiling: support authentication of the profiling api 

rleungx:
This PR supports the authentication of the profiling API.

### DIFF
--- a/ui/src/apps/logSearching/components/SearchProgress.tsx
+++ b/ui/src/apps/logSearching/components/SearchProgress.tsx
@@ -1,46 +1,54 @@
-import client, { DASHBOARD_API_URL } from '@/utils/client';
-import { LogsearchSearchTarget, LogsearchTaskModel } from '@/utils/dashboard_client';
-import { Button, Card, Modal, Tree, Typography } from 'antd';
-import { AntTreeNodeCheckedEvent } from 'antd/lib/tree/Tree';
-import React, { useContext, useEffect, useRef, useState } from "react";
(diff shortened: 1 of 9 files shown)


breeswish:
This will break the download functionality. I think you can wait @Deardrops to give a solution for it.

breeswish:
You can update your PR according to https://github.com/pingcap-incubator/tidb-dashboard/pull/165 now :)

ui/src/apps/logSearching/components/SearchProgress.tsx
```diff
@@ -207,14 +200,14 @@ export default function SearchProgress({
...
-    
+
     const res = await client.dashboard.logsDownloadAcquireTokenGet(keys)
     const token = res.data
     if (!token) {
       return
     }
     const url = `${DASHBOARD_API_URL}/logs/download?token=${token}`
-    downloadFile(url)
+    window.open(url)
```
Deardrops: Open a new window may be intercepted by some browser, I suggest download file by creating a hidden `<a>` element and click it.

breeswish: If opening a new window is intercepted, how will this method work?

breeswish: How about this solution? https://stackoverflow.com/questions/19026162/javascript-window-open-from-callback Looks like this should be the correct answer.

Deardrops: OK for me.

PR #148, opened 2020-03-04
diagnosis: fix miss bracket syntax error

baurine:
This PR is to fix the browser reported "Uncaught SyntaxError: missing ) after argument list" error, this error makes the "expand" button doesn't work in the diagnosis report page.


baurine:
@breeswish @HunDunDM PTAL

PR #157, opened 2020-03-04
Profiling duration and progress UI

breeswish:
Profiling duration can be selected:

image

Profiling progress is supported:

image

Based on https://github.com/pingcap-incubator/tidb-dashboard/pull/127

### DIFF
--- a/pkg/apiserver/profiling/model.go
+++ b/pkg/apiserver/profiling/model.go
@@ -33,16 +33,23 @@ const (
 	TaskStateFinish
 )
 
+type NodeType string
+
+const (
+	NodeTypeTiKV NodeType = "tikv"
+	NodeTypeTiDB NodeType = "tidb"
+	NodeTypePD   NodeType = "pd"
+)
+
 type TaskModel struct {
 	ID          uint      `json:"id" gorm:"primary_key"`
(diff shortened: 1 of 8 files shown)


pkg/apiserver/profiling/model.go
```diff
@@ -33,25 +33,33 @@ const (
...
 	ID          uint      `json:"id" gorm:"primary_key"`
 	TaskGroupID uint      `json:"task_group_id" gorm:"index"`
 	State       TaskState `json:"state" gorm:"index"`
-	Addr        string    `json:"address"`
+	Addr        string    `json:"address" gorm:"size:32"`
+	TargetKind  NodeType  `json:"target_kind" gorm:"size:10"`
 	FilePath    string    `json:"file_path" gorm:"type:text"`
-	Component   string    `json:"component"`
-	CreatedAt   int64     `json:"created_at"`
-	FinishedAt  int64     `json:"finished_at"`
```
rleungx: Do we need to keep the finish time since history may use it?

breeswish: For history, maybe adding a CreatedAt for Task**Group** is better (since it can keep the same order even if we add retry to tasks later)?

PR #164, opened 2020-03-04
Sql diagnosis: make GetReportTables perform concurrently 

reafans:
cluster：
image

before：
image

this pr：
2 cores：
image

4 cores：
image

6 cores：
image

8 cores：
image

15 cores：
image

We can see that over 6 cores it would not get better performance. It due to some siggle table cost too much time that makes the whole time equal it’s time. So I set the cores not beyond 8 cores.

### DIFF
--- a/pkg/apiserver/diagnose/report.go
+++ b/pkg/apiserver/diagnose/report.go
@@ -243,15 +243,16 @@ func GetReportTables(startTime, endTime string, db *gorm.DB) []*TableDef {
 	// all task done, close the resChan
 	close(resChan)
 
-	var tblAndErrSlice []tblAndErr
+	tblAndErrSlice := make([]tblAndErr, 0, cap(resChan))
 	for tblAndErr := range resChan {
 		tblAndErrSlice = append(tblAndErrSlice, *tblAndErr)
 	}
 	sort.Slice(tblAndErrSlice, func(i, j int) bool {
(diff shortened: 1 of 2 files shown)


pkg/apiserver/diagnose/report.go
```diff
@@ -217,25 +220,91 @@ func GetReportTables(startTime, endTime string, db *gorm.DB) []*TableDef {
...
+		go doGetTable(taskChan, resChan, doneChan, startTime, endTime, db)
+	}
+
+	//block until all doGetTable done
+	for i := 0; i < conc; i++ {
+		<-doneChan
+	}
+	// all task done, close the resChan
+	close(resChan)
+	m := make(map[int]*tblAndErr)
```
crazycs520: remove the map use sort.slice instead.

pkg/apiserver/diagnose/report.go
```diff
@@ -217,25 +220,91 @@ func GetReportTables(startTime, endTime string, db *gorm.DB) []*TableDef {
...
+			}
 		}
 		if tbl != nil {
-			tables = append(tables, tbl)
+			tblAndErr.tbl = tbl
+		}
+		tblAndErr.taskID = task.taskID
+		resChan <- &tblAndErr
+		if tblAndErr.taskID == cap(taskChan)-1 {
+			close(taskChan)
```
crazycs520: you can close the channel after sender finish.

Issue #150, opened 2020-03-04
UCP: Support debug html templates without rebuild 


baurine:
## Description

The diagnosis report is rendered by the backend go template.

At first, we put the templates in the individual html files like `diagnosis/index.html`, it easy to debug, when we modify the html file, it just needs to refresh the browsers instead of rebuilding the whole go code, but this solution prevents us to release the project as an independent binary. (note: you need to set the gin in debug mode by commenting the code `gin.SetMode(gin.ReleaseMode)` in the `pkg/apiserver/apiserver.go`).

Then we move the templates' contents into go code as strings, you can see the code in `pkg/apiserver/diagnose/templates.go`, it helps us to release the package but not easy to debug the templates, every time we modify the template's contents, we need to rebuild and re-run the go code.

So please help us to resolve this problem, debug the template without rebuild go code, and still can release the repo as an independent binary.

Ref: https://tech.townsourced.com/post/embedding-static-files-in-go/

The diagnosis report page:

![d-0](https://user-images.githubusercontent.com/1284531/75842253-cf5ec580-5e0a-11ea-915d-4b336586be83.png)

## Score

* 200

## Mentor(s)

* @breeswish 
* @baurine 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Go

STRRL:
Here is what I do:

- move html content into new .html file: `index.html`, `table.html`
- create a customized `PreferLocalFileHTMLProduction` to replace `HTMLProduction`, It will reload content from local file if exists
- use `go-bindata-assetfs` generate resource .go file for production usage.
- use build tag to split debug and prod env (WIP)

My question is: Should I add the resource file generated by `go-bindata-assetfs`  into git repo?

If I do not do this, new developer need manually execute `go-bindata-assetfs` before start their work. It may cause some confuse.

baurine:
hey @STRRL , I think you can make generate go file more automatical by adding the  extra step to Makefile

STRRL:
OK, I will do this.

breeswish:
Closed by #171

PR #171, opened 2020-03-05
diagnose: support hot-swap diagnose report template file during debugging

STRRL:
UCP #150 

## What have you changed? 

Support hot-swap with `.gohtml` template file without restart `tidb-dashboard`.

- move html content into new .html file: index.html, table.html
- use `go-bindata-assetfs` generate resource .go file for production usage.
- create a customized PreferLocalFileHTMLProduction to replace HTMLProduction(in gin), It will reload content from origin local file **if exists**
- new makefile step `template`, it generates resource .go file. 
- use build tag `hot_wap` to switch this feature

## What are the type of the changes? 

- Improvement 

## How has this PR been tested?

Manually tested, execute `tidb-dashboard` compiled by `HOT_SWAP=1 make server`, I can see `index.gohtml` changes with just refreshing the browser.

## Does this PR affect documentation (docs/docs-cn) update? 

No.

## Refer to a related PR or issue link

https://github.com/pingcap-incubator/tidb-dashboard/issues/150


Makefile
```diff
@@ -45,7 +50,10 @@ ui: swagger_client
 	src/apps/keyvis/download_dummydata.sh &&\
 	REACT_APP_DASHBOARD_API_URL="" npm run build
 
-server:
+template:
+	scripts/generate_diagnose_report_template.sh
```
breeswish: Prefer to call it `embed_template` for two reasons:
- It embeds template into the source code instead of generates a new template
- We may have other templates in future.

Makefile
```diff
@@ -16,6 +16,10 @@ ifeq ($(UI),1)
 BUILD_TAGS += ui_server
 endif
 
+ifeq ($(HOT_SWAP),1)
+	BUILD_TAGS += hot_swap
```
breeswish: hot_swap_template?

pkg/apiserver/diagnose/templates.go
```diff
@@ -17,182 +17,11 @@ import (
...
-          {{ end }}
-        {{ end }}
-      </tbody>
-    </table>
-  </div>
-{{ end }}
-`
+func FetchTemplates() []utils.TemplateInfoWithFilename {
+	var result []utils.TemplateInfoWithFilename
+	result = append(result, readFromFS("sql-diagnosis/index", "pkg/apiserver/diagnose/templates/index.gohtml"))
```
breeswish: Is it possible to support adding new templates?

pkg/apiserver/diagnose/.gitignore
```diff
@@ -0,0 +1 @@
+/embedded_diagnose_report_template.go
```
breeswish: I prefer to embed the generated template source code into the repository, since Dashboard will also be included in the PD (see https://github.com/pingcap-incubator/tidb-dashboard#getting-started-pd-with-tidb-dashboard ). Embedding source code in the repository now can make it easy.

pkg/apiserver/diagnose/templates.go
STRRL: Of cause yes. Just duplicating this line and change template name and filename.

breeswish: Hmm, I mean if new templates are added on the disk, can we support tracking them without reboot the app? For example, as you can see, `sql-diagnosis/table` template is referenced by `sql-diagnosis/index` template, so if newly added templates are not tracked, then what we can do with existing template is limited.

STRRL:
> I'm mostly thinking to support loading new files and maybe libraries in https://tech.townsourced.com/post/embedding-static-files-in-go/ can help you achieve that without much code.

I use `go-bindata-assetfs` to do this, because I have seen you already use this library to serve frontend file.

Other library seems like do similar thing, based on http/fs. It will not make many differents with switching to another library.

pkg/apiserver/diagnose/templates.go
STRRL: Oh, it can not right now. I will improve this!

breeswish:
@STRRL Actually I'm fine with switching another library, since go-bindata-assetfs has some other problems, for example, it cannot serve a compressed stream. If there could be a library that does similar things, but natively supports hot loading from disk then it could be awesome, means we don't need to craft these part by ourselves.

STRRL:
@breeswish  Here are still two thing I want to figure out:

First, about 
> it cannot serve a compressed stream.

What does "serve" mean, I notice that `go-bindata-assetfs` already use gzip in generated .go file.

And then, if we want to use a library that "natively" supports load local file from disk, it still need do other things. Because neither `html/template` nor `go-gin` do NOT support reload template from `http/fs`.

Am I still on right road? Let me know if I miss something important.

STRRL:
@breeswish Can you help me with this github action CI build failed?

https://github.com/pingcap-incubator/tidb-dashboard/pull/171/checks?check_run_id=487564336

I can NOT reproduce this problem neither my macos nor ubuntu vm by running `make dev`.

And I can not find what exactly what the diffs.

breeswish:
Hi!

> First, about
> 
> > it cannot serve a compressed stream.
> 
> What does "serve" mean, I notice that `go-bindata-assetfs` already use gzip in generated .go file.

Normally we would want UI assets served as gzip stream to users. go-bindata-assetfs already stores gzipped data in the source file, but these gzipped data cannot be served directly to user. Instead, it only provides uncompressed interface. Thus, the total pipeline becomes: gzip uncompress -> gzip compress again -> send to user. This pipeline can be improved by some of the other asset solutions in the link.


> And then, if we want to use a library that "natively" supports load local file from disk, we still need do other things. Because neither `html/template` nor `go-gin` do NOT support reload template from `http/fs`.

According to the gin's implementation, it will reload the template every time in debug mode:

```golang
func (r HTMLDebug) Instance(name string, data interface{}) Render {
	return HTML{
		Template: r.loadTemplate(),  // <-------- note this line
		Name:     name,
		Data:     data,
	}
}
```

breeswish:
@STRRL 

According to CI, the failure is caused by template files being re-generated on CI which produces different content. Currently I don't have good ideas to solve it, since we indeed want to check whether embedded source code is update to date. I cannot figure out a way for binasset not to include the modification time in the source code as well. Maybe switching to other asset solution can solve this with ease.

STRRL:
@breeswish 
I can feel your strong will to use another library. XD   
I will try others.

breeswish:
@STRRL I simply think it not good to rely more on binassetfs, which is destined to be replaced.. LOL

STRRL:
I haven't found the silver bullet. Each library has different issues.  

I will choose one from [fileb0x](https://github.com/UnnoTed/fileb0x) and [vfsgen](https://github.com/shurcooL/vfsgen).

STRRL:
Finally I use vfsgen. 

I create a func `loadGlobFromVfs` just like `LoadHTMLGlob`in gin. Now it supports reload new `.gohtml` file just created.

breeswish:
Sorry for the late reply. I will take a look at your solution soon!

baurine:
Hi @STRRL , I wonder how you generate the vfs_vfsdata.go this file, I run `go generate` in the project root folder, get following error:

```
> go generate
can't load package: package .: build constraints exclude all Go files in .../pingcap-incubator/tidb-dashboard
```

Then I run `go generate` in `pkg/apiserver/diagnose` folder, get following error:

```
> pwd
.../pingcap-incubator/tidb-dashboard/pkg/apiserver/diagnose
> go generate
2020-04-02 09:54:49.298813 I | open pkg/apiserver/diagnose/templates: no such file or directory
exit status 1
2020/04/02 09:54:49 exit status 1
assets_hot_swap.go:2: running "vfsgendev": exit status 1
```

baurine:
Tried, this works for me:

```go
func getWd() string {
	path, err := os.Getwd()
	if err != nil {
		log.Println(err)
	}
	return path
}
var Vfs http.FileSystem = http.Dir(getWd() + "/templates")
```

```
> go generate pkg/apiserver/diagnose/assets_hot_swap.go
writing vfs_vfsdata.go
```

breeswish:
@all-contributors please add @STRRL for code

Issue #108, opened 2020-03-01
UCP: Refactor the `gorm.DB.Raw()` code


baurine:
## Description

This task is to eliminate the `gorm.DB.Raw()` code in `pkg/apiserver/diagnose/query.go` to make it more elegant.

Example: see this [commit](https://github.com/pingcap-incubator/tidb-dashboard/pull/34/commits/175c39a5f36b11388896a3acc09e53d2ffca0ea9)

For example, a function is implemented by gorm.DB.Raw():

```go
func QueryTimeRanges(db *gorm.DB) ([]*TimeRange, error) {
	sql := `select
	distinct summary_begin_time,summary_end_time
	from cluster_events_statements_summary_by_digest_history
	order by summary_begin_time desc`
	timeRanges := []*TimeRange{}

	db.Exec(selectPerformanceDB)
	rows, err := db.Raw(sql).Rows()
	if err != nil {
		return timeRanges, err
	}
	defer rows.Close()

	for rows.Next() {
		timeRange := new(TimeRange)
		err = rows.Scan(&timeRange.BeginTime, &timeRange.EndTime)
		if err != nil {
			return timeRanges, err
		}
		timeRanges = append(timeRanges, timeRange)
	}
	return timeRanges, nil
}
```

It can be simplified and get more elegant by the following code:

```go
func QueryTimeRanges(db *gorm.DB) (result []*TimeRange, err error) {
	err = db.
		Select(`
			DISTINCT
			summary_begin_time AS begin_time,
			summary_end_time AS end_time
		`).
		Table("PERFORMANCE_SCHEMA.cluster_events_statements_summary_by_digest_history").
		Order("summary_begin_time DESC").
		Find(&result).Error
	return result, err
}
```

## Score 

* 300

## Mentor(s)

* @crazycs520 
* @baurine 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Go

zzh-wisdom:
I have a question about function `genSumarySQLs`. I run this function separately, and the printed result is as follows.
[...]

PR #176, opened 2020-03-05
pkg/apiserver/diagnose/query.go: eliminate the gorm.DB.Raw()

zzh-wisdom:
UCP #108

## What have you changed? 

eliminate the `gorm.DB.Raw()`, and use `rows()`
Define a new function `scanRows`, which is used to scan `sql.Rows`
delete func `genSummarySQLs` and `genDetailSQLs`
add func `querySummary` and `queryDetail`

## What are the type of the changes? 

Improvement

## How has this PR been tested? 

Have run command make run successfully

## Does this PR affect documentation (docs/docs-cn) update?

No


zzh-wisdom:
@baurine

breeswish:
Hi, please avoid constructing SQLs that includes user specified parameters directly. Instead you need to use placeholders as examples in http://gorm.io/docs/sql_builder.html#Run-Raw-SQL and https://github.com/pingcap-incubator/tidb-dashboard/pull/34/commits/175c39a5f36b11388896a3acc09e53d2ffca0ea9 .

breeswish:
Additionally the output of the SQL should be changed to schema aware, i.e. the result is an array of a structure, instead of HashMap.

zzh-wisdom:
I don't quite understand what “use placeholders” means。
I guess I shoud change the codes such as
`conditions := fmt.Sprintf("t0.time >= '%s' and t0.time < '%s'", startTime, endTime)` 
into
`db.where("t0.time >= ? and t0.time < ?", startTime, endTime)`
is it?

zzh-wisdom:
In addition to the above, do I need to modify other codes? 
For example,
`fields := fmt.Sprintf("'%s','', sum(t0.value)", t.name)`
`tables := fmt.Sprintf("metrics_schema.%s t0", t.sumTbl)`
the select fields and table name are also created by constructing

zzh-wisdom:
@breeswish

zzh-wisdom:
Some queries return non-fixed columns, how to use an array of structure to get the results.
For example,
`sql = fmt.Sprintf("select '%[1]v',`%[2]v`, sum(value) from metrics_schema.%[3]v %[4]s group by `%[2]v` having sum(value) > 0 order by sum(value) desc",
		t.name, strings.Join(t.labels, "`,`"), t.tbl, condition)`
it returns Variable `t.labels`.
@breeswish

zzh-wisdom:
Hi, I have improved the code. Please review, thanks

zzh-wisdom:
@breeswish

zzh-wisdom:
@baurine

pkg/apiserver/diagnose/query.go
```diff
@@ -55,30 +55,69 @@ func (t AvgMaxMinTableDef) queryRow(arg *queryArg, db *gorm.DB) (*TableRowDef, e
...
+	row := avgMaxMinToSlice(result)
 	if len(t.label) == 0 {
-		return t.genRow(rows[0], nil), nil
+		return t.genRow(row, nil), nil
+	}
+
+	label := "`" + t.label + "`"
+	fields = fmt.Sprintf(`
+		'%s' as name,
+		%s as label,
```
breeswish: Maybe this is better:

```
`%s` as label,
```

pkg/apiserver/diagnose/query.go
```diff
@@ -382,67 +524,154 @@ func (t totalValueAndTotalCountTableDef) genRow(values []string, subValues [][]s
...
-		fields += fmt.Sprintf("t%v.*", i)
-		tbls += fmt.Sprintf(" (%s) as t%v ", sql, i)
+		aggrQuantileMaxField += fmt.Sprintf("max(%v.value)", tableAlias)
+		tableNum++
 	}
-	joinSQL := fmt.Sprintf("select %v from %v", fields, tbls)
-	return joinSQL
+	if len(aggrQuantileMaxField) == 0 {
+		aggrQuantileMaxField = "''"
+	}
```
breeswish: you can avoid building the field list if `quantiles []float64` are changed to a closed set, like this:

```golang
type QuantileFields {
  Q999 float64
  Q99 float64
  Q90 float64
  Q80 float64
}

type Quantiles {
  Q999 bool
  Q99 bool
  Q90 bool
  Q80 bool
}
```

When using it:

```golang
type XxxRow {
  QuantileFields
}

func (...) querySummary(..., quantiles: Quantiles) ([]XxxRow, error) {
  ...
}
```

Actually quantiles are indeed a closed set:

```golang
quantiles: []float64{0.999, 0.99, 0.90, 0.80},
```

zzh-wisdom:
> Hi, can you modify the return signature of these interfaces to make them structural? As you can see, finally all of these tables have a "schema":
> 
> ```go
> func GenerateReportError(errRows []TableRowDef) *TableDef {
> 	return &TableDef{
> 		Category:  []string{CategoryError},
> 		Title:     "Generate Report Error",
> 		CommentEN: "",
> 		CommentCN: "",
> 		Column:    []string{"CATEGORY", "TABLE", "ERROR"},
> 		Rows:      errRows,
> 	}
> }
> 
> func GetHeaderTimeTable(startTime, endTime string, db *gorm.DB) (TableDef, error) {
> 	return TableDef{
> 		Category:  []string{CategoryHeader},
> 		Title:     "Report Time Range",
> 		CommentEN: "",
> 		CommentCN: "",
> 		Column:    []string{"START_TIME", "END_TIME"},
> 		Rows: []TableRowDef{
> 			{Values: []string{startTime, endTime}},
> 		},
> 	}, nil
> }
> ```
> 
> So the best way is to simply make each table definition schema-aware (e.g. returning `[]AvgMaxMin`), instead of returning implicit schema data via `*TableRowDef` which is just an 2d array.

Does it mean I also need to modify the codes where use these interfaces?

breeswish:
> > Hi, can you modify the return signature of these interfaces to make them structural? As you can see, finally all of these tables have a "schema":
> > ```go
> > func GenerateReportError(errRows []TableRowDef) *TableDef {
> > 	return &TableDef{
> > 		Category:  []string{CategoryError},
> > 		Title:     "Generate Report Error",
> > 		CommentEN: "",
> > 		CommentCN: "",
> > 		Column:    []string{"CATEGORY", "TABLE", "ERROR"},
> > 		Rows:      errRows,
> > 	}
> > }
> > 
> > func GetHeaderTimeTable(startTime, endTime string, db *gorm.DB) (TableDef, error) {
> > 	return TableDef{
> > 		Category:  []string{CategoryHeader},
> > 		Title:     "Report Time Range",
> > 		CommentEN: "",
> > 		CommentCN: "",
> > 		Column:    []string{"START_TIME", "END_TIME"},
> > 		Rows: []TableRowDef{
> > 			{Values: []string{startTime, endTime}},
> > 		},
> > 	}, nil
> > }
> > ```
> > 
> > 
> > So the best way is to simply make each table definition schema-aware (e.g. returning `[]AvgMaxMin`), instead of returning implicit schema data via `*TableRowDef` which is just an 2d array.
> 
> Does it mean I also need to modify the codes where use these interfaces?

Yes.

zzh-wisdom:
I modified the code to my own understanding.
I found that in the code where calls function `func (t totalValueAndTotalCountTableDef) queryRow(arg *queryArg, db *gorm.DB) (*TableRowDef, error)`, the parameter `arg.quantiles` is either empty or four float64. So I added a field `flag bool` to the returned structure 
`type TotalValueAndTotalCount struct {
	Name       string
	Label      string
	TotalSum   float64
	TotalCount float64
	MaxValueOfQuantile
	flag bool
}`
Because wanting to avoid building quantile fields list, each time do query will always return four `max(value)` of each quantiles and sometimes the `arg.quantiles` , in the old code, is empty. In the new code, when the value of a field in structure `Quantile` is `0.0`, it means empty.

But in  `func (t totalTimeByLabelsTableDef) queryRow(arg *queryArg, db *gorm.DB) (*TableRowDef, error) `, I found that `arg.quantiles` was always not empty. So I did not add `flag bool` in
`
type TotalTimeByLabels struct {
	Name          string
	Label         string
	RatioSumValue float64
	TotalTime     float64
	TotalCount    float64
	MaxValueOfQuantile
}
`

I didn't modify the interface because it involves a lot of code. For each information query, I have defined the function separately and returned the structure type, such as ` []*TotalTimeByLabels`. Maybe this also is schema-aware.
@breeswish

zzh-wisdom:
I modified the code to my own understanding.
I found that in the code where calls function `func (t totalValueAndTotalCountTableDef) queryRow(arg *queryArg, db *gorm.DB) (*TableRowDef, error)`, the parameter `arg.quantiles` is either empty or four float64. So I added a field `flag bool` to the returned structure 
```
type TotalValueAndTotalCount struct {
	Name       string
	Label      string
	TotalSum   float64
	TotalCount float64
	MaxValueOfQuantile
	flag bool
}
```
Because wanting to avoid building quantile fields list, each time do query will always return four `max(value)` of each quantiles and sometimes the `arg.quantiles` , in the old code, is empty. In the new code, when the value of a field in structure `Quantile` is `0.0`, it means empty.

But in  `func (t totalTimeByLabelsTableDef) queryRow(arg *queryArg, db *gorm.DB) (*TableRowDef, error) `, I found that `arg.quantiles` was always not empty. So I did not add `flag bool` in
```
type TotalTimeByLabels struct {
	Name          string
	Label         string
	RatioSumValue float64
	TotalTime     float64
	TotalCount    float64
	MaxValueOfQuantile
}
```

I didn't modify the interface because it involves a lot of code. For each information query, I have defined the function separately and returned the structure type, such as ` []*TotalTimeByLabels`. Maybe this also is schema-aware.
@breeswish

pkg/apiserver/diagnose/query.go
```diff
@@ -55,53 +71,107 @@ func (t AvgMaxMinTableDef) queryRow(arg *queryArg, db *gorm.DB) (*TableRowDef, e
...
+		min(value) as min
+		`, t.name)
+	table := fmt.Sprintf("metrics_schema.%s", t.tbl)
+
+	query := db.
+		Select(fields).
+		Table(table).
+		Where("time >= ? AND time < ?", arg.startTime, arg.endTime)
+	if len(t.condition) > 0 {
+		query = query.Where(t.condition)
```
breeswish: I think this should be parameterized.

zzh-wisdom:
I probably understand what you mean. I think it takes too many changes to achieve this goal. Many codes in the file `report.go` should be also modified. So I accept the second solution that opening another UCP to achieve the goal. I'm sorry that I can not modify the code properly. By the way, what else code I need to modify for this PR?

zzh-wisdom:
@breeswish

breeswish:
Hi, @zzh-wisdom For this PR, you can revert the change of the structural return if you wish to do it in another PR. This PR can focus on changing the raw SQL string using parameterized `where()` and other query primitives. Sorry for the inconvenience!

zzh-wisdom:
@breeswish ,I have modified the codes. Please take a look.

pkg/apiserver/diagnose/query.go
```diff
@@ -118,37 +152,65 @@ func (t sumValueQuery) queryRow(arg *queryArg, db *gorm.DB) (*TableRowDef, error
...
-	if len(rows) == 0 {
+	if len(results) == 0 {
 		return nil, nil
 	}
 	if len(t.labels) == 0 {
-		return t.genRow(rows[0], nil), nil
+		return t.genRow(results[0], nil), nil
+	}
+
+	labelFields := fmt.Sprintf("`%v`", strings.Join(t.labels, "`||','||`"))
```
breeswish: the original SQL is assembled using `strings.Join(t.labels, "`,`")`

pkg/apiserver/diagnose/query.go
```diff
@@ -118,37 +152,65 @@ func (t sumValueQuery) queryRow(arg *queryArg, db *gorm.DB) (*TableRowDef, error
...
+		Rows()
 	if err != nil {
 		return nil, err
 	}
-	for i := range subRows {
-		row := subRows[i]
-		row[1] = strings.Join(row[1:1+len(t.labels)], ",")
-		newRow := row[:2]
-		newRow = append(newRow, row[1+len(t.labels):]...)
-		subRows[i] = newRow
```
breeswish: maybe need to revert?

pkg/apiserver/diagnose/query.go
```diff
@@ -181,40 +243,26 @@ type totalTimeByLabelsTableDef struct {
...
 func (t totalTimeByLabelsTableDef) queryRow(arg *queryArg, db *gorm.DB) (*TableRowDef, error) {
-	sql := t.genSumarySQLs(arg.totalTime, arg.startTime, arg.endTime, arg.quantiles)
-	rows, err := querySQL(db, sql)
+	rows, err := t.querySummary(db, arg.totalTime, arg.startTime, arg.endTime, arg.quantiles)
 	if err != nil {
 		return nil, err
 	}
-	if len(rows) == 0 {
-		return nil, nil
-	}
```
breeswish: why can we remove these lines?

pkg/apiserver/diagnose/query.go
```diff
@@ -181,40 +243,26 @@ type totalTimeByLabelsTableDef struct {
...
 	}
-
-	for i := range subRows {
-		row := subRows[i]
-		row[1] = strings.Join(row[1:1+len(t.labels)], ",")
-		newRow := row[:2]
-		newRow = append(newRow, row[1+len(t.labels):]...)
-		subRows[i] = newRow
-	}
-	return t.genRow(rows[0], subRows), nil
```
breeswish: ditto

pkg/apiserver/diagnose/query.go
zzh-wisdom: `||` is used as a connector in `gorm`, equivalent to concatenating strings

zzh-wisdom:
@breeswish  Are you convenient to take a look

PR #189, opened 2020-03-08
diagnose: fix cpu usage bug and refine report format 

crazycs520:
related PR: https://github.com/pingcap/tidb/pull/15210

### DIFF
--- a/pkg/apiserver/diagnose/report.go
+++ b/pkg/apiserver/diagnose/report.go
@@ -76,7 +76,7 @@ const (
 	// Category names.
 	CategoryHeader   = "header"
 	CategoryDiagnose = "diagnose"
-	CategoryNode     = "node"
+	CategoryLoad     = "load"
 	CategoryOverview = "overview"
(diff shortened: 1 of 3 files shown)


pkg/apiserver/diagnose/report.go
```diff
@@ -1524,19 +1534,36 @@ func getAvgMaxMinMemoryUsage(startTime, endTime string, db *gorm.DB) (*TableRowD
...
 		joinColumns:    []int{0, 1},
 		compareColumns: []int{2, 3, 4},
-		Column:         []string{"instance", "job", "AVG", "MAX", "MIN"},
+		Column:         []string{"INSTANCE", "JOB", "AVG", "MAX", "MIN"},
 	}
 	rows, err := getSQLRoundRows(db, sql, []int{2, 3, 4}, "")
 	if err != nil {
 		return table, err
 	}
+	specialHandle := func(row []string) []string {
```
reafans: How about extract specialHandle as a func

PR #191, opened 2020-03-09
clusterinfo: add alert-count checking

mapleFU:
Signed-off-by: mapleFU <1506118561@qq.com>


ui/src/apps/clusterInfo/components/MonitorAlertBar.js
```diff
@@ -1,11 +1,27 @@
...
   const { t } = useTranslation()
+  const [alertCounter, setAlertCounter] = useState(0)
+
+  useEffect(() => {
+    const fetchNum = async () => {
+      if (cluster === null || cluster.alert_manager === null) {
+        return
+      }
+      let resp = await axios.get(
+        `http://${cluster.alert_manager.ip}:${cluster.alert_manager.port}/api/v2/alerts`
```
breeswish: what if TLS?

ui/src/apps/clusterInfo/translations/zh-CN.yaml
```diff
@@ -18,7 +18,8 @@ cluster_info:
     title: 监控和告警
     view_monitor: 查看监控
     view_monitor_warn: 监控不可用
-    view_alerts: 查看告警
+    view_alerts: '查看{{alertCount}}条告警'
```
breeswish: Please add space between wide char and half char.

ui/src/apps/clusterInfo/components/MonitorAlertBar.js
```diff
@@ -1,11 +1,17 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
 import { Icon, Card, Skeleton } from 'antd'
 import { Link } from 'react-router-dom'
 import { useTranslation } from 'react-i18next'
+import axios from 'axios'
```
breeswish: What is this used for?

pkg/apiserver/clusterinfo/fetch.go
```diff
@@ -35,6 +35,7 @@ type ClusterInfo struct {
 	} `json:"pd"`
 	Grafana      *clusterinfo.GrafanaInfo      `json:"grafana"`
 	AlertManager *clusterinfo.AlertManagerInfo `json:"alert_manager"`
+	AlertCount   *int                          `json:"alert_count"`
```
breeswish: I prefer to use another structure and interface for alert count, because it is not used elsewhere.

pkg/utils/clusterinfo/fetcher.go
```diff
@@ -376,3 +376,38 @@ func getPDNodesHealth(pdEndPoint string, httpClient *http.Client) (map[string]st
...
+
+	return len(alerts), nil
+}
+
+func GetAlertCountByAddress(address string, httpClient *http.Client) (int, error) {
+	ip, port, err := parseHostAndPortFromAddress(address)
+	if err != nil {
+		return 0, err
+	}
+	return GetAlertCount(&AlertManagerInfo{
```
breeswish: why binaryPath need to be specified when getting alert count?

mapleFU: Removed, PTAL again.

pkg/utils/clusterinfo/fetcher.go
```diff
@@ -376,3 +376,34 @@ func getPDNodesHealth(pdEndPoint string, httpClient *http.Client) (map[string]st
...
+
+// GetAlertCountByAddress receives alert manager's address like "ip:port", and it returns the
+//  alert number of the alert manager.
+func GetAlertCountByAddress(address string, httpClient *http.Client) (int, error) {
+	ip, port, err := parseHostAndPortFromAddress(address)
+	if err != nil {
+		return 0, err
+	}
+
+	apiAddress := fmt.Sprintf("http://%s:%d/api/v2/alerts", ip, port)
```
breeswish: why not specify address use %s directly?

mapleFU: I think here needs checkings, else the response will be ambigious.

PR #196, opened 2020-03-10
profiling: support auto collection

rleungx:
This PR supports auto profiling.


pkg/apiserver/profiling/service.go
```diff
@@ -131,8 +168,79 @@ func (s *Service) start(c *gin.Context) {
...
+// @Summary Start auto profiling
+// @Description Start auto profiling
+// @Produce json
+// @Param req body StartRequest true "auto profiling request"
+// @Security JwtAuth
+// @Success 200 {string} string "success"
+// @Failure 400 {object} utils.APIError
+// @Router /profiling/auto/start [post]
+func (s *Service) autoStart(c *gin.Context) {
+	if atomic.LoadUint32(&s.isAutoCollectRunning) > 0 {
```
HunDunDM: Would it be better to use `CompareAndSwap`?

breeswish: Yes, we need to use CAS.

pkg/apiserver/profiling/service.go
```diff
@@ -131,8 +168,79 @@ func (s *Service) start(c *gin.Context) {
...
+}
+
+// @Summary Stop auto profiling
+// @Description Stop auto profiling
+// @Produce json
+// @Security JwtAuth
+// @Success 200 {string} string "success"
+// @Router /profiling/auto/stop [post]
+func (s *Service) autoStop(c *gin.Context) {
+	if atomic.LoadUint32(&s.isAutoCollectRunning) == 0 {
```
breeswish: CAS

pkg/apiserver/profiling/service.go
```diff
@@ -131,8 +168,79 @@ func (s *Service) start(c *gin.Context) {
...
+		c.JSON(http.StatusBadRequest, "auto profiling job has been stopped")
+	}
+
+	atomic.StoreUint32(&s.isAutoCollectRunning, 0)
+	c.JSON(http.StatusOK, "success")
```
breeswish: we can use APIEmptyResponse

rleungx:
PTAL @breeswish @HunDunDM

pkg/apiserver/profiling/service.go
```diff
@@ -131,8 +168,78 @@ func (s *Service) start(c *gin.Context) {
...
+		if atomic.LoadUint32(&s.isAutoCollectRunning) == 0 {
+			return
+		}
+		select {
+		case <-s.ctx.Done():
+			return
+		case <-timer.C:
+			_, _ = s.collect(req)
+			timer.Reset(interval)
+		default:
```
HunDunDM: This `for` loop does not block.

HunDunDM:
close by #439

PR #198, opened 2020-03-10
diagnose: refine compare diff table and add process memory usage. 

crazycs520:
Before

image

This PR:

image
image


reafans:
rest lgtm, please fix the CI

PR #208, opened 2020-03-12
Re-organize documents for contributing and highlight SIG and UCP

breeswish:


CONTRIBUTING.md
```diff
@@ -2,149 +2,152 @@
...
-### Building and running
+The followings are required for developing TiDB Dashboard:
 
-At this point, you can build and run TiDB Dashboard. 
+- git - Version control
+- make - Build tool (run common workflows)
+- [Golang 1.13+](https://golang.org/) - To compile the server.
+- [Node.js 12+](https://nodejs.org/) - To compile the front-end.
+- [Yarn 1.21+](https://classic.yarnpkg.com/en/docs/install) - To manage front-end dependencies。
+- [Java 8+](https://www.java.com/ES/download/) - To generate JavaScript API client by OpenAPI specification.
```
Deardrops: Do we need add [flamegraph](https://github.com/brendangregg/FlameGraph) here?

breeswish: It doesn't affect other features for now. I think we don't need to explicit mention it.

Issue #177, opened 2020-03-05
support TLS when TLS authentication between the MySQL client and TiDB server is enabled


LinuxGit:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

If [TLS authentication between the MySQL client and TiDB server](https://github.com/pingcap/docs/blob/7a62321b3ce9318cbda8697503c920b2a01aeb3d/how-to/secure/enable-tls-clients.md) is enabled,  it is required to specify the client certificate when TiDB Dashboard wants to create sql connection to TiDB. For example, login and statement API need to create sql connection.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

breeswish:
Resolved by #219

PR #219, opened 2020-03-16
add TLS support for db connection

niedhui:
UCP #235

Add TLS support between MySQL client and TiDB  #177 
start tidb-dashboard with extra arguments below
```
--tidb-ca=/path/ca.pem --tidb-cert=/path/client.pem --tidb-key=/path/client-key.pem
```


pkg/config/config.go
```diff
@@ -20,5 +20,10 @@ import (
 type Config struct {
 	DataDir    string
 	PDEndPoint string
-	TLSConfig  *tls.Config
+
+	// TLS config for mTLS authentication between TiDB components.
+	TLSConfig *tls.Config
```
breeswish: Maybe call it ClusterTLS?

PR #229, opened 2020-03-17
refactor: refine project with fx

HunDunDM:
* dashboard: can start and stop like an app
* keyvisual: can start and stop like an app
* pd: remove `EtcdProvide`
* middleware: handle requests when the app is stopped


pkg/apiserver/apiserver.go
```diff
@@ -36,49 +38,133 @@ import (
...
+			foo.Register,
+			info.Register,
+			clusterinfo.Register,
+			profiling.Register,
+			logsearch.Register,
+			statement.Register,
+			diagnose.Register,
+			keyvisual.Register,
+			// Must be at the end
+			s.status.Invoke,
```
breeswish: prefer to call it `s.runningStatus.HookLifecycle`

pkg/apiserver/apiserver.go
```diff
@@ -36,49 +38,133 @@ import (
...
 
-type Services struct {
-	Store         *dbstore.DB
-	TiDBForwarder *tidb.Forwarder
-	KeyVisual     *keyvisual.Service
-	EtcdProvider  pd.EtcdProvider
-	HTTPClient    *http.Client
+type Service struct {
+	app    *fx.App
+	status *utils2.ServiceStatus
```
breeswish: `runningStatus *utils2.ServiceRunningStatus`

pkg/apiserver/apiserver.go
```diff
@@ -36,49 +38,133 @@ import (
...
+	endpoint.Use(s.status.MWHandleStopped(gin.WrapH(s.stoppedHandler)))
+	endpoint.Any("/*any", s.handler)
+}
+
+func (s *Service) Start(ctx context.Context) error {
+	s.app = fx.New(
+		fx.Logger(utils2.NewFxPrinter()),
+		fx.Provide(
+			newAPIHandlerEngine,
+			s.provide,
```
breeswish: `s.provideLocals`?

pkg/apiserver/apiserver.go
```diff
@@ -36,49 +38,133 @@ import (
...
+	s.apiHandlerEngine.HandleContext(c)
 }
 
-func NewTemplate(r *gin.Engine, name string) *template.Template {
-	return template.New(name).Funcs(r.FuncMap)
+func (s *Service) provide() *config.Config {
+	return s.config
 }
+
+var StoppedHandler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
breeswish: Should we move this handler to cmd package?

cmd/tidb-dashboard/main.go
```diff
@@ -155,65 +151,49 @@ func getContext() context.Context {
...
-		Store:         store,
-		KeyVisual:     keyvisualService,
-		TiDBForwarder: tidbForwarder,
-		EtcdProvider:  etcdProvider,
-		HTTPClient:    httpClient,
-	}
 	mux := http.DefaultServeMux
-	mux.Handle("/dashboard/", http.StripPrefix("/dashboard", uiserver.Handler()))
-	mux.Handle("/dashboard/api/", apiserver.Handler("/dashboard/api", cliConfig.CoreConfig, services))
+	mux.Handle("/dashboard/", s.NewStatefulHandler(http.StripPrefix("/dashboard", uiserver.Handler())))
```
breeswish: Why only UI is protected?

breeswish: When this is running standlone, maybe it is not very necessary to reject the UI since it will not be stopped.. Then `s.NewStatefulHandler` can be removed.

pkg/apiserver/apiserver.go
HunDunDM: I think it is more unified to change to `Register`. It may depend on more than just `Lifecycle`

pkg/apiserver/apiserver.go
HunDunDM: It looks like `Running` should be an enum of `ServiceStatus`

pkg/apiserver/apiserver.go
HunDunDM: It needs to be used in PD.

PR #231, opened 2020-03-17
clusterinfo: Bugfix clusterinfo warning

mapleFU:


ui/src/apps/clusterInfo/components/ClusterInfoTable.js
```diff
@@ -76,13 +76,14 @@ function pushNodes(key, cluster, dataSource, t) {
   ) {
     const nodes = cluster[key].nodes
     dataSource.push({
+      key: key + '(' + nodes.length + ')',
```
breeswish: why key does include `nodes.length`?

ui/src/apps/clusterInfo/components/ClusterInfoTable.js
```diff
@@ -101,6 +102,7 @@ function wrapNode(node, comp, id, t) {
     )
   }
   return {
+    key: `${comp}:${node.ip}:${node.port}`,
```
breeswish: i guess ip+port is enough for a key.

mapleFU: no, I'll tried it, but different components may have same key, and it will raise warning.

ui/src/apps/clusterInfo/components/ClusterInfoTable.js
mapleFU: You're write, It doesn't need. I'll change it.

ui/src/apps/clusterInfo/components/ClusterInfoTable.js
breeswish: why will different compoennt have same ip+port?

mapleFU: Well, thats my testing config...I think leave a comp here doesn't increase any complexity, and it's proper because it's just a sub-key of `comp`.

ui/src/apps/clusterInfo/components/ClusterInfoTable.js
```diff
@@ -76,13 +76,14 @@ function pushNodes(key, cluster, dataSource, t) {
   ) {
     const nodes = cluster[key].nodes
     dataSource.push({
+      key: key,
```
breeswish: `key,`

PR #238, opened 2020-03-18
Extract keyvis

baurine:
- extract keyvis as a package
- fix circular dependencies warning

Test it in https://github.com/baurine/test-tidb-dashboard-packages , works fine.

Usage: https://github.com/baurine/test-tidb-dashboard-packages/blob/master/src/App.js


ui/packages/keyvis/src/components/KeyVis.less
```diff
@@ -1,9 +1,11 @@
-@import '~antd/es/style/themes/default.less';
+// @import '~antd/es/style/themes/default.less';
+@import '../../../../node_modules/antd/es/style/themes/default.less';
```
baurine: rollup-plugin-postcss can't recognize '~antd/...` path, should miss some configuration, temporary work around it by using the absolute path.

ui/packages/keyvis/src/components/KeyVis.less
```diff
@@ -1,9 +1,11 @@
...
 
 .PD-KeyVis {
-  height: 100vh;
-  display: flex;
-  flex-direction: column;
+  font-size: 14px;
+  width: 100%;
+  height: calc(100vh - 170px);
+  padding: 16px;
```
baurine: Because remove the `Card` component, so revert the css style back.

baurine:
All done, @breeswish PTAL.

baurine:
replace by #254 , closed

PR #254, opened 2020-03-19
Extract KeyVis

baurine:
- extract keyvis as a package
- fix circular dependencies warning

Test it in https://github.com/baurine/test-tidb-dashboard-packages , works fine.

Usage: https://github.com/baurine/test-tidb-dashboard-packages/blob/master/src/App.js

Keep the same effect as master branch:

企业微信20200319090629


baurine:
@breeswish , re-commit the PR, PTAL.

ui/packages/.packages_release_version
```diff
@@ -1,3 +1,3 @@
 # This file contains a version number which will be used to publish ui packages.
 # To trigger a new publish, simply increase this version number.
```
breeswish: To trigger a new publish, increase this version number, as well as bump all package versions by running `yarn run bump:packages`.

baurine: good advice, will add in the next PR.

Issue #113, opened 2020-03-02
UCP: Popup error prompt when request failed for `log searching` page 


Deardrops:
## Description

Popup error prompt when request failed for `log searching` page.

## Score 

* 50

## Mentor(s)

* @Deardrops 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Web programming, includes
	* Typescript
	* React (React Hook)

SSebo:
@you06 Is tikv and dashboard can pick only one challenge at same time?

SSebo:
@Deardrops mentor, PTAL

Deardrops:
> @you06 Is tikv and dashboard can pick only one challenge at same time?

@SSebo yes, You'd only have one picked Issue, not matter in which repo.

you06:
@SSebo No you can't pick up 2 issues from different repos. But you can work on both, and pick up before submit a PR, where there is a PR waiting for reviewed, you're allowed to pick up a new one.

SSebo:
@you06 @Deardrops thanks for your explanation.

PR #270, opened 2020-03-22
ucp logs: popup error prompt when bad request

SSebo:
UCP #113 
add popup when request error.


breeswish:
Thanks for the contribution!

- Need to handle the case that there is network failure, which does not have `error?.response?.data?.message` field.
- Also need to handle task errors in the search progress page:
image

  (You can simulate a task error by closing a TiDB and search logs over that TiDB.)

SSebo:
@breeswish hi, mentor, I extract a util function to handle search log page request error with message prompt.

ui/src/apps/searchLogs/components/utils.ts
```diff
@@ -139,3 +140,17 @@ export function getAddress(target: LogsearchSearchTarget | undefined): string {
...
+
+export async function req_with_err_prompt<T>(
+  req: Promise<T>
+): Promise<T | null> {
+  try {
+    return await req
+  } catch (error) {
+    let msg = error?.response?.data?.message
+    if (msg) {
+      message.error(msg)
```
Deardrops: Show error message directly may confuse user.
We'd better prompt some human-readable error message.
e.g.
`Failed to get download token for selected logs.`
`Failed to start a new search task, please try again later.`

SSebo: Agree, I'll change the function to accept a message parameter.

SSebo:
@Deardrops @breeswish Right now prompt function receive a message param to show reasonable error message to user, PTAL

Deardrops:
Here also need popup error prompt:
https://github.com/pingcap-incubator/tidb-dashboard/blob/318fbc0f575843679db7a182a14b12f1602aa271/ui/src/apps/searchLogs/components/SearchProgress.tsx#L232

https://github.com/pingcap-incubator/tidb-dashboard/blob/318fbc0f575843679db7a182a14b12f1602aa271/ui/src/apps/searchLogs/components/SearchProgress.tsx#L252

https://github.com/pingcap-incubator/tidb-dashboard/blob/master/ui/src/apps/searchLogs/components/SearchHistory.tsx#L73

ui/src/apps/searchLogs/components/SearchHeader.tsx
```diff
@@ -112,12 +113,15 @@ export default function SearchHeader({ taskGroupID }: Props) {
         patterns: searchValue.split(/\s+/), // 'foo boo' => ['foo', 'boo']
       },
     }
-    const result = await client.getInstance().logsTaskgroupPut(params)
-    const id = result.data.task_group?.id
-    if (!id) {
-      // promp error here
+
+    let result = await req_with_err_prompt(
```
breeswish: I think we can do it globally.

SSebo:
@breeswish @Deardrops hi mentors, I move `reqWithErrPrompt` to `packages/dashboard_clients`, and add more error message, PTAL.

ui/packages/dashboard_client/package.json
```diff
@@ -21,7 +21,8 @@
     "p": "yarn publish"
   },
   "dependencies": {
-    "axios": "^0.19.0"
+    "axios": "^0.19.0",
+    "antd": "^4.0.0"
```
Deardrops, line 25: Remove this?

ui/packages/dashboard_client/src/index.ts
```diff
@@ -16,6 +17,21 @@ function getBasePath(): string {
...
 }
 
+export async function reqWithErrPrompt<T>(
+  req: Promise<T>,
+  msg: string
+): Promise<T | null> {
+  try {
+    return await req
+  } catch (error) {
+    console.log(error)
```
Deardrops, line 27: console.error
```suggestion
    console.error(error)
```

ui/packages/dashboard_client/package.json
SSebo: prompt need `message` from `antd`

Deardrops:
LGTM, @breeswish PTAL

ui/packages/dashboard_client/src/index.ts
SSebo: fixed

SSebo:
looks like tiup download failed.

ui/packages/dashboard_client/src/index.ts
```diff
@@ -16,6 +17,21 @@ function getBasePath(): string {
   return basePath
 }
 
+export async function reqWithErrPrompt<T>(
```
breeswish, line 20: I think you can change current [axios interceptors](https://github.com/pingcap-incubator/tidb-dashboard/blob/master/ui/src/utils/client.js#L21) to introduce this logic.

Note that sometimes we want to handle errors manually, for example, in the sign in page we want to perform custom actions when password is incorrect. Considering this, there are two solutions:

Solution 1. Maybe better to change sign in page to return http code 200 or some other special http code in such scenario.

Solution 2. Pass additional custom configuration (like `skipDefaultErrorHandler`) in sign in page indicating that we want to handle error response manually.

Personally I prefer solution 2 if it works.

SSebo: @breeswish mentor, 
- If we put all error handle logic in axios interceptor, it is hard to pass front end i18n error message to popup
- Maybe we can just prompt what response `error.message` contains, let backend return i18n error message depends on request header like `language=en`.

breeswish: @SSebo 

> If we put all error handle logic in axios interceptor, it is hard to pass front end i18n error message to popup for each request.

Maybe you can just use the `i18next.t()` function, which will behave correctly, following the global i18n settings. Actually now in axios interceptors it is already being used to prompt an expired session message.

SSebo: @breeswish  For example `t('search_logs.error.search_log')`, how to pass `search_logs.error.search_log` into request to know which error message to prompt ?

breeswish: Maybe there is no need, since search log request will only fail when there are network errors and auth issues. If a custom message is needed, we need to follow the way of signIn.

breeswish:
Cleaning up outdated PRs. Feel free to reopen if there are new updates.

PR #272, opened 2020-03-23
clusterinfo: add more status for component tikv

mapleFU:
Signed-off-by: mapleFU <1506118561@qq.com>


pkg/utils/clusterinfo/fetcher.go
```diff
@@ -351,6 +351,10 @@ func storeStateToStatus(state string) ComponentStatus {
 		return ComponentStatusTombstone
 	case "offline":
 		return ComponentStatusOffline
+	case "down":
+		return ComponentStatusDown
+	case "disconnected":
+		return ComponentStatusDisconnected
```
breeswish: disconnected should be mapped to unreachable. For default maybe we should add Unknown status.

pkg/utils/clusterinfo/clusterinfo.go
```diff
@@ -20,6 +20,10 @@ const (
 	ComponentStatusUp          ComponentStatus = 1
 	ComponentStatusTombstone   ComponentStatus = 2
 	ComponentStatusOffline     ComponentStatus = 3
+
+	// PD's Store may have state name down or disconnected.
+	ComponentStatusDisconnected ComponentStatus = 4
```
breeswish: We can remove disconnected (or rename "unreachable" to "disconnected")

Issue #224, opened 2020-03-16
UCP: Support overriding TiDB node to request


breeswish:
## Description

When serving functionalities involves with TiDB SQL, like statements, TiDB Dashboard retrieves the TiDB node list from PD and then select one to send TiDB SQL.

However in K8S environment (by using tidb-operator), TiDB Dashboard need to send request to a fixed TiDB endpoint. Thus we need to support this behaviour.

Goals:

- Support overriding the TiDB node to request by specifying in dotenv file.

## Score

* 200

## Mentor(s)

* @breeswish 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Golang

breeswish:
Implemented via #276

PR #276, opened 2020-03-23
Add Tidb Endpoint

Yisaer:
UCP #224

If `TIDB_OVERRIDE_ENDPOINT` ENV is set, the server would read it to connect to the tidb-server. Otherwise it would work as the previous.

### DIFF
--- a/pkg/tidb/conn.go
+++ b/pkg/tidb/conn.go
@@ -31,7 +31,7 @@ import (
 )
 
 const (
-	EnvTidbOverrideEndpointKey = "TIDB_OVERRIDE_ENDPOINT"
+	envTidbOverrideEndpointKey = "TIDB_OVERRIDE_ENDPOINT"
 )
 
 func (f *Forwarder) GetDBConnProps() (host string, port int, err error) {
@@ -45,7 +45,7 @@ func (f *Forwarder) GetDBConnProps() (host string, port int, err error) {
 
 func (f *Forwarder) OpenTiDB(user string, pass string) (*gorm.DB, error) {
 	var addr string
-	addr = os.Getenv(EnvTidbOverrideEndpointKey)
+	addr = os.Getenv(envTidbOverrideEndpointKey)
 	if len(addr) < 1 {
 		host, port, err := f.GetDBConnProps()
 		if err != nil {
--- a/pkg/tidb/conn.go
+++ b/pkg/tidb/conn.go
@@ -34,7 +34,7 @@ const (
 	envTidbOverrideEndpointKey = "TIDB_OVERRIDE_ENDPOINT"
 )
 
-func (f *Forwarder) GetDBConnProps() (host string, port int, err error) {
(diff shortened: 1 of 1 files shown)


breeswish:
Thanks for the PR! I see that you are using command line arguments. This is indeed a common way to solve this problem, but not in this case. Dashboard is supposed to be also embedded in PD. So you need to either:

- Open another PR in PD to accept this override as well.
- Or change to use dotenv as described in the issue, which does not need to change PD.

Also I would prefer to call it `tidb-endpoint-override` (which is a combination of the host and port part) for simplicity. Also notice that there are TiDB status port calls, so you might also need a `tidb-status-endpoint-override`.

Yisaer:
@breeswish Do you mean passing the env var like `TIDB_ENDPOINT_OVERRIDE` and `TIDB_STATUS_ENDPOINT_OVERRIDE` ?

breeswish:
@Yisaer Yes. Or you can keep the current implementation but also modify the PD part.

Yisaer:
@breeswish  I'm ok to passing by env which is also convenient for operator to integrated with. I would change to this method sooner.

Yisaer:
@breeswish PTAL

pkg/tidb/forwarder.go
```diff
@@ -103,6 +138,13 @@ func NewForwarder(lc fx.Lifecycle, config *ForwarderConfig, etcdClient *clientv3
 			return f.Close()
 		},
 	})
+	host, port, err := f.parseOverrideTidbEndpoint()
```
breeswish: what if this env is not specified at all?

pkg/tidb/conn.go
```diff
@@ -29,6 +29,11 @@ import (
 )
 
 func (f *Forwarder) GetDBConnProps() (host string, port int, err error) {
```
breeswish: we can change this function to `getDBConnProps`, and move the override from here to `dsnConfig.Addr = fmt.Sprintf("%s:%d", host, port)` in `OpenTiDB`. In this way you don't need to parse the override config, like: 

```go
if tidbEndpointOverride != nil {
  dsnConfig.Addr = *tidbEndpointOverride
}
```

Yisaer: updated. PTAL

pkg/tidb/conn.go
```diff
@@ -28,6 +30,10 @@ import (
 	_ "github.com/jinzhu/gorm/dialects/mysql"
 )
 
+const (
+	EnvTidbOverrideEndpointKey = "TIDB_OVERRIDE_ENDPOINT"
+)
+
 func (f *Forwarder) GetDBConnProps() (host string, port int, err error) {
```
breeswish: prefer to mark this function as private, since it does not provide an overridden address.

Yisaer: updated.

breeswish:
@all-contributors please add @Yisaer   for code

PR #278, opened 2020-03-24
UI e2e test

baurine:
- E2E test by puppeteer and jest

PR #281, opened 2020-03-24
e2e test for login by jest and puppeteer

baurine:


baurine:
what did compared with #278 :
- move tests out of src folder
- add related doc

@breeswish PTAL

btw, it seems the UI build time doesn't decline much 😢

ui/tests/e2e/login.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    'should login success by correct password',
+    async () => {
+      const page = await browser.newPage()
+
+      await page.goto(LOGIN_URL)
+
+      const title = await page.title()
+      expect(title).toBe('TiDB Dashboard')
+
+      await page.waitForSelector('button[type=submit]')
```
breeswish: Looks like we'd better assign an ID (or a customized attribute) to pinpoint those e2e interact targets, to avoid accidentally selecting others

ui/tests/e2e/login.test.ts
```diff
@@ -0,0 +1,76 @@
...
+      await page.waitForSelector('.ant-message-success')
+      const success = await page.$eval('.ant-message-success', (el) =>
+        el ? true : false
+      )
+      expect(success).toBe(true)
+      const content = await page.$eval(
+        '.ant-message-success span',
+        (el) => el.textContent
+      )
+      expect(content).toEqual('Sign in successfully')
```
breeswish: prefer to check on page URL for the success case

ui/tests/e2e/login.test.ts
```diff
@@ -0,0 +1,76 @@
...
+      await page.goto(LOGIN_URL)
+
+      await page.waitForSelector('input#tidb_signin_password')
+      await page.type('input#tidb_signin_password', 'any')
+      await page.click('button[type=submit]')
+
+      await page.waitForSelector('.ant-form-explain')
+      const fail = await page.$eval('.ant-form-explain', (el) =>
+        el ? true : false
+      )
```
breeswish: relying on ant design's className is very risky, can we have better ways? We can enlarge the select target (i.e. selecting a parent element that have a fixed ID specified by us).

ui/tests/e2e/login.test.ts
breeswish:
```js
const buttonEl = await page.waitForSelector(..)
await buttonEl.click()
```

ui/tests/e2e/login.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    'should login fail by incorrect password',
+    async () => {
+      const page = await browser.newPage()
+      await page.goto(LOGIN_URL)
+
+      await page.waitForSelector('input#tidb_signin_password')
+      await page.type('input#tidb_signin_password', 'any')
+      await page.click('button[type=submit]')
+
+      await page.waitForSelector('.ant-form-explain')
```
breeswish:
```
const messageEl = ...
const messageText = await messageEl.evaluate(node => node.innerText)
expect(messageText).toContain('TiDB authentication failed')
```

ui/tests/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "tests",
+  "version": "1.0.0",
+  "main": "index.js",
```
breeswish: I think this line should be removed since there is no `index.js`

ui/tests/e2e/login.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  ? 'http://127.0.0.1:12333/dashboard/#/signin'
+  : 'http://localhost:3000/#/signin'
+
+describe('Login', () => {
+  let browser
+  beforeAll(async () => {
+    // browser = await puppeteer.launch({
+    //   headless: false,
+    //   slowMo: 100,
+    // })
```
breeswish: we can disable the headless when `process.env.CI` is not set

.github/workflows/e2e-test.yaml
```diff
@@ -0,0 +1,74 @@
...
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://tiup-mirrors.pingcap.com/install.sh | sh
+          source /home/runner/.profile
+          tiup update --nightly --all
+          tiup playground nightly &
+      - name: Build UI
+        run: |
+          make ui
+        env:
+          NO_MINIMIZE: true
```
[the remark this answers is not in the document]
baurine: done!

.github/workflows/e2e-test.yaml
```diff
@@ -0,0 +1,74 @@
...
+        with:
+          path: ~/.tiup/components
+          key: ${{ runner.os }}-tiup
+          restore-keys: |
+            ${{ runner.os }}-tiup
+      - name: Install and run TiUP in the background
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://tiup-mirrors.pingcap.com/install.sh | sh
+          source /home/runner/.profile
+          tiup update --nightly --all
```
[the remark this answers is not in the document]
baurine: I tried before, it didn't work, I just tried again, the error message is:

```
Failed to start component `playground`
Error: component `playground` doesn't install
```

I remember the command `tiup playground nightly` can auto download the missed components before, but now this behavior change, it is also weird the cache seems don't work.

Maybe I can try to install the playground first, then call `tiup playground nightly`.

ui/tests/e2e/test_config.ts
```diff
@@ -0,0 +1,12 @@
+const URL_PREFIX = process.env.CI
```
HunDunDM: Can the environment variable directly specify the address?

baurine: You mean pass the address direclty vis env? like this: `ADDRESS=xxx yarn test`?

.github/workflows/e2e-test.yaml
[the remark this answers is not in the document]
baurine: Add `tiup install playground` doesn't work either, I think this should be tiup's bug, not sure whether they test in ubuntu, so I add the `tiup update --nightly  --all` back, we can remove it later if tiup fix it.

ui/tests/e2e/login.test.ts
baurine: tried but didn't find a very good way, I just replace it by judging 'div.has-error' element.

ui/tests/e2e/login.test.ts
baurine: It is a better way!

baurine:
hi @breeswish , so thank you for your detail review, learned much from it, all update is done, PTAL!

ui/tests/e2e/test_config.ts
HunDunDM: Yes. If I want to test the dashboard in PD, I need to pass the PD's address into it.

baurine: Got it, ok, I will update it soon.

baurine: done!

ui/tests/e2e/login.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      const errorContainer = await page.waitForSelector('div.has-error')
+      const containsPwdInput = await errorContainer.$eval(
+        'input#tidb_signin_password',
+        (el) => (el ? true : false)
+      )
+      expect(containsPwdInput).toBe(true)
+      const errorContent = await errorContainer.$eval(
+        '.ant-form-explain',
+        (el) => el.textContent
+      )
```
breeswish: I have a solution for this one: https://www.npmjs.com/package/expect-puppeteer

Use `await expect(page).toMatch('TiDB authentication failed')`. No more magic css selectors.

Other utilities provided by this library is also pretty handy.. like

`await expect(page).toClick('button#signin_btn')`

ui/tests/e2e/login.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    async () => {
+      const page = await browser.newPage()
+
+      await page.goto(LOGIN_URL)
+
+      const title = await page.title()
+      expect(title).toBe('TiDB Dashboard')
+
+      const loginBtn = await page.waitForSelector('button#signin_btn')
+      await Promise.all([page.waitForNavigation(), loginBtn.click()])
```
breeswish: Does this parallel really work? I guess loginBtn promise is always resolved before waitForNavigation

ui/tests/e2e/test_config.ts
```diff
@@ -0,0 +1,12 @@
+export let SERVER_URL = `${
+  process.env.SERVER_URL || 'http://localhost:3000/'
+}#/`
```
breeswish: prefer to remove this trailing `/` and adding it to the routes in next line.

ui/tests/e2e/login.test.ts
baurine: This is the official recommendation: https://pptr.dev/#?product=Puppeteer&version=v2.1.1&show=api-pagewaitfornavigationoptions

```
const [response] = await Promise.all([
  page.waitForNavigation(), // The promise resolves after navigation has finished
  page.click('a.my-link'), // Clicking the link will indirectly cause a navigation
]);
```

ui/tests/e2e/login.test.ts
baurine: OK, let me try it soon.

ui/tests/e2e/test_config.ts
baurine: Got, will update it soon.

ui/tests/e2e/login.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      await page.click('button#signin_btn')
+
+      const errorContainer = await page.waitForSelector('div.has-error')
+      const containsPwdInput = await errorContainer.$eval(
+        'input#tidb_signin_password',
+        (el) => (el ? true : false)
+      )
+      expect(containsPwdInput).toBe(true)
+      const errorContent = await errorContainer.$eval(
+        '.ant-form-explain',
```
breeswish: please avoid introducing ant design specific selectors, which does not have compatibility guarantees.

ui/tests/e2e/login.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    'should login fail by incorrect password',
+    async () => {
+      const page = await browser.newPage()
+      await page.goto(LOGIN_URL)
+
+      const pwdInput = await page.waitForSelector('input#tidb_signin_password')
+      await pwdInput.type('any')
+      await page.click('button#signin_btn')
+
+      const errorContainer = await page.waitForSelector('div.has-error')
```
breeswish: ditto. This test heavily rely on the implementation of ant design itself, which is not good.

baurine: Tried to add "id" to the component but some places work, some places don't, the "id" can't pass to real dom node, it needs to change the original code much, I think we can do that later.

ui/tests/e2e/test_config.ts
baurine: done!

baurine:
Create an issue https://github.com/pingcap-incubator/tidb-dashboard/issues/287 to refine the subsequent work, will do it this week if have time.

ui/tests/e2e/login.test.ts
baurine: > Remember to open subsequent PRs to drop the code that heavily rely on ant design.

Create an issue https://github.com/pingcap-incubator/tidb-dashboard/issues/287 to refine the subsequent work, will do it this week if have time.

baurine:
> Remember to open subsequent PRs to drop the code that heavily rely on ant design.

Create an issue https://github.com/pingcap-incubator/tidb-dashboard/issues/287 to refine the subsequent work, will do it this week if have time.

PR #293, opened 2020-03-26
ui: Use DetailsTable in FabricUI

breeswish:
This PR adds a new Component `CardTableV2`, which uses [DetailsTable of FabricUI](https://developer.microsoft.com/en-us/fabric#/controls/web/detailslist) internally to provide a better table experience, as a substitution of the antd's Table.

Unique features provided by FabricUI DetailsTable includes:
- **Column resize**
  image

- Collapse extra columns automatically

  No need to worry about displaying too many columns causing important columns too short:

  image

  In the case above time and duration columns are hidden automatically because of small screen.

- Sticky header

  image

- Better grouping
- Reorder columns (drag & drop)

Currently CardTableV2 is placed in `@/components` since there is a bug that extra CSSs are not resolved in `packages/dashboard_components`.

The FabricUI's Font and Icon is licensed and limited to developing MS's Apps. Thus no default fonts and icons are introduced (see `FabricRoot`).


ui/src/react-app-env.d.ts
```diff
@@ -1 +1,21 @@
...
+
+declare module '*.module.scss' {
+  const classes: { readonly [key: string]: string }
+  export default classes
+}
+
+declare module '*.module.sass' {
+  const classes: { readonly [key: string]: string }
+  export default classes
+}
```
baurine: Maybe we don't need these 2 temporarily.

PR #297, opened 2020-03-27
replace sql-formatter with sql-formatter-plus

qxhy123:


breeswish:
Hi, could you rformat your code by using `yarn format`? Thanks!

qxhy123:
> Hi, could you rformat your code by using `yarn format`? Thanks!

yea, just did that

qxhy123:
here is the screenshot:
image

PR #298, opened 2020-03-27
ui: reduce bundle size

baurine:
- only import sql syntax highlight instead of importing all languages to reduce bundle size

This can help reduce the highlight.js bundle size from ~900KB to ~30KB.

Ref: https://www.npmjs.com/package/react-syntax-highlighter#light-build

### DIFF
--- a/ui/packages/statement/src/components/StatementSummaryTable.tsx
+++ b/ui/packages/statement/src/components/StatementSummaryTable.tsx
@@ -1,11 +1,14 @@
 import React from 'react'
 import { Table } from 'antd'
 import moment from 'moment'
-import { StatementDetailInfo } from './statement-types'
 import { useTranslation } from 'react-i18next'
-import SyntaxHighlighter from 'react-syntax-highlighter'
-import { atomOneLight } from 'react-syntax-highlighter/dist/esm/styles/hljs'
 import sqlFormatter from 'sql-formatter-plus'
+import { Light as SyntaxHighlighter } from 'react-syntax-highlighter'
+import sql from 'react-syntax-highlighter/dist/esm/languages/hljs/sql'
+import { atomOneLight } from 'react-syntax-highlighter/dist/esm/styles/hljs'
+import { StatementDetailInfo } from './statement-types'
+
+SyntaxHighlighter.registerLanguage('sql', sql)
 
 type align = 'left' | 'right' | 'center'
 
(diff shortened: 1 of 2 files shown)


ui/packages/statement/src/components/StatementSummaryTable.tsx
```diff
@@ -1,11 +1,14 @@
...
 import { Table } from 'antd'
 import moment from 'moment'
-import { StatementDetailInfo } from './statement-types'
 import { useTranslation } from 'react-i18next'
-import SyntaxHighlighter from 'react-syntax-highlighter'
-import { atomOneLight } from 'react-syntax-highlighter/dist/esm/styles/hljs'
 import sqlFormatter from 'sql-formatter-plus'
+import { Light as SyntaxHighlighter } from 'react-syntax-highlighter'
+import sql from 'react-syntax-highlighter/dist/esm/languages/hljs/sql'
+import { atomOneLight } from 'react-syntax-highlighter/dist/esm/styles/hljs'
```
breeswish: maybe `import 'react-syntax-highlighter/dist/esm/styles/hljs/atomOneLight'` is better?

ui/packages/statement/src/components/StatementSummaryTable.tsx
```diff
@@ -1,11 +1,14 @@
...
+import { StatementDetailInfo } from './statement-types'
+
+SyntaxHighlighter.registerLanguage('sql', sql)
```
breeswish: maybe we need to do it in the entry file to avoid registering multiple times when it is used in somewhere else in future?

baurine: yep, I thought that, now we only have this one place need it, if we have mulitple places need it, we need to extract.

ui/packages/statement/src/components/StatementSummaryTable.tsx
baurine: `<SyntaxHighlighter language="sql" style={atomOneLight}>` this place need it.

baurine: but I can try whether we can do that then we don't declare style for SyntaxHighlighter

breeswish: how about `import atomOneLight from 'react-syntax-highlighter/dist/esm/styles/hljs/atomOneLight'`

breeswish: using `import { xxx }` will include all styles I believe.

baurine: make sense, let me have a try.

baurine: The 2 ways have the same bundle size after trying.

PR #299, opened 2020-03-27
ui:  support customize range duration for diagnosis

baurine:
- support to customize the range duration for generating diagnosis report

screenshots:

image


baurine:
duplicated with #300 , close it.

PR #300, opened 2020-03-27
ui: Customize diagnose duration

breeswish:

### DIFF
--- a/ui/src/apps/diagnose/components/DiagnoseGenerator.tsx
+++ b/ui/src/apps/diagnose/components/DiagnoseGenerator.tsx
@@ -1,5 +1,14 @@
 import React from 'react'
-import { Button, DatePicker, Form, Select, Switch, message } from 'antd'
+import {
+  Button,
+  DatePicker,
+  Form,
+  Select,
+  Switch,
+  Input,
+  InputNumber,
+  message,
+} from 'antd'
 import { useTranslation } from 'react-i18next'
 import { Card } from '@pingcap-incubator/dashboard_components'
 import { useHistory } from 'react-router-dom'
@@ -8,7 +17,10 @@ import client from '@pingcap-incubator/dashboard_client'
 const useFinishHandler = (history) => {
   return async (fieldsValue) => {
     const start_time = fieldsValue['rangeBegin'].unix()
-    const range_duration = fieldsValue['rangeDuration']
+    let range_duration = fieldsValue['rangeDuration']
+    if (fieldsValue['rangeDuration'] === 0) {
+      range_duration = fieldsValue['rangeDurationCustom']
+    }
     const is_compare = fieldsValue['isCompare']
     const compare_range_begin = fieldsValue['compareRangeBegin']
 
@@ -43,7 +55,7 @@ export default function DiagnoseGenerator() {
         layout="vertical"
         style={{ minWidth: 500 }}
         onFinish={handleFinish}
-        initialValues={{ rangeDuration: 10 }}
+        initialValues={{ rangeDuration: 10, rangeDurationCustom: 10 }}
       >
         <Form.Item
           name="rangeBegin"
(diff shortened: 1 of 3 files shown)


ui/src/apps/diagnose/components/DiagnoseGenerator.tsx
```diff
@@ -52,18 +64,49 @@ export default function DiagnoseGenerator() {
...
-          label={t('diagnose.generate.range_duration')}
-        >
-          <Select style={{ width: 120 }}>
-            <Select.Option value={5}>5 min</Select.Option>
-            <Select.Option value={10}>10 min</Select.Option>
-            <Select.Option value={30}>30 min</Select.Option>
-            <Select.Option value={60}>1 hour</Select.Option>
-            <Select.Option value={24 * 60}>1 day</Select.Option>
-          </Select>
+        <Form.Item label={t('diagnose.generate.range_duration')}>
```
baurine: We can `required` props so the asterisk won't miss.

ui/src/apps/diagnose/components/DiagnoseGenerator.tsx
```diff
@@ -52,18 +64,49 @@ export default function DiagnoseGenerator() {
...
+              rules={[{ required: true }]}
+              noStyle
+            >
+              <Select style={{ width: 120 }}>
+                <Select.Option value={5}>5 min</Select.Option>
+                <Select.Option value={10}>10 min</Select.Option>
+                <Select.Option value={30}>30 min</Select.Option>
+                <Select.Option value={60}>1 hour</Select.Option>
+                <Select.Option value={24 * 60}>1 day</Select.Option>
+                <Select.Option value={0}>Custom</Select.Option>
```
baurine: Can we support i18n here?

baurine:
the rest LGTM!

ui/src/apps/diagnose/components/DiagnoseGenerator.tsx
```diff
@@ -52,18 +64,49 @@ export default function DiagnoseGenerator() {
...
+                  getFieldValue('rangeDuration') === 0 && (
+                    <Form.Item
+                      noStyle
+                      name="rangeDurationCustom"
+                      rules={[{ required: true }]}
+                    >
+                      <InputNumber
+                        min={1}
+                        max={43200}
+                        formatter={(value) => `${value} min`}
```
baurine: Can we support i18n here as well?

baurine:
btw, I think maybe we should change `Range Size` to `Range Duration`, `Range Size` to represent the time length sounds weird...

ui/src/apps/diagnose/components/DiagnoseGenerator.tsx
breeswish: I guess it should be 5min (consider other abbreviations 5km, 5kg, etc)

ui/src/apps/diagnose/components/DiagnoseGenerator.tsx
breeswish: It doesn't work. The star does not appear. Maybe it is because there is no values inside.

breeswish:
> btw, I think maybe we should change `Range Size` to `Range Duration`, `Range Size` to represent the time length sounds weird...

I don't like "Size" as well, however "duration" will give false information that it will diagnose for this long time.

ui/src/apps/diagnose/components/DiagnoseGenerator.tsx
baurine: I mean to use `required` as props, likes `<Form.Item required .../>`, I tested and it works:

image

ui/src/apps/diagnose/components/DiagnoseGenerator.tsx
baurine: er, make sense.

baurine: Let me add i18n support soon.

baurine:
image

image

Issue #249, opened 2020-03-19
UCP: Update to React Router v6


breeswish:
## Description

React Router v6 is more easier to use. We can update to this version.

## Score

* 100

## Mentor(s)

* @baurine 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* React

PR #301, opened 2020-03-30
update to react-router 6

Fullstop000:
UCP https://github.com/pingcap-incubator/tidb-dashboard/issues/249

Update to react-router6

### DIFF
--- a/ui/packages/statement/src/RootComponent.tsx
+++ b/ui/packages/statement/src/RootComponent.tsx
@@ -1,10 +1,10 @@
-import React, { useState } from 'react'
+import React, { useState, useEffect } from 'react'
 import {
   HashRouter as Router,
   Routes,
   Route,
-  Redirect,
   Link,
+  useNavigate,
   useLocation,
 } from 'react-router-dom'
 import { Breadcrumb } from 'antd'
@@ -15,7 +15,7 @@ import { SearchContext, SearchOptions } from './components'
 import { StatementsOverviewPage, StatementDetailPage } from './pages'
 
 const App = (props) => {
-  // const { location } = props
+  const navigate = useNavigate()
   const location = useLocation()
   const page = location.pathname.split('/').pop()
 
@@ -25,7 +25,9 @@ const App = (props) => {
     curTimeRange: undefined,
   } as SearchOptions)
   const searchContext = { searchOptions, setSearchOptions }
-
+  useEffect(() => {
+    navigate('/statement/overview')
+  }, [])
   return (
     <SearchContext.Provider value={searchContext}>
       <div>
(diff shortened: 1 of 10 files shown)


baurine:
hi @Fullstop000 , thanks for your contribution, I will test and review soon.

baurine:
Hi @Fullstop000 , after my test,  it have some errors, statements / diagnose / search logs / profile pages can't show normally, they are all empty, and the console output following errors:

```
Uncaught Error: Route(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.
```

image

ui/packages/statement/src/RootComponent.tsx
```diff
@@ -39,7 +40,7 @@ const App = withRouter((props) => {
           </Breadcrumb>
         </div>
         <div style={{ margin: 12 }}>
-          <Switch>
+          <Routes>
             <Route path="/statement/overview">
               <StatementsOverviewPage
                 dashboardClient={client.getInstance()}
```
baurine: After my test, nest the component inside the `<Route>` doesn't work in v6, we need to make the component as `element` prop value, like this: `<Route path="/statement/overview" element={<StatementsOverviewPage .../>}`

Fullstop000:
Got something pained after upgrading to react-router v6
image

baurine:
> Got something pained after upgrading to react-router v6
> image

yep, that's a problem, we need to have a look.

Fullstop000:
And I got this warning only when unmount `AppStatement`
image

baurine:
> And I got this warning only when unmount `AppStatement`
> image

I will have a look today.

baurine:
There is some discussion in https://github.com/facebook/react/issues/18178 , I tried to upgrade the  react to 16.13.1 , it doesn't dismiss the warning.

It should be caused by the new version react-router, above warning doesn't happen for react-router v5, someone has already reported in https://github.com/ReactTraining/react-router/issues/7199 as well.

So I think let's wait a few days more to see whether react-router will fix it, if it doesn't, we can merge it first and upgrade it in the future.

Fullstop000:
> So I think let's wait a few days more to see whether react-router will fix it, if it doesn't, we can merge it first and upgrade it in the future.

Sounds reasonable :)

baurine:
hi  @Fullstop000 , react-router fixed above warning in https://github.com/ReactTraining/react-router/releases/tag/v6.0.0-alpha.3 , you can have a try, but unfortunately they removed `<Redirect/>` component as well which we used, not sure whether it is easy to migrate, if it doesn't, I suggest let update react-router v6 until it is stable, @breeswish , how do you think about it?

Fullstop000:
Maybe using a history API instead `<Redirect/>` ? I'll have a try later today.

breeswish:
@baurine I'm fine with both. This task is not ergant.

Fullstop000:
@baurine The second warning has been fixed and the first warning only shows on the initial render.

Issue #131, opened 2020-03-02
UCP: Support TiDB retry


breeswish:
## Description

Now we have a skeleton `tidb.Forwarder`, which connects directly to the first TiDB entry. However the first TiDB entry can be broken and we need to retry other TiDB nodes in the TiDB list.

The methods can be as follows:

1. Implement real health check and forwarding in forwarder. 

  - The forwarder need to support forwarding both MySQL TCP connection (which normally uses port 4000) and TiDB Status HTTP connection (which normally uses port 10080). You can retrieve both ports from tidb info entry.

  - The forwarder listens two port for forwarding MySQL TCP connection and TiDB Status HTTP connection in separate. The two ports are allocated dynamically depends on current available ports in the host and should be restricted to loopback interface (127.0.0.1) only.

  - The forwarder should check the health for each TiDB node (both 4000 and 10080) in background periodically, as well as switch to try other connection on demand when connect failed.

2. Change Dashboard's TiDB MySQL connection to connect to the forwarder's port.

3. Change Dashboard's TiDB Status connection to connect to the forwarder's port.

## Score 

* 1680

## Mentor(s)

* @breeswish 
* @HunDunDM 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Go TCP Programming

Fullstop000:
@you06 How many scores do we (.* team) have in this project actually 😅

you06:
@Fullstop000 I find #301 only which awarded your team 100 points, did I miss any PRs?

Fullstop000:
@you06 just to make sure sre-bot is right 🤣.

Fullstop000:
@you06 https://github.com/pingcap-incubator/tidb-dashboard/pull/304 has been merged.

you06:
@Fullstop000 Fixed.

you06:
> 
> 
> This pick has been automatically canceled after more than a week.

Caused by some mistakes, I think this will not take effect to the score, ignore it.

PR #304, opened 2020-03-30
Add TCP load balancer and TiDB cluster info poller

Fullstop000:
UCP https://github.com/pingcap-incubator/tidb-dashboard/issues/131


pkg/utils/tcp/proxy.go
```diff
@@ -0,0 +1,217 @@
...
+		return err
+	}
+	conn.Close()
+	r.mu.Lock()
+	r.inactive = false
+	r.mu.Unlock()
+	return nil
+}
+
+type Proxy struct {
```
breeswish, line 44: actually this is a forwarder

pkg/utils/tcp/proxy.go
```diff
@@ -0,0 +1,217 @@
...
+	l             net.Listener
+	checkInterval time.Duration
+	dialTimeout   time.Duration
+	donec         chan struct{}
+	errc          chan error
+
+	mu        sync.Mutex
+	remotes   map[string]*remote
```
breeswish, line 52: how about using sync.Map to avoid these mutex?

pkg/utils/tcp/proxy.go
```diff
@@ -0,0 +1,217 @@
...
+func (p *Proxy) serve(in net.Conn) {
+	var (
+		err    error
+		out    net.Conn
+		picked *remote
+	)
+	for {
+		p.mu.Lock()
+		picked = p.pick()
+		p.mu.Unlock()
```
breeswish, line 120: some mutexs are placed in the function definition while some are placed at call site, prefer to unify them.

pkg/utils/tcp/proxy.go
```diff
@@ -0,0 +1,217 @@
...
+	remotes := make(map[string]*remote)
+	for key, e := range endpoints {
+		remotes[key] = &remote{
+			addr:     e,
+			inactive: true,
+		}
+	}
+	return &Proxy{
+		l:             l,
+		errc:          make(chan error),
```
breeswish, line 72: prefer better namings, like `listener`, `errCh`, `doneCh`, etc.

pkg/utils/tcp/proxy.go
```diff
@@ -0,0 +1,217 @@
...
+	r.mu.Unlock()
+	return nil
+}
+
+type Proxy struct {
+	l             net.Listener
+	checkInterval time.Duration
+	dialTimeout   time.Duration
+	donec         chan struct{}
+	errc          chan error
```
breeswish, line 49: looks like there is no where using `errc`

pkg/tidb/forwarder.go
```diff
@@ -51,23 +61,57 @@ func NewForwarderConfig(cfg *config.Config) *ForwarderConfig {
...
+	statusPort   int
+	donec        chan struct{}
+	pollInterval time.Duration
 }
 
-// TODO: Requires load balancing and health checks.
 func (f *Forwarder) Open() error {
-	// Currently this function does nothing.
+	cluster, err := f.getServerInfo()
+	if err != nil {
```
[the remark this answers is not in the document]
Fullstop000: Use an polling style goroutine with retrying count limit instead ?

pkg/utils/tcp/manager.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	l, err := net.ListenTCP("tcp", addr)
+	if err != nil {
+		return 0, err
+	}
+	defer l.Close()
+	return l.Addr().(*net.TCPAddr).Port, nil
```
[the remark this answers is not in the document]
Fullstop000: Is there a proper way to avoid this?

pkg/utils/tcp/proxy.go
```diff
@@ -0,0 +1,217 @@
...
+	remotes   map[string]*remote
+	pickCount int // for RoundRobin count
+}
+
+func NewProxy(l net.Listener, endpoints map[string]string, checkInterval time.Duration, timeout time.Duration) *Proxy {
+	if checkInterval == 0 {
+		checkInterval = 5 * time.Second
+	}
+	if timeout == 0 {
+		timeout = 10 * time.Second
```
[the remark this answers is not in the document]
Fullstop000: Use 3s instead

pkg/utils/tcp/proxy.go
```diff
@@ -0,0 +1,217 @@
...
+	if out == nil {
+		log.Warn("no alive remote, drop incoming conn")
+		// Do we need issue a error here?
+		in.Close()
+		return
+	}
+	// bidirectional copy
+	go func() {
+		//nolint
+		io.Copy(in, out)
```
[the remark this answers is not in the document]
Fullstop000: Since it's a bi-directional copy, I'm not sure how to handle IO error.

pkg/utils/tcp/proxy.go
```diff
@@ -0,0 +1,217 @@
...
+type Proxy struct {
+	l             net.Listener
+	checkInterval time.Duration
+	dialTimeout   time.Duration
+	donec         chan struct{}
+	errc          chan error
+
+	mu        sync.Mutex
+	remotes   map[string]*remote
+	pickCount int // for RoundRobin count
```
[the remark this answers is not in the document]
Fullstop000: In the current implementation, the underlying incoming connection will be re-used unless it is closed manually.

pkg/utils/tcp/proxy.go
Fullstop000: Proxy is ok I think 😁

pkg/keyvisual/decorator/tidb.go
```diff
@@ -76,7 +79,7 @@ func (s *tidbLabelStrategy) Background(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			s.updateAddress(ctx)
+			//s.updateAddress(ctx)
```
Deardrops, line 82: remove this comment?

pkg/keyvisual/decorator/tidb_requests.go
```diff
@@ -117,12 +77,13 @@ func (s *tidbLabelStrategy) updateMap(ctx context.Context) {
...
+	hostname, port := s.forwarder.GetStatusConnProps()
+	target := fmt.Sprintf("%s:%d", hostname, port)
+	reqEndpoint := fmt.Sprintf("%s://%s", reqScheme, target)
+	if err := request(reqEndpoint, "schema", &dbInfos, s.HTTPClient); err == nil {
+		tidbEndpoint = reqEndpoint
+	} else {
+		log.Error("fail to send schema reqeust to tidb", zap.String("endpoint", reqEndpoint), zap.Error(err))
 	}
 	if dbInfos == nil {
 		return
```
Deardrops, line 89:
```suggestion
	if err := request(reqEndpoint, "schema", &dbInfos, s.HTTPClient); err != nil {
		log.Error("fail to send schema reqeust to tidb", zap.String("endpoint", reqEndpoint), zap.Error(err))
		if dbInfos == nil {
			return
		}
	}
	tidbEndpoint = reqEndpoint
```

pkg/tidb/conn.go
```diff
@@ -34,23 +34,19 @@ const (
...
 )
 
-func (f *Forwarder) getDBConnProps() (host string, port int, err error) {
-	info, err := f.getServerInfo()
-	if err == nil {
-		host = info.IP
-		port = info.Port
-	}
-	return
+func (f *Forwarder) getDBConnProps() (host string, port int) {
```
Deardrops, line 37:
```suggestion
func (f *Forwarder) getDBConnProps() (string, int) {
```

pkg/tidb/conn.go
```diff
@@ -34,23 +34,19 @@ const (
...
+	return "127.0.0.1", f.tidbPort
+}
+
+func (f *Forwarder) GetStatusConnProps() (host string, port int) {
```
Deardrops, line 41: ditto

pkg/tidb/conn.go
```diff
@@ -34,23 +34,19 @@ const (
...
 
-func (f *Forwarder) getDBConnProps() (host string, port int, err error) {
-	info, err := f.getServerInfo()
-	if err == nil {
-		host = info.IP
-		port = info.Port
-	}
-	return
+func (f *Forwarder) getDBConnProps() (host string, port int) {
+	return "127.0.0.1", f.tidbPort
```
Deardrops, line 38: Does this work well in IPv6 envelopment?

pkg/tidb/forwarder.go
```diff
@@ -17,16 +17,28 @@ import (
 	"context"
 	"crypto/tls"
 	"encoding/json"
+	"fmt"
+	"github.com/cenkalti/backoff/v4"
+	"github.com/joomcode/errorx"
```
Deardrops, line 22: Please rearrange the import.

pkg/tidb/forwarder.go
```diff
@@ -47,27 +60,63 @@ func NewForwarderConfig(cfg *config.Config) *ForwarderConfig {
...
+	for _, server := range cluster {
+		statusEndpoints[server.ID] = fmt.Sprintf("%s:%d", server.IP, server.StatusPort)
+		tidbEndpoints[server.ID] = fmt.Sprintf("%s:%d", server.IP, server.Port)
+	}
+	pr, err := f.createProxy(tidbProxyLabel, tidbEndpoints)
+	if err != nil {
+		return err
+	}
+	go pr.Run()
+	f.tidbPort = pr.port()
```
Deardrops, line 95:
```suggestion
	f.tidbPort = pr.port()
	go pr.Run()
```

pkg/tidb/forwarder.go
```diff
@@ -47,27 +60,63 @@ func NewForwarderConfig(cfg *config.Config) *ForwarderConfig {
...
+	if err != nil {
+		return err
+	}
+	f.statusPort = pr.port()
+	go pr.Run()
+	go f.pollingForTiDB()
 	return nil
 }
 
 func (f *Forwarder) Close() error {
```
Deardrops, line 106: Does this returned error is necessary here?

pkg/tidb/forwarder.go
```diff
@@ -76,23 +125,97 @@ func (f *Forwarder) getServerInfo() (*tidbServerInfo, error) {
...
 
-	return nil, ErrNoAliveTiDB.New("no TiDB is alive")
+	return allTiDB, nil
+}
+
+func (f *Forwarder) createProxy(key string, endpoints map[string]string) (*Proxy, error) {
+	if len(endpoints) == 0 {
+		return nil, fmt.Errorf("empty endpoints")
+	}
+	port, err := getFreePort()
```
Deardrops, line 148: Why not bind an available port by `net.ListenTCP("tcp", "127.0.0.1:0")` directly?

pkg/tidb/forwarder.go
```diff
@@ -76,23 +125,97 @@ func (f *Forwarder) getServerInfo() (*tidbServerInfo, error) {
...
 func NewForwarder(lc fx.Lifecycle, config *ForwarderConfig, etcdClient *clientv3.Client) *Forwarder {
 	f := &Forwarder{
-		etcdClient: etcdClient,
-		config:     config,
-		ctx:        context.Background(),
+		etcdClient:   etcdClient,
+		config:       config,
+		ctx:          context.Background(),
+		donec:        make(chan struct{}),
+		pollInterval: 5 * time.Second, // TODO: add this into config?
```
Deardrops, line 218: @breeswish PTAL, in my view, add this to config is OK.

pkg/tidb/proxy.go
```diff
@@ -0,0 +1,215 @@
...
+	r := p.pickCount % len(activeRemotes)
+	p.pickCount++
+	return activeRemotes[r]
+}
+
+func (p *Proxy) doCheck() {
+	for {
+		select {
+		case <-time.After(p.checkInterval):
+			//p.mu.Lock()
```
Deardrops, line 164: remove this comment.

pkg/tidb/proxy_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestProxy(t *testing.T) {
+	l, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err)
+	}
```
Deardrops, line 21:
```
defer l.Close()
```

pkg/tidb/proxy_test.go
```diff
@@ -0,0 +1,96 @@
...
+	assert.Equal(t, want, string(got))
+}
+
+func TestProxyPick(t *testing.T) {
+	l, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err)
+	}
```
Deardrops, line 54:
```
defer l.Close()
```

pkg/keyvisual/decorator/tidb_requests.go
Fullstop000: Seems the github suggested change is out of format ( such a poor feature :(

pkg/tidb/forwarder.go
Fullstop000: Involve the resolver

pkg/tidb/forwarder.go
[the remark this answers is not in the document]
breeswish: we can retry infinitely, with a maximum retry interval (like 5s / retry)

pkg/utils/tcp/manager.go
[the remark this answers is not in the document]
breeswish: maybe https://stackoverflow.com/questions/43424787/how-to-use-next-available-port-in-http-listenandserve

pkg/utils/tcp/proxy.go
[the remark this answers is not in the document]
breeswish: Dashboard establish a SQL connection for each request (and close it after the response is finished)

pkg/utils/tcp/proxy.go
breeswish: Then maybe you can remove the "forwarder", since it was created exactly for this purpose (but without any implementation XD)

pkg/tidb/conn.go
Fullstop000: Not sure about it. Is there any proper way to test this out?

Fullstop000:
@Deardrops @breeswish Almost addressed all the comments (except removing `proxy.go` :P) PTAL

pkg/tidb/conn.go
Deardrops: @Fullstop000 Start TiDB with parameter:

```bash
./bin/tidb-server -host=[::1]
# or ./bin/tidb-server -host=\[::1\]
```

Then it only listens on ipv6 address.

Deardrops: @breeswish PTAL, Shall we handle ipv6 here?

breeswish: According to the doc if you use `tcp4` as listener then only IPv4 address is listened. When using `tcp`, both IPv4 and IPv6 are listened.

pkg/keyvisual/decorator/tidb_requests.go
```diff
@@ -117,16 +77,16 @@ func (s *tidbLabelStrategy) updateMap(ctx context.Context) {
 	if s.Config.ClusterTLSConfig != nil {
 		reqScheme = "https"
 	}
-	for _, addr := range s.TidbAddress {
-		reqEndpoint := fmt.Sprintf("%s://%s", reqScheme, addr)
-		if err := request(reqEndpoint, "schema", &dbInfos, s.HTTPClient); err == nil {
-			tidbEndpoint = reqEndpoint
-			break
+	hostname, port := s.forwarder.GetStatusConnProps()
```
breeswish, line 80: @HunDunDM PTAL about this part

pkg/tidb/forwarder.go
```diff
@@ -47,27 +60,71 @@ func NewForwarderConfig(cfg *config.Config) *ForwarderConfig {
...
+	)
+	for {
+		cluster, err = f.getServerInfo()
+		if err == nil {
+			break
+		}
+		log.Error("Fail to get server info", zap.Error(err))
+		// TODO: config this or just use CheckInterval ?
+		<-time.After(time.Second * 5)
+	}
```
breeswish, line 92: In the case when TiDB / PD is not started, this function is trapped in the loop, so that it will block other constructors. I suspect this behaviour will cause many bugs (in other places)..

pkg/tidb/proxy.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+type proxy struct {
+	listener      net.Listener
+	checkInterval time.Duration
+	dialTimeout   time.Duration
+	doneCh        chan struct{}
+
+	// The key is server ddl id
+	remotes sync.Map
```
breeswish, line 45: I think it's better to not couple the proxy logic with any TiDB logic. Otherwise this proxy cannot be reused when we want to proxy more things. Can we simply use IpAddr as key?

pkg/tidb/forwarder.go
Fullstop000: What about using a goroutine to serve the codes above and add a flag to indicate whether the forwarder is ready? And if the forwarder is not ready, then any requests to the forwarder should be rejected.

breeswish: LGTM. Also what will happen if after some time all TiDBs are down? Will the `127.0.0.1:proxyPort` connection being blocked and waited, or being rejected instantly (the same behaviour as you described)?

Fullstop000: Under this circumstance, no `remote` will be picked on serving every connection and the request shall be rejected instantly.

pkg/tidb/forwarder.go
```diff
@@ -76,31 +133,103 @@ func (f *Forwarder) getServerInfo() (*tidbServerInfo, error) {
...
+		select {
+		case <-time.After(f.pollInterval):
+			var allTiDB []*tidbServerInfo
+			var err error
+			err = backoff.Retry(func() error {
+				allTiDB, err = f.getServerInfo()
+				return err
+			}, backoff.NewExponentialBackOff())
+			if err != nil {
+				if errorx.IsOfType(err, ErrNoAliveTiDB) {
```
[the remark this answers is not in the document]
Fullstop000: As a default, the max trying time is 15minutes before calling `Stop()`

pkg/tidb/forwarder.go
Fullstop000: I found out that I can just use `pollingForTiDB` directly...

pkg/tidb/forwarder.go
```diff
@@ -76,31 +133,103 @@ func (f *Forwarder) getServerInfo() (*tidbServerInfo, error) {
...
+func (f *Forwarder) pollingForTiDB() {
+	for {
+		select {
+		case <-time.After(f.pollInterval):
+			var allTiDB []*tidbServerInfo
+			var err error
+			err = backoff.Retry(func() error {
+				allTiDB, err = f.getServerInfo()
+				return err
+			}, backoff.NewExponentialBackOff())
```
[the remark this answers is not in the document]
Fullstop000: Use 5s as MaxInterval time for now

pkg/tidb/forwarder.go
```diff
@@ -76,31 +104,86 @@ func (f *Forwarder) getServerInfo() (*tidbServerInfo, error) {
...
+		allTiDB = append(allTiDB, info)
+	}
+	if len(allTiDB) == 0 {
+		return nil, ErrNoAliveTiDB.New("no TiDB is alive")
+	}
+
+	return allTiDB, nil
+}
+
+func (f *Forwarder) createProxy(endpoints map[string]string) (*proxy, error) {
```
breeswish, line 123: This function may be refined now (to not accepting `endpoints`)

pkg/tidb/proxy.go
```diff
@@ -0,0 +1,228 @@
...
+func (r *remote) isActive() bool {
+	return !r.inactive.Load()
+}
+
+func (r *remote) becomeInactive() {
+	r.inactive.Store(true)
+	log.Debug("remote become inactive", zap.String("remote", r.addr))
+}
+
+func (r *remote) dial(timeout time.Duration) error {
```
breeswish, line 28: maybe we can name it `checkAlive` / `checkActive` (since it dials and closes)

pkg/keyvisual/decorator/tidb_requests.go
HunDunDM: LGTM

PR #306, opened 2020-03-31
logs: fix ui bug when download button not be disabled

Deardrops:


breeswish:
I think we can also remount the page when search id changes, which will prevent similar issue more thoroughly.

Deardrops:
Closed, this bug fixed in #380

PR #307, opened 2020-03-31
Support i18n for diagnosis report

baurine:
What did:

- Render diagnosis report by frontend instead of backend
- Support i18n for diagnosis report

Screenshots:

image


ui/packages/diagnosis_report/src/types.ts
```diff
@@ -0,0 +1,21 @@
...
+  ratio: number
+  Comment: string
+}
+
+export interface TableDef {
+  Category: string[]
+  Title: string
+  CommentEN: string
+  CommentCN: string
+  joinColumns: number[]
```
breeswish, line 15: why some are UpperCased while some are lowerCased?

ui/packages/diagnosis_report/src/utils/i18n.js
```diff
@@ -0,0 +1,35 @@
+import i18next from 'i18next'
```
breeswish, line 1: We should avoid duplicating this utilities

ui/packages/diagnosis_report/scripts/watch.js
```diff
@@ -0,0 +1,53 @@
+// https://gist.github.com/int128/e0cdec598c5b3db728ff35758abdbafd
+process.env.NODE_ENV = 'development'
```
breeswish, line 2: Now there are 3 build systems..1 is CRA, 1 is rollup, 1 is this watch file. It is horrible for such a simple project that there is actually only one App targeting at users.

baurine:
Finally, embed the assets to go code:

ui/packages/diagnosis_report/src/types.ts
baurine: just copied from go code ...

ui/packages/diagnosis_report/src/utils/i18n.js
baurine: yep, I agree, don't know how to do yet.

ui/packages/diagnosis_report/scripts/watch.js
baurine: yes it is! it is horrible for me as well... do you have any good ideas? but I think the diagnosis report is an independent app with the dashboard ui.

pkg/apiserver/diagnose/assets_dev.go
```diff
@@ -0,0 +1,10 @@
+//+build dev
+
+package diagnose
+
+import (
+	"net/http"
+)
+
+// var Vfs http.FileSystem = http.Dir("ui/packages/diagnosis_report/build")
+var Vfs http.FileSystem = http.Dir("/Users/baurine/Codes/Work/pingcap-incubator/tidb-dashboard/ui/packages/diagnosis_report/build")
```
baurine, line 10: To fix, relative path can't work when run `go generate pkg/apiserver/diagnose/assets_embed.go`, it reports can't find "ui/packages/diagnosis_report/build".

baurine: fixed it, get the absolute path by os.Getwd() method

baurine:
Hi @breeswish , almost done, PTAL, thanks!

I will continue to resolve the conflicts.

ui/packages/diagnosis_report/src/types.ts
breeswish: @crazycs520 We should fix it..

ui/packages/diagnosis_report/src/utils/i18n.js
breeswish: Can we extract i18n to a standalone package?

pkg/apiserver/diagnose/assets_dev.go
```diff
@@ -0,0 +1,31 @@
...
+	if err != nil {
+		log.Println(err)
+	}
+	var assetsPath string
+	if strings.HasSuffix(path, "tidb-dashboard") {
+		// dev mode
+		assetsPath = path + "/ui/packages/diagnosis_report/build"
+	} else if strings.HasSuffix(path, "diagnose") {
+		// work with `go generate ./pkg/apiserver/diagnose` in the project root folder
+		assetsPath = path + "/../../../ui/packages/diagnosis_report/build"
```
breeswish, line 23: There will be no issues if you use webpack-dev-server to serve this page

ui/packages/diagnosis_report/src/utils/i18n.js
baurine: yep, that's a way, we can do it later.

pkg/apiserver/diagnose/assets_dev.go
baurine: 😂in fact I tried this,  the biggest problem is the homepage value in the package.json won't add to the index.html by webpack-dev-server, it brings more changes for other files. I tried kinds of solutions, and it seems this solution changes code least (maybe I am too weak...)

pkg/utils/html_template.go
```diff
@@ -14,15 +14,29 @@
...
+func NewHTMLRender(templ *template.Template, infos []TemplateInfo) render.HTMLRender {
+	for _, info := range infos {
+		t := templ.New(info.Name)
+		if _, err := t.Parse(info.Text); err != nil {
+			log.Fatal("Failed to parse template", zap.String("name", info.Name), zap.Error(err))
+		}
+	}
+	return render.HTMLProduction{Template: templ}
+}
+
```
baurine, line 39: These contents are added by when resolving conflicts, in fact this whole file can be removed, because the go doesn't need any code about template, but I intend to handle these legacy codes next time.

ui/packages/diagnosis_report/scripts/watch.js
baurine: if creat-react-app can add one more option for react-scripts to enable webpack watch mode, we don't need this file at all, the discussion is here: https://github.com/facebookincubator/create-react-app/issues/1070 , I have viewed all comments and tried some methods someone supplied, and this should be the easiest and works one.

I also tried to change the webpack config in the config-overrides.js manually, change `config.watch = true` but unfortunately, it doesn't work ...

baurine:
Resolved conflicts!

baurine:
Replaced by #331 , close it.

Issue #88, opened 2020-02-28
UCP: Support super user login without TiDB


breeswish:
## Description

Now TiDB Dashboard can only be logged in with existing TiDB node. However for pure TiKV deployments there is no TiDB node, makes user unable to use TiDB Dashboard.

For such scenario, we need to provide alternative login method: supply a super user login username and password. The username and hash of the password is stored in PD etcd so that multiple PD nodes can share the same authenticator. 

Even if there is a TiDB node in the cluster, when log in use super user, TiDB related functionalities like Statements should not be available (since they requires a valid TiDB connection).

## Score 

* 4007

## Mentor(s)

* @breeswish

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Security
* etcd
* Go web programming

SSebo:
@breeswish 
- How to distinguish the two login mode? 
- Is TiKV mode password changeable?

breeswish:
@SSebo 

- User should be able to choose the login mode at UI. Currently there is already a field for this in the API: `is_tidb_auth`.
- Yes. User should be able to update the password in UI after a successful sign in.

SSebo:
@breeswish 
- need a way to set initial password like in some config file ?
- need a etcd path to store this hashed password, then fetched by golang backend
- in what situation user will login without TiDB, and there is a tidb instance in cluster? Is it mean that if user login with this super password, the ability to use TiDB related functionalities is disabled? no matter there is TiDB node or not ?

breeswish:
Hi @SSebo 

> How to set initial password? like in some config file ? or some constant string?

There should be no initial account (and no initial password). i.e. by default user is not allowed to sign in use TiKV mode.

The initial account and password can be manually set via a binary command. We can implement the following utility library functions:

- `clearInternalAccount()`
  Removes the internal account from an etcd endpoint
- `verifyInternalAccount(username, password) -> bool`
  Verifies the internal account according to the given credential.
- `resetInternalAccount(username, password)`
  Creates or resets the internal account according to the given username ans password. 

Then, we can:

[...]

Issue #89, opened 2020-02-28
UCP: Add authorization support


breeswish:
## Description

Now we only allow log in use TiDB root user. However we can allow other users to log in, as long as Dashboard functionalities are protected behind authorization.

- For TiDB SQL based features, like statements, we can simply utilize TiDB's authorization itself to ensure access safety.

- For TiDB / PD / TiKV API based features, like KeyVis and Log Exporting, we need to check current user's SQL privilege before allowing user to perform such action. We can limit most of the features to `SUPER` privilege for now. However your PR should be flexible enough so that we can also adjust the privilege requirement of a single feature.

## Score 

* 2530

## Mentor(s)

* @breeswish

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* MySQL
* Go web programming

SSebo:
How about `Role-Based Access Control`?

breeswish:
Thanks for the interest! The table you are talking about actually already exists, that the native SQL user already have a role-based access control. There are different privileges and roles: https://dev.mysql.com/doc/refman/5.7/en/privileges-provided.html

breeswish:
Still relevant.

PR #308, opened 2020-04-01
 UCP88: Support super user login without TiDB

SSebo:
UCP #88 

Signed-off-by: ssebo <ssebo.zzz@gmail.com>

### What problem does this PR solve?

We need to provide alternative login method: supply a super user login username and password. The username and hash of the password is stored in PD etcd so that multiple PD nodes can share the same authenticator.

### What is changed and how it works?

What's Changed:

- [x] add `pkg/util/kv_auth` to support both command line and ui auth by tikv mode login.
- [x] support command line ops to clear, reset tikv mode login account.
- [x] tikv mode ui login, disable tidb related functionality. 
- [x] pd-ctl clear, reset tikv mode login account.

How it Works:

- pass hashed by `cryptopasta.HashPassword` 
- pass store in `/dashboard/kv_mode/auth_key`
- verified by `cryptopasta.CheckPasswordHash`

### Related changes

None

### Check List 

Tests 
- Integration test

###  Refer to a related PR or issue link (optional)

https://github.com/pingcap-incubator/tidb-dashboard/issues/88
https://github.com/pingcap/pd/pull/2360 WIP

### Side effects

- Performance regression

- Breaking backward compatibility

### Release note


SSebo:
@breeswish hi, mentor
- I found etcd path `/dashboard/internal_account/$user`  is sufficient, no need one for username one for pass.
- `tidb-dashboard user clear` seems like sub-sub command, flags seem like only support one level sub command. It is ok to use `tidb-dashboard user -a 'clear' -u 'username'` format command?

breeswish:
> I found etcd path /dashboard/internal_accounts/$user is sufficient, value in etcd is hashed pass, no need one for username one for pass.

This might lead to possibility that there can be multiple log in accounts. I'm not sure whether it's good or bad 🙃

> tidb-dashboard user reset --username X is sub-sub command, flags seems like only support one level sub command. It is ok to use tidb-dashboard user --action 'clear' --username 'username' command?

Yuo can change to use [cobra](https://github.com/spf13/cobra) in order to support sub commands.

> I found hard to get etcdClient instance in cmd , any suggestion?

Indeed. However you can build etcdClient manually considering that current logics inside `cmd` are all for another functionality (`dashboard start`)

SSebo:
@breeswish mentor,
- How about we just use `auth-key` to verify kv mode auth, because only one account no need to set username.
- How about change `internal_account` to `kv_mode_auth`, `internal` is not easy to understand what it really means.

breeswish:
> How about we just use auth-key to verify kv mode auth, because only one account no need to set username.

This is a reserved behaviour for future change. At some time maybe we will let user create multiple accounts, but not now :)

> How about change internal_account to kv_mode_auth, internal is not easy to understand what it really means.

Good idea! Then how about just call it `/dashboard/account`?

pkg/apiserver/user/auth.go
```diff
@@ -55,23 +56,26 @@ type TokenResponse struct {
...
 
 func (f *authenticateForm) Authenticate(tidbForwarder *tidb.Forwarder) (*utils.SessionUser, error) {
-	// TODO: Support non TiDB auth
 	if !f.IsTiDBAuth {
-		return nil, ErrSignInUnsupportedAuthType.New("unsupported auth type, only TiDB auth is supported")
-	}
-	db, err := tidbForwarder.OpenTiDB(f.Username, f.Password)
-	if err != nil {
-		if errorx.Cast(err) == nil {
+		err := globalUtil.VerifyKvModeAuthKey(tidbForwarder.EtcdClient, f.Password)
```
breeswish, line 60: you should not use the EtcdClient inside the tidbForwarder, it's the forwarder's own impl detail and should not be exposed. Better to accept another parameter instead.

SSebo: Got it

SSebo:
@breeswish  mentor, I remove `WIP` prefix,  split this PR into smaller tasks and add check box to indicate the progress.

SSebo:
@breeswish mentor,
- `etcdClient` get from dependency injection
- Add a dropdown in UI to switch login mode, currently are `SQL User` and `TiKV User`, need a better name
- Should I add a etcd path to store username when `TiKV mode` sign in ? Currently username is ignored.
- TiDB related API just return `ErrNotTidbAuth` error. Disabled in UI should be better, I can improve it when pick  [UCP89](https://github.com/pingcap-incubator/tidb-dashboard/issues/89)

SSebo:
@breeswish mentor, add sub command to `pd-ctl` is WIP, https://github.com/pingcap/pd/pull/2360

cmd/tidb-dashboard/cmd/kvmode.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/pingcap-incubator/tidb-dashboard/pkg/pd"
+	globalUtil "github.com/pingcap-incubator/tidb-dashboard/pkg/utils"
+)
+
+//var cfg = &DashboardCLIConfig{}
+
+var kvModeCmd = &cobra.Command{
+	Use:   "kvmode",
+	Short: "tikv mode related ops, including reset, clear tikv mode username, password",
```
breeswish, line 30: Let's call it `kvauth`, since it only includes the auth part and is likely to not include others in future as well.

cmd/tidb-dashboard/cmd/kvmode.go
```diff
@@ -0,0 +1,69 @@
...
+	Long:  ``,
+	Run: func(cmd *cobra.Command, args []string) {
+		if !kvModeDoAuth {
```
breeswish, line 33: I think this command does not need any checks as a parent command. You can move the auth check to the reset command.

cmd/tidb-dashboard/cmd/kvmode_clear.go
```diff
@@ -0,0 +1,46 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
```
breeswish, line 14: maybe you can rearrange this file as `cmd/kvauth/clear.go`

cmd/tidb-dashboard/cmd/kvmode_clear.go
```diff
@@ -0,0 +1,46 @@
...
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/pingcap-incubator/tidb-dashboard/pkg/pd"
+	globalUtil "github.com/pingcap-incubator/tidb-dashboard/pkg/utils"
+)
+
+var kvModeAuthClearCmd = &cobra.Command{
+	Use:   "auth-clear",
```
breeswish, line 27: how about calling it `revoke`, since kvmode log in is forbidden after doing this operation.

cmd/tidb-dashboard/cmd/root.go
```diff
@@ -0,0 +1,231 @@
...
+	keyvisualregion "github.com/pingcap-incubator/tidb-dashboard/pkg/keyvisual/region"
+	"github.com/pingcap-incubator/tidb-dashboard/pkg/swaggerserver"
+	"github.com/pingcap-incubator/tidb-dashboard/pkg/uiserver"
+	"github.com/pingcap-incubator/tidb-dashboard/pkg/utils"
+)
+
+var cfg = &DashboardCLIConfig{}
+var showVersion bool
+
+var rootCmd = &cobra.Command{
```
breeswish, line 49: We can arrange it under a `run` command.

pkg/apiserver/statement/statement.go
```diff
@@ -56,7 +56,11 @@ func Register(r *gin.RouterGroup, auth *user.AuthService, s *Service) {
 // @Security JwtAuth
 // @Failure 401 {object} utils.APIError "Unauthorized failure"
 func (s *Service) schemasHandler(c *gin.Context) {
-	db := utils.GetTiDBConnection(c)
+	db, err := utils.GetTiDBConnection(c)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
```
breeswish, line 63: You don't need to change the signature as long as you introduce some authorization like concept to the `MWAuthRequired`, like:

`auth.MWAuthRequired(AuthOptions{RequireTiDB: true})`

When `RequireTiDB` is set to true, KvModeAuth will fail.

Note that there is another UCP for [authorization](https://github.com/pingcap-incubator/tidb-dashboard/issues/89) so a simple authorization is enough here.

pkg/apiserver/user/auth.go
```diff
@@ -54,34 +56,41 @@ type TokenResponse struct {
...
-	if !f.IsTiDBAuth {
-		return nil, ErrSignInUnsupportedAuthType.New("unsupported auth type, only TiDB auth is supported")
-	}
-	db, err := tidbForwarder.OpenTiDB(f.Username, f.Password)
-	if err != nil {
-		if errorx.Cast(err) == nil {
+func (f *authenticateForm) Authenticate(tidbForwarder *tidb.Forwarder, etcdClient *clientv3.Client) (*utils.SessionUser, error) {
+	if !*f.IsTiDBAuth {
+		//currently only allow root username
+		if f.Username != "root" {
```
breeswish, line 62: why do we only allow root user even in KvAuth mode?

pkg/apiserver/user/auth.go
```diff
@@ -44,7 +46,7 @@ type AuthService struct {
 }
 
 type authenticateForm struct {
-	IsTiDBAuth bool   `json:"is_tidb_auth" binding:"required"`
+	IsTiDBAuth *bool  `json:"is_tidb_auth" binding:"required"`
```
breeswish, line 49: What's the difference of this change?

pkg/apiserver/user/auth.go
```diff
@@ -26,10 +26,12 @@ import (
 	"github.com/gtank/cryptopasta"
 	"github.com/joomcode/errorx"
 	"github.com/pingcap/log"
+	"go.etcd.io/etcd/clientv3"
 	"go.uber.org/zap"
 
 	"github.com/pingcap-incubator/tidb-dashboard/pkg/apiserver/utils"
 	"github.com/pingcap-incubator/tidb-dashboard/pkg/tidb"
+	globalUtil "github.com/pingcap-incubator/tidb-dashboard/pkg/utils"
```
breeswish, line 34: how about calling it `pkg/utils/kvauth`

pkg/pd/etcd.go
```diff
@@ -45,7 +45,7 @@ func init() {
 	_ = zap.RegisterEncoder("etcd-client", newZapEncoder)
 }
 
-func NewEtcdClient(lc fx.Lifecycle, config *config.Config) (*clientv3.Client, error) {
+func NewEtcdClientNoLC(config *config.Config) (*clientv3.Client, error) {
```
breeswish, line 48: @HunDunDM Do you have some suggestions to handle this more elegantly?

pkg/utils/kv_auth.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"context"
+	"errors"
+	"time"
+
+	"github.com/gtank/cryptopasta"
+	"go.etcd.io/etcd/clientv3"
+)
+
+const etcdKvModeAuthKeyPath = "/dashboard/kv_mode/auth_key"
```
breeswish, line 25: Maybe better to call it `kv_auth` directly since there will be no other entities in `kv_mode`.

pkg/utils/kv_auth.go
```diff
@@ -0,0 +1,68 @@
...
+	resp, err := etcdClient.Get(ctx, path)
+	if err != nil {
+		return err
+	}
+	for _, kv := range resp.Kvs {
+		hashedPass = kv.Value
+	}
+
+	if string(hashedPass) == "" {
+		return errors.New("set auth key before verify")
```
breeswish, line 51: Please make explicit errors by using `NewNamespace` and `NewType` since it needs to be translated.

breeswish: Also I believe the error message here should be something like.. `KvAuth account not found`? (since it will happen when the account is cleared or not set)

cmd/kvauth/main.go
```diff
@@ -0,0 +1,132 @@
...
+			os.Exit(1)
+		}
+	},
+}
+
+var kvAuthKey string
+var isHiddenVerify bool
+
+func init() {
+	rootCmd.Flags().StringVar(&coreConfig.DataDir, "data-dir", "/tmp/dashboard-data", "Path to the Dashboard Server data directory")
```
breeswish, line 119: this should not be necessary for kvauth command

cmd/kvauth/main.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+
+		if err := kvauth.VerifyKvAuthKey(client, kvAuthKey); err != nil {
+			fmt.Println("Test auth failed: " + err.Error())
+			os.Exit(1)
+		} else {
+			fmt.Println("Test auth success")
+			os.Exit(0)
+		}
+
```
breeswish, line 70: Looks like these logics are not useful?

cmd/kvauth/main.go
```diff
@@ -0,0 +1,132 @@
...
+var isHiddenVerify bool
+
+func init() {
+	rootCmd.Flags().StringVar(&coreConfig.DataDir, "data-dir", "/tmp/dashboard-data", "Path to the Dashboard Server data directory")
+	rootCmd.PersistentFlags().StringVar(&coreConfig.PDEndPoint, "pd", "http://127.0.0.1:2379", "The PD endpoint that Dashboard Server connects to")
+
+	common.SetClusterTLS(rootCmd, coreConfig)
+	common.SetPDEndPoint(coreConfig)
+	rootCmd.Flags().StringVarP(&kvAuthKey, "key", "k", "", "auth key")
+	rootCmd.Flags().BoolVar(&isHiddenVerify, "hidden-test-auth", false, "do auth")
```
breeswish, line 125: What is hidden-test-auth?

cmd/tidb-dashboard/main.go
```diff
@@ -53,6 +53,30 @@ import (
...
+	execute()
+}
+
+var cfg = &DashboardCLIConfig{}
+var showVersion bool
+
+var rootCmd = &cobra.Command{
+	Use:   "tidb-dashboard",
+	Short: "tidb-dashboard",
+	Long:  `tidb-dashboard which is a tool to monitor tidb cluster system status.`,
```
breeswish, line 66: CLI utility for TiDB Dashboard

cmd/tidb-dashboard/main.go
```diff
@@ -228,3 +184,62 @@ func main() {
 	wg.Wait()
 	log.Info("Stop dashboard server")
 }
+
+func init() {
+	cfg.CoreConfig = &config.Config{}
+	rootCmd.Flags().BoolVarP(&showVersion, "version", "v", false, "Print version information and exit")
```
breeswish, line 190: maybe we can utilize cobra's own version functionality https://github.com/spf13/cobra#version-flag and use `version` command.

pkg/apiserver/user/auth.go
breeswish: we need to allow user to supply a custom super user name as described in the UCP body: "supply a super user login username and password"

pkg/apiserver/user/auth.go
```diff
@@ -54,34 +56,41 @@ type TokenResponse struct {
...
-	if err != nil {
-		if errorx.Cast(err) == nil {
+func (f *authenticateForm) Authenticate(tidbForwarder *tidb.Forwarder, etcdClient *clientv3.Client) (*utils.SessionUser, error) {
+	if !*f.IsTiDBAuth {
+		//currently only allow root username
+		if f.Username != "root" {
+			return nil, ErrSignInOther.NewWithNoMessage()
+		}
+		err := kvauth.VerifyKvAuthKey(etcdClient, f.Password)
+		if err != nil {
```
breeswish, line 66: Please assign a correct error type for it.

pkg/apiserver/utils/tidb_conn.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 PingCAP, Inc.
+// Copyright 2021 PingCAP, Inc.
```
breeswish, line 1: 2020?

pkg/pd/etcd.go
breeswish, line 48: Maybe you can create an fx app and then introduce this as dependency, so that no need to introduce extra function.

pkg/utils/kvauth/kvauth.go
```diff
@@ -0,0 +1,74 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+
+	path := etcdKvAuthKeyPath
+	hashedPass := []byte("")
+	resp, err := etcdClient.Get(ctx, path)
+	if err != nil {
+		return err
+	}
+	for _, kv := range resp.Kvs {
```
breeswish, line 52: maybe better to check kv length to generate a "AccountNotFound"?

cmd/tidb-dashboard/main.go
SSebo: got it

pkg/pd/etcd.go
SSebo: Maybe the  complexity is increased. So far `kvauth` is very simple.

pkg/utils/kvauth/kvauth.go
```diff
@@ -31,44 +31,48 @@ var (
...
 	defer cancel()
 
-	hashedPass, err := cryptopasta.HashPassword([]byte(authKey))
+	hashedPass, err := cryptopasta.HashPassword([]byte(password))
 	if err != nil {
 		return err
 	}
-	_, err = etcdClient.Put(ctx, etcdKvAuthKeyPath, string(hashedPass))
+
+	path := etcdKvAuthKeyPath + "/" + username
```
breeswish, line 75: I prefer not to put username in the key, otherwise you need to additionally work for special symbols and letter cases

SSebo: You are right

cmd/common/common.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/spf13/cobra"
+	"go.etcd.io/etcd/pkg/transport"
+	"go.uber.org/zap"
+
+	"github.com/pingcap-incubator/tidb-dashboard/pkg/config"
+)
+
+// TODO: refactor tidb-dashboard main.go
+
+type DashboardCLIConfig struct {
```
breeswish, line 30: This configuration is only useful when starting the dashboard server instead of running Dashboard CLI tools. It is not very good to put it in `common`.

cmd/kvauth/main.go
```diff
@@ -0,0 +1,113 @@
...
+
+var kvAuthUsername string
+var kvAuthPassword string
+
+func init() {
+	rootCmd.PersistentFlags().StringVar(&coreConfig.PDEndPoint, "pd", "http://127.0.0.1:2379", "The PD endpoint that Dashboard Server connects to")
+	common.SetClusterTLS(rootCmd, coreConfig)
+	common.SetPDEndPoint(coreConfig)
+	kvAuthResetCmd.Flags().StringVarP(&kvAuthUsername, "username", "u", "", "username")
+	kvAuthResetCmd.Flags().StringVarP(&kvAuthPassword, "password", "p", "", "password")
```
breeswish, line 110: It is not a good practice to pass passwords in the paramters (since it will be logged to history). Prefer to prompt for username and password

cmd/common/common.go
```diff
@@ -0,0 +1,74 @@
+// Copyright 2020 PingCAP, Inc.
```
breeswish, line 1: This file may be better to put in `cmd/tidb-dashboard`

cmd/kvauth/main.go
```diff
@@ -0,0 +1,113 @@
+// Copyright 2020 PingCAP, Inc.
```
breeswish, line 1: Maybe better to put in `cmd/tidb-dashbaord`?

cmd/kvauth/main.go
```diff
@@ -0,0 +1,113 @@
...
+	"go.uber.org/fx"
+
+	"github.com/pingcap-incubator/tidb-dashboard/cmd/common"
+	"github.com/pingcap-incubator/tidb-dashboard/pkg/config"
+	"github.com/pingcap-incubator/tidb-dashboard/pkg/pd"
+	"github.com/pingcap-incubator/tidb-dashboard/pkg/utils"
+	"github.com/pingcap-incubator/tidb-dashboard/pkg/utils/kvauth"
+)
+
+func main() {
```
breeswish, line 31: Don't produce multiple binaries. We only need one `tidb-dashboard` binary and have two commands:

```
./bin/tidb-dashboard run ...
./bin/tidb-dashboard kvauth ...
```

pkg/apiserver/user/auth.go
breeswish, line 49: Why does it become a pointer now?

pkg/apiserver/user/auth.go
```diff
@@ -54,34 +56,37 @@ type TokenResponse struct {
...
-		return nil, ErrSignInUnsupportedAuthType.New("unsupported auth type, only TiDB auth is supported")
-	}
-	db, err := tidbForwarder.OpenTiDB(f.Username, f.Password)
-	if err != nil {
-		if errorx.Cast(err) == nil {
+func (f *authenticateForm) Authenticate(tidbForwarder *tidb.Forwarder, etcdClient *clientv3.Client) (*utils.SessionUser, error) {
+	if !*f.IsTiDBAuth {
+		err := kvauth.VerifyKvAuthKey(etcdClient, f.Username, f.Password)
+		if err != nil {
 			return nil, ErrSignInOther.WrapWithNoMessage(err)
```
breeswish, line 63: `ErrSignInOther.WrapWithNoMessage(err)` is not an appropiate error code. If you are already responding an errorx in `VerifyKvAuthKey`, then you can follow L67 ~ L77. Actually they can be extracted to the outside to share the common logic.

pkg/apiserver/utils/error.go
```diff
@@ -27,6 +27,7 @@ var (
 	ErrUnauthorized          = ErrNS.NewType("unauthorized")
 	ErrInsufficientPrivilege = ErrNS.NewType("insufficient_privilege")
 	ErrInvalidRequest        = ErrNS.NewType("invalid_request")
+	ErrNotTidbAuth           = ErrNS.NewType("not_tidb_auth")
```
breeswish, line 30: Seems that this error is not used elsewhere?

pkg/utils/kvauth/kvauth.go
```diff
@@ -0,0 +1,107 @@
...
+	ErrAccountNotFound  = ErrNS.NewType("account_not_found")
+	ErrPasswordNotMatch = ErrNS.NewType("password_not_match")
+)
+
+type Auth struct {
+	Username string `json:"username"`
+	Password string `json:"password"`
+}
+
+// RevokeKvAuthKey delete the etcd path of KV mode user account
```
breeswish, line 40: The function comment should mainly describe its user oriented behaviour, instead of the internal implementation. A better one could be:

```
// RevokeKvAuthKey revokes the KV mode user account stored in etcd.
```

pkg/utils/kvauth/kvauth.go
```diff
@@ -0,0 +1,107 @@
...
+
+	err = cryptopasta.CheckPasswordHash([]byte(auth.Password), []byte(password))
+	if err != nil {
+		return ErrPasswordNotMatch.NewWithNoMessage()
+	}
+
+	return nil
+}
+
+// ResetKvAuthKey set new auth key to etcd
```
breeswish, line 85: ditto

pkg/utils/kvauth/kvauth.go
```diff
@@ -0,0 +1,107 @@
...
+
+// RevokeKvAuthKey delete the etcd path of KV mode user account
+func RevokeKvAuthKey(etcdClient *clientv3.Client) error {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+	_, err := etcdClient.Delete(ctx, etcdKvAuthKeyPath)
+	return err
+}
+
+// VerifyKvAuthKey get hashed pass from etcd and check
```
breeswish, line 48: ditto

ui/dashboardApp/layout/signin/index.js
```diff
@@ -155,6 +172,19 @@ function TiDBSignInForm({ registry }) {
             </Button>
           </Form.Item>
         </AnimationItem>
+        <AnimationItem>
+          <div className={styles.extraLink}>
+            <Dropdown overlay={menu}>
+              <a
+                className="ant-dropdown-link"
```
breeswish, line 179: why not simply `<a>`?

SSebo: user can select `sql mode` or `kvauth mode`.

cmd/tidb-dashboard/main.go
```diff
@@ -63,36 +80,58 @@ type DashboardCLIConfig struct {
...
+}
 
-	clusterCaPath := flag.String("cluster-ca", "", "path of file that contains list of trusted SSL CAs.")
-	clusterCertPath := flag.String("cluster-cert", "", "path of file that contains X509 certificate in PEM format.")
-	clusterKeyPath := flag.String("cluster-key", "", "path of file that contains X509 key in PEM format.")
+func init() {
+	cfg.CoreConfig = &config.Config{}
+	rootCmd.Version = utils.ReleaseVersion
+
+	rootCmd.Flags().StringVar(&cfg.ListenHost, "host", "0.0.0.0", "The listen address of the Dashboard Server")
```
breeswish, line 115: I think you'd better arrange `run` related parameters in `runCmd` and `kvauth` related parameters in `kvauthCmd`, instead of here, since they are not truly "global".

pkg/apiserver/user/auth.go
breeswish: ping

pkg/apiserver/user/auth.go
breeswish: Still returning a SignInOther error instead of a concrete error?

breeswish:
https://github.com/pingcap-incubator/tidb-dashboard/pull/308#discussion_r429747063
https://github.com/pingcap-incubator/tidb-dashboard/pull/308#discussion_r429747223
https://github.com/pingcap-incubator/tidb-dashboard/pull/308#discussion_r433190231

Please also address these comments. Thanks!

pkg/apiserver/user/auth.go
SSebo: I found `bool` will always true, if some param passed,  and found a solution to switch to `*bool`  https://github.com/gin-gonic/gin/issues/814

cmd/tidb-dashboard/main.go
SSebo: got it.

SSebo:
@breeswish mentor，you can continue to review this PR, or I make a new one.

PR #321, opened 2020-04-03
Top5 statement

baurine:
- Show recent 30 mins top 5 statements in the overview page
- Refine some UI: truncate the SQL digest text, show bar for sum latency column

Screenshots:

企业微信20200403033934

image

Some additional notes:

1. At first, I added a new API `/recent_tops` and query method `QueryRecentTopStatements()`, later I realized we can re-use the API `/overviews`, just need to limit the result to 5.

1. For limiting the result to 5, we can add a new query parameter `limit` for `/overviews` API, or truncate the response in client-side, considering we have limited the time range to recent 30 mins, and the max result length is to 200, so I think to do it in client-side is acceptable currently.

1. About how to recognize the recent 30 mins, at first I handled it as the real recent 30 mins, for example, now it is `15:45`, so the recent 30 mins is `15:15  ~ 15:45`, but in fact this condition can't match any records from the database, see the following diagram:

   image

   All records in the statement table, their begin time is 0 or 30 min while end time is 30 or 60 min, there is no record whose begin time >= 15 and end time <= 45 at the same time.

   (Of course, we can just use the begin time condition, for example: `begin_time >= 15`, but it works same as the following way)

   So I changed the recent 30 mins definition. If now is `15:45`, then the recent 30 mins is `15:30 ~ 16:00`, so the real effective time duration is from 0 to 30 mins, for example, an extreme case, now it is `15:31`, the recent 30 mins is recognized `15:30 ~ 16:00`, but the real time duration is 1 min, does it make sense? or should we enlarge its range to `15:00 ~ 16:00` to avoid the above extreme case? how do you think about it?

   Btw, we will meet the same problem when we enable the user to customize the time range.

Hi @breeswish , please help review, thank you! 

Because these commits based on the diagnosis-i18n branch, I will cherry-pick commits to the master branch and change the PR merge to master after the diagnosis-i18n is merged.


ui/src/apps/overview/RootComponent.js
```diff
@@ -6,20 +6,49 @@ import client from '@pingcap-incubator/dashboard_client'
...
+
+const timeFormat = 'YYYY-MM-DD HH:mm:ss'
 
 const App = () => {
   const [cluster, setCluster] = useState(null)
+  const [topStatements, setTopStatements] = useState([])
+  const [loadingStatements, setLoadingStatements] = useState(false)
   const { t } = useTranslation()
+  const [timeRange] = useState(() => {
+    // TODO: unify to use timestamp instead of string
```
breeswish, line 20: can we address this comment?

ui/src/apps/overview/RootComponent.js
```diff
@@ -42,7 +71,16 @@ const App = () => {
                 bordered={false}
                 title={t('overview.top_statements.title')}
               >
-                <Skeleton active />
+                {loadingStatements ? (
```
breeswish, line 74: Prefer to also add a redirection link to see statement details.
We also need to display the cover time of these statements.

ui/src/apps/overview/RootComponent.js
```diff
@@ -6,20 +6,49 @@ import client from '@pingcap-incubator/dashboard_client'
...
+import moment from 'moment'
+
+const timeFormat = 'YYYY-MM-DD HH:mm:ss'
 
 const App = () => {
   const [cluster, setCluster] = useState(null)
+  const [topStatements, setTopStatements] = useState([])
+  const [loadingStatements, setLoadingStatements] = useState(false)
   const { t } = useTranslation()
+  const [timeRange] = useState(() => {
```
breeswish, line 19: how about just use the latest available period?

baurine: Good idea!

ui/src/apps/overview/RootComponent.js
baurine: Clicking the SQL digest will jump to its detail page, here you mean to add a link to statement overview page?

ui/src/apps/overview/RootComponent.js
baurine: How about creating an issue #334 to handle it individually?

baurine:
Hi @breeswish , updated according to your review comments:

- use the latest available time range
- add a link to the statements overview page

The screenshot:

企业微信20200409022050

Tried to put the link in the right of the header, but it seems it is too far away with the title, like this:

企业微信20200409022237

If you have better design, please let me know, I am not good at design.

ui/src/apps/overview/RootComponent.js
```diff
@@ -40,9 +64,28 @@ const App = () => {
               <Card
                 size="small"
                 bordered={false}
-                title={t('overview.top_statements.title')}
+                title={
+                  <div style={{ display: 'flex', alignItems: 'baseline' }}>
```
breeswish, line 68: is it better to use `center`?

ui/src/apps/overview/RootComponent.js
```diff
@@ -40,9 +64,28 @@ const App = () => {
...
+                    <h3 style={{ marginRight: 8 }}>
+                      {t('overview.top_statements.title')}
```
breeswish, line 70: We can also display the time range here.

ui/src/apps/overview/RootComponent.js
baurine: Tried, but it looks like this:

image

because the h3 tag has some additional padding-bottom. (but of course, we can use div replace the h3 and increase its font-size).

ui/src/apps/overview/RootComponent.js
baurine: Good point, will add it soon.

baurine:
After adding the time range to the title, it seems it is suitable to put the "More" link to the header right.

企业微信20200409032021

Issue #228, opened 2020-03-17
UCP: Refine Log Searching Models


breeswish:
## Description

The models in log searching module can be improved:

1. Currently it embed diagnosepb structure directly, which is a potential risk: in protobuffer field rename is a compatible change but it will break Dashboard models.
2. Can be changed to use the new `RequestTargetNode` and `RequestTargetStatistics`. See https://github.com/pingcap-incubator/tidb-dashboard/pull/214 for examples.

## Score

* 900

## Mentor(s)

* @Deardrops 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Golang
* TypeScript

PR #322, opened 2020-04-04
refine log searching models

STRRL:
UCP #228

Roadmap:
- [x] replace struct fields with `RequestTargetNode`
- [x] related frontend things
- [x] replace struct fields with `RequestTargetStatistics `
- [x] related frontend things

# What have you changed?
Refactor `SearchTarget` and related frontend codes:

Before:
```go
type SearchTarget struct {
	Kind       ServerType `json:"kind" example:"1"`
	IP         string     `json:"ip" gorm:"size:32" example:"127.0.0.1"`
	Port       int        `json:"port" example:"4000"`
	StatusPort int        `json:"status_port" example:"10080"`
}
```

After:
```go
type SearchTarget struct {
	Target     utils.RequestTargetNode `json:"target" gorm:"embedded;embedded_prefix:target_"`
	StatusPort int                     `json:"status_port" example:"10080"`
}
```

# What are the type of the changes?
Improvement.

# How has this PR been tested?
Manually tested.

I have no idea how to write unit test for it. 😅

# Does this PR affect documentation (docs/docs-cn) update?
No.

# Refer to a related PR or issue link
#228


STRRL:
Hi @breeswish,  I found no where to use `RequestTargetStatistics`, so I think log-searching is no need to use this struct, is it right?

breeswish:
@STRRL Currently `profiling` is using `RequestTargetStatistics`, you can take a look at `profiling/model.go`.

STRRL:
@breeswish I already see these usage, but I could NOT find out the way suing `RequestTargetStatistics ` in log-seraching.

It seems that nowhere need "count the number of components(tidb/tidkv/pd)" in log-searching page, or I miss anything?

breeswish:
@STRRL It is only to be displayed in front-end, where you can find its usage by grep `target_stats`. In log searching, it is supposed to be only used for display purpose as well.

breeswish:
@STRRL It is only to be displayed in front-end, where you can find its usage by grep `target_stats`. In log searching, it is supposed to be only used for display purpose as well.

STRRL:
@breeswish Oh, I known when to use `RequestTargetStatistics`.  

After that I still make sure that log-searching do NOT need using this struct.🤔

Could you take a look on log-searching web page and confirm that please?

breeswish:
@STRRL In log search backend you only need to update it, and no need to read it. In log search front-end, you need to read it and display. Does this make sense to you?

STRRL:
Here is not a simple demand like “show task statistics” in profiling

profiling

log-searching

With log-searching, it use `TaskModel` to rendering this tree.

This tree need more detail such like display "${IP}:${port}", or aggregating number of tasks state(Running/Finished/Error). 

I think it is NOT necessary to use `RequestTargetStatistics` with this situation. In other world, `RequestTargetStatistics` is too simple to work with this complex demand. 

Please tell me what should I do for applying `RequestTargetStatistics` if you think using `RequestTargetStatistics` is necessary. 🙈

breeswish:
@STRRL `RequestTargetStatistics` is only supposed to be used in log search history page. Currently in log search history page, not only `TaskGroup` is loaded, but also all `Tasks`, which is not efficient. By using `RequestTargetStatistics`, we only need to load `TaskGroup`.

STRRL:
@breeswish Thanks! I did missing "log search history" page.

STRRL:
Hi @breeswish, PTAL

pkg/apiserver/logsearch/service.go
```diff
@@ -84,6 +84,11 @@ type TaskGroupResponse struct {
 	Tasks     []*TaskModel   `json:"tasks"`
 }
 
+type TaskGroupStats struct {
+	TaskGroup   TaskGroupModel                `json:"task_group"`
```
breeswish, line 88: You need to change the `TaskGroup` model to include the number, otherwise all `Task`s are still being queried and it is very inefficient: `err = s.db.Where("task_group_id = ?", taskGroup.ID).Find(&tasks).Error`.

STRRL: Oh I know what your mean. `RequestTargetStatistics` should be persisted into DB, ike `TaskGroupModel` in profiling module.

I will do this very soon.

STRRL:
@breeswish  PTAL : )

pkg/apiserver/logsearch/models.go
```diff
@@ -106,33 +94,31 @@ func (r *SearchLogRequest) Value() (driver.Value, error) {
 }
 
 type SearchTarget struct {
```
breeswish, line 96: There is no need to leave this structure :) `RequestTargetNode` already includes everything it needs. The **actual port (10080 / 2379 / ...)** shall be stored in `port`, while the **port (4000 / 2379 / ...)** shall be stored in `DisplayName` (to be used for front-end display only). You can take a look at Profiling to see how it works.

pkg/apiserver/logsearch/service.go
```diff
@@ -154,15 +159,8 @@ func (s *Service) GetAllTaskGroups(c *gin.Context) {
 	}
 	var resp = make([]TaskGroupResponse, 0, len(taskGroups))
```
breeswish, line 160: This function can return `// @Success 200 {array} TaskGroupModel` directly and no need to loop again building another structure.

pkg/apiserver/logsearch/service.go
```diff
@@ -106,10 +106,15 @@ func (s *Service) CreateTaskGroup(c *gin.Context) {
 		_ = c.Error(utils.ErrInvalidRequest.NewWithNoMessage())
 		return
 	}
-
+	nodes := make([]utils.RequestTargetNode, 0)
```
breeswish, line 109: Your life can be easy if `CreateTaskGroupRequest` is simply:

```go
type CreateTaskGroupRequest struct {
    Request       SearchLogRequest `json:"request" binding:"required"`
    Targets      []utils.RequestTargetNode `json:"targets"`
}
```

STRRL:
@breeswish  PTAL : )

ui/src/apps/searchLogs/components/SearchHistory.tsx
```diff
@@ -134,20 +127,17 @@ export default function SearchHistory() {
...
 
   const historyList: History[] = taskGroups.map((taskGroup) => {
-    const {
-      timeRange,
-      logLevel,
-      components,
-      searchValue,
-    } = parseSearchingParams(taskGroup)
-    const taskGroupID = taskGroup.task_group?.id || 0
+    const { timeRange, logLevel, stats, searchValue } = parseHistoryStatsParams(
```
breeswish, line 130: I think `parseHistoryStatsParams` can be simply removed. It doesn't do anything useful now.

ui/src/apps/searchLogs/components/SearchHistory.tsx
```diff
@@ -134,20 +127,17 @@ export default function SearchHistory() {
   }
 
   const historyList: History[] = taskGroups.map((taskGroup) => {
```
breeswish, line 129: it is totally fine not to map the taskgroup now, since there is no need to compute something complex.

ui/src/apps/searchLogs/components/utils.ts
```diff
@@ -109,33 +118,46 @@ export function parseSearchingParams(resp: LogsearchTaskGroupResponse): Params {
...
+    if (task.target === undefined) {
       return
     }
-    targets.push(task.search_target)
+    targets.push(task.target)
   })
   return targets
 }
 
 export function getGRPCAddress(
```
breeswish, line 147: This function is no longer useful I guess.

ui/src/apps/searchLogs/components/SearchHeader.tsx
```diff
@@ -15,36 +15,36 @@ import styles from './Styles.module.css'
...
 
-  return ServerTypeList.filter((kind) => servers[kind].length > 0).map(
+  return NodeKindList.filter((kind) => servers[kind].length > 0).map(
     (kind) => ({
       title: namingMap[kind],
       value: kind,
       key: kind,
-      children: servers[kind].map((item: LogsearchSearchTarget) => {
+      children: servers[kind].map((item: UtilsRequestTargetNode) => {
         const addr = getAddress(item)
```
Deardrops, line 48: Simply use:
```
const addr = item.display_name ?? ''
```
And remove function `getAddress`

ui/src/apps/searchLogs/components/utils.ts
Deardrops: Yes, just remove this funtion and use `target.display_name`

ui/src/apps/searchLogs/components/SearchHistory.tsx
STRRL: I have a little confusion about this.

Here is a TypeScript type `History` as model for rendering `CardTable`. 

Line 129 to 143 will mapping the data from `LogserachTaskGroupModel` to `History`, need using `parseHistoryStatsParams()` to transfer some fields and resolve default values.

You means  using `LogserachTaskGroupModel` directly instead of `History`?

breeswish: Yes, I think it is totally fine to remove `History` and just use `LogserachTaskGroupModel`.

STRRL:
Hi @breeswish, could you take a look on CI e2e test failed?

I could reproduce this issue on my host at first. But after I deleted `tests/node_modules` and rerun `yarn install`, it changes to another different error.

STRRL:
Hi, @breeswish PTAL.

I write some `xxxRenders`(), it accepts `LogsearchSearchLogRequest` as parameter which might be bad smell. Please tell me if there is another elegant way.

ui/lib/apps/SearchLogs/components/SearchHeader.tsx
```diff
@@ -13,40 +13,39 @@ import { useNavigate } from 'react-router-dom'
...
-  return ServerTypeList.filter((kind) => servers[kind].length > 0).map(
+  return NodeKindList.filter((kind) => servers[kind].length > 0).map(
     (kind) => ({
       title: namingMap[kind],
       value: kind,
       key: kind,
-      children: servers[kind].map((item: LogsearchSearchTarget) => {
-        const addr = getAddress(item)
+      children: servers[kind].map((item: UtilsRequestTargetNode) => {
+        const addr = item.display_name ?? ''
```
breeswish, line 48: can we use `!` instead? since `display_name` shall never be null

breeswish: what do you think @Deardrops

ui/lib/apps/SearchLogs/components/SearchHistory.tsx
```diff
@@ -70,9 +68,20 @@ export default function SearchHistory() {
       const res = await client.getInstance().logsTaskgroupsGet()
       setTaskGroups(res.data)
     }
+
     getData()
   }, [])
 
+  function levelRender(request: LogsearchSearchLogRequest) {
+    return LogLevelMap[request?.min_level ?? 0]
```
breeswish, line 76: maybe can use `!`

ui/lib/apps/SearchLogs/components/SearchHistory.tsx
```diff
@@ -176,34 +166,36 @@ export default function SearchHistory() {
...
+            dataIndex="target_stats"
+            key="target_stats"
             render={componentRender}
           />
           <Column
             title={t('search_logs.common.keywords')}
-            dataIndex="keywords"
-            key="keywords"
+            dataIndex="search_request"
+            key="searchValue"
```
breeswish, line 197: `key="keywords"`

ui/lib/apps/SearchLogs/components/SearchHistory.tsx
```diff
@@ -213,8 +205,8 @@ export default function SearchHistory() {
           />
           <Column
             title={t('search_logs.history.action')}
-            dataIndex="action"
-            key="action"
+            dataIndex="id"
```
breeswish, line 208: maybe actually no need to pass `dataIndex`?

ui/lib/apps/SearchLogs/components/SearchHistory.tsx
```diff
@@ -213,8 +205,8 @@ export default function SearchHistory() {
...
+            key="id"
```
breeswish, line 209: key="action"

ui/lib/apps/SearchLogs/components/SearchHeader.tsx
Deardrops: OK

STRRL:
Hi @breeswish, I just make a new commit for cleaning up more code. PTAL.

PR #325, opened 2020-04-06
Improve KeyViz auto refresh

breeswish:
image

image

image

image

image

### DIFF
--- a/ui/packages/keyvis/src/components/ToolBar.tsx
+++ b/ui/packages/keyvis/src/components/ToolBar.tsx
@@ -1,5 +1,5 @@
-import React, { Component } from 'react'
-import {
+import React, { Component, useEffect } from 'react'
+import Icon, {
   AreaChartOutlined,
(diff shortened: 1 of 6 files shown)


ui/packages/keyvis/src/components/ToolBar.tsx
```diff
@@ -93,6 +140,28 @@ class KeyVisToolBar extends Component<IKeyVisToolBarProps & WithTranslation> {
       { text: t('keyvis.toolbar.view_type.all'), value: 'integration' },
     ]
 
+    // in seconds
+    const autoRefreshOptions = [15, 30, 1 * 60, 2 * 60, 5 * 60, 10 * 60]
```
HunDunDM, line 144: I think 5s/15s/30s/1m/5m is enough?

breeswish: actually I have no idea 🙃

PR #328, opened 2020-04-07
profiling: remove some dependencies

rleungx:


pkg/apiserver/profiling/profile.go
```diff
@@ -15,64 +15,169 @@ package profiling
...
+
+type fetcher struct {
+	ctx        context.Context
+	httpClient *http.Client
+	target     *utils.RequestTargetNode
+	output     string
+	tls        bool
+}
+
+func (f *fetcher) Fetch(src string, duration, timeout time.Duration) (*profile.Profile, string, error) {
```
breeswish, line 104: do we still need this function?

pkg/apiserver/profiling/profile.go
```diff
@@ -15,64 +15,169 @@ package profiling
...
+	case utils.NodeKindTiKV:
+		return fetchFlameGraphSVG(ctx, httpClient, target, fileNameWithoutExt, profileDurationSecs, tls)
+	case utils.NodeKindPD, utils.NodeKindTiDB:
+		return fetchPprofSVG(ctx, httpClient, target, fileNameWithoutExt, "dot", profileDurationSecs, tls)
 	default:
 		return "", fmt.Errorf("unsupported target %s", target)
 	}
+}
+
+func fetchFlameGraphSVG(ctx context.Context, httpClient *http.Client, target *utils.RequestTargetNode, fileNameWithoutExt string, profileDurationSecs uint, tls bool) (string, error) {
```
breeswish, line 157: maybe we should call it something like fetchTiKVProfiling since this function will not work for all components

pkg/apiserver/profiling/profile.go
```diff
@@ -15,64 +15,169 @@ package profiling
...
-	// TiKV dose not support TLS for the status server currently
-	if target.Kind != utils.NodeKindTiKV && tls {
+	if tls {
 		schema = "https"
 	}
-	return fmt.Sprintf("%s://%s:%d%s", schema, target.IP, target.Port, url), nil
+	url := fmt.Sprintf("%s://%s:%d/debug/pprof/profile?seconds=%d", schema, target.IP, target.Port, profileDurationSecs)
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return "", err
```
breeswish, line 165: please make some more readable errors

pkg/apiserver/profiling/profile.go
```diff
@@ -15,64 +15,169 @@ package profiling
...
-	var url string
-	secs := strconv.Itoa(int(profileDurationSecs))
+func profileAndWriteSVG(ctx context.Context, target *utils.RequestTargetNode, fileNameWithoutExt string, profileDurationSecs uint, httpClient *http.Client, tls bool) (string, error) {
 	switch target.Kind {
-	case utils.NodeKindPD:
-		url = "/pd/api/v1/debug/pprof/profile?seconds=" + secs
-	case utils.NodeKindTiKV, utils.NodeKindTiDB:
-		url = "/debug/pprof/profile?seconds=" + secs
+	case utils.NodeKindTiKV:
+		return fetchFlameGraphSVG(ctx, httpClient, target, fileNameWithoutExt, profileDurationSecs, tls)
```
breeswish, line 149: PD-Allow-follower-handle header will not be set for PD, is it ok?

pkg/apiserver/profiling/profile.go
rleungx: Yes, we need to use it to get DOT output

pkg/apiserver/profiling/profile.go
rleungx: `PD-Allow-follower-handle` will be set for PD in `getProfile` function

breeswish:
Manually confirmed that it works 👏

PR #333, opened 2020-04-08
Use /statement instead of /statement/overview

Fullstop000:
After https://github.com/pingcap-incubator/tidb-dashboard/pull/301, entering `/statement` will be redirected to `/statement/overview` after mounted but this does not affect the later routing to `/statement`. This PR fixes this behavior.

### DIFF
--- a/ui/packages/statement/src/RootComponent.tsx
+++ b/ui/packages/statement/src/RootComponent.tsx
@@ -23,7 +23,6 @@ const App = () => {
     curTimeRange: undefined,
   } as SearchOptions)
   const searchContext = { searchOptions, setSearchOptions }
-
   return (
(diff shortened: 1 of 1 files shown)


Fullstop000:
@baurine PTAL

breeswish:
How about changing `/statement/overview` to `/statement`? Then we don't need to redirect any more.

baurine:
> How about changing `/statement/overview` to `/statement`? Then we don't need to redirect any more.

Agree, I thought it too, to keep the same as `/diagnose` and `/diagnose/:id`.

Fullstop000:
@baurine @breeswish done

Fullstop000:
Seems the CI is broken unexpectly :(

PR #342, opened 2020-04-10
cluster_info: display host status in cluster info page

Deardrops:
Preview:
Screenshot_2020-04-14 TiDB Dashboard

image

image


pkg/apiserver/clusterinfo/host.go
```diff
@@ -0,0 +1,384 @@
...
+}
+
+type PartitionMap map[string]Partition
+
+func queryPartition(db *gorm.DB, instance Instance) (PartitionMap, error) {
+	sql := fmt.Sprintf("select DEVICE_NAME, NAME, VALUE from INFORMATION_SCHEMA.CLUSTER_HARDWARE where type = '%s' and instance = '%s' and device_type = 'disk';",
+		instance.ServerType,
+		instance.Address,
+	)
+	rows, err := querySQL(db, sql)
```
breeswish, line 276: Your life will be easier if you use gorm as a ORM: https://gorm.io/docs/sql_builder.html#Run-Raw-SQL

ui/src/apps/clusterInfo/components/HostTable.js
```diff
@@ -0,0 +1,218 @@
+import client from '@pingcap-incubator/dashboard_client'
+import { CardTable } from '@pingcap-incubator/dashboard_components'
+import { Progress, Tooltip } from 'antd'
+import byteSize from 'byte-size'
+import React, { useEffect, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+
+function useDataSource() {
```
breeswish, line 8: you can use `useClientRequest`

ui/src/apps/clusterInfo/components/HostTable.js
```diff
@@ -0,0 +1,222 @@
...
+          const used = total - free
+          const usedPercent = (used / total).toFixed(3)
+          const title = (
+            <div>
+              Used: {byteSizeToStr(used, 1)} ({toPercentStr(usedPercent)})
+            </div>
+          )
+          return (
+            <Tooltip title={title} key={i}>
+              <Progress
```
breeswish, line 203: You can use the new `Bar` component.

ui/src/apps/clusterInfo/components/InstanceTable.js
```diff
@@ -0,0 +1,227 @@
...
+  return async (node) => {
+    await client
+      .getInstance()
+      .topologyTidbAddressDelete(`${node.ip}:${node.port}`)
+    updateData()
+  }
+}
+
+function useClusterNodeDataSource() {
+  const [isLoading, setIsLoading] = useState(true)
```
breeswish, line 104: do you mind changing this to use `useClientRequest` by hand?

pkg/apiserver/clusterinfo/host.go
```diff
@@ -0,0 +1,354 @@
...
+
+	return m, nil
+}
+
+type PartitionMap map[string]Partition
+
+func queryPartition(db *gorm.DB, instance Instance) (PartitionMap, error) {
+	var m = make(PartitionMap)
+
+	sql := fmt.Sprintf("select DEVICE_NAME, NAME, VALUE from INFORMATION_SCHEMA.CLUSTER_HARDWARE where type = '%s' and instance = '%s' and device_type = 'disk';",
```
breeswish, line 283: this should use placeholders (`?`) to avoid SQL injection

pkg/apiserver/clusterinfo/host.go
```diff
@@ -0,0 +1,354 @@
...
+			targetDisk = disk
+		}
+	}
+
+	return targetDisk
+}
+
+type HostMap map[string][]Instance
+
+func loadHosts(db *gorm.DB) (HostMap, error) {
```
breeswish, line 138: There may be problems if you load instances from TiDB instead of using our own cluster info interface, that closed TiDB is not in the list.

pkg/apiserver/clusterinfo/host.go
```diff
@@ -0,0 +1,354 @@
...
+		if instances, ok := hostMap[ip]; ok {
+			list = instances
+		}
+
+		list = append(list, Instance{
+			Address:    instance,
+			ServerType: serverType,
+		})
+		hostMap[ip] = list
+	}
```
breeswish, line 164: Maybe we should load this table once, then parse it? Otherwise I guess full hardware info is loaded multiple times.

```go
type clusterInfoTableModel struct {
    Type ...
    Instance ...
    DeviceType ...
    DeviceName ...
    Name ...
    Value ...
}

var records []clusterInfoTableModel

db.Table('INFORMATION_SCHEMA.CLUSTER_INFO').Find(&records)

for ... {
    switch rec.DeviceType {
        ....
    }
}
```

pkg/apiserver/clusterinfo/host.go
```diff
@@ -0,0 +1,381 @@
...
+	return memoryMap, cpuMap, nil
+}
+
+func saveMemory(row ClusterTableModel, m *MemoryMap) {
+	ip := parseIP(row.Instance)
+
+	memory, ok := (*m)[ip]
+	if !ok {
+		memory = &Memory{}
```
breeswish, line 233: A small optimization:

```golang
memory, ok := (*m)[ip]
if !ok {
    memory = &Memory{}
    (*m)[ip] = memory
}   
```

No need to re-assign `(*m)[ip] = memory` when entry exists.

pkg/apiserver/clusterinfo/service.go
```diff
@@ -140,3 +153,85 @@ func (s *Service) topologyGetAlertCount(c *gin.Context) {
...
+// @Summary Get all host information in the cluster
+// @Description Get information about host in the cluster
+// @Produce json
+// @Success 200 {array} HostInfo
+// @Router /host/all [get]
+// @Security JwtAuth
+// @Failure 401 {object} utils.APIError "Unauthorized failure"
+func (s *Service) hostHandler(c *gin.Context) {
+	db := utils.GetTiDBConnection(c)
+	db.LogMode(true)
```
breeswish, line 166: Maybe need to remove

pkg/apiserver/clusterinfo/service.go
```diff
@@ -140,3 +153,85 @@ func (s *Service) topologyGetAlertCount(c *gin.Context) {
...
+		infos, err = GetAllHostInfo(db)
+	}()
+	wg.Wait()
+
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	AllHosts := loadHostsFromClusterInfo(clusterInfo)
```
breeswish, line 201: allHosts

pkg/apiserver/clusterinfo/service.go
```diff
@@ -140,3 +153,85 @@ func (s *Service) topologyGetAlertCount(c *gin.Context) {
...
+		infos = append(infos, HostInfo{
+			IP:          host,
+			Unavailable: true,
+		})
+	}
+
+	c.JSON(http.StatusOK, infos)
+}
+
+func loadHostsFromClusterInfo(info ClusterInfo) []string {
```
breeswish, line 219: this is actually loadUnavailableHosts

pkg/apiserver/clusterinfo/service.go
```diff
@@ -140,3 +153,85 @@ func (s *Service) topologyGetAlertCount(c *gin.Context) {
...
+	var allNodes []string
+	for _, node := range info.TiDB.Nodes {
+		if node.Status != clusterinfo.ComponentStatusUp {
+			allNodes = append(allNodes, node.IP)
+		}
+	}
+	for _, node := range info.TiKV.Nodes {
+		if node.Status != clusterinfo.ComponentStatusUp {
```
breeswish, line 227: for TiKV, offline and tomestone states are also accessible

pkg/apiserver/clusterinfo/host.go
```diff
@@ -0,0 +1,381 @@
...
+}
+
+func splitPath(path string) []string {
+	if path == "" {
+		return []string{}
+	}
+	return strings.Split(path, string(filepath.Separator))
+}
+
+func inferPartition(dataDir string, diskMap PartitionMap) Partition {
```
breeswish, line 123: can we simply use prefix check? The longest prefix should be the match one

ui/lib/apps/ClusterInfo/translations/en.yaml
```diff
@@ -20,5 +20,17 @@ cluster_info:
...
+      columns:
+        ip: Address
+        cpu: CPU
+        cpu_usage: CPU Usage
+        memory: Memory
+        memory_usage: Memory Usage
+        deploy: Disk
+        disk_size: Disk Size
+        disk_usage: Disk Usage
+      instanceUnavailable: Instance on the host is unavailable
```
breeswish, line 34: Host information is unknown due to instance unreachable

ui/lib/apps/ClusterInfo/translations/zh-CN.yaml
```diff
@@ -20,5 +20,17 @@ cluster_info:
...
+      columns:
+        ip: 机器地址
+        cpu: CPU
+        cpu_usage: CPU 使用率
+        memory: 物理内存
+        memory_usage: 内存使用率
+        deploy: 部署磁盘
+        disk_size: 磁盘容量
+        disk_usage: 磁盘使用率
+      instanceUnavailable: 这个节点上的实例不可用
```
breeswish, line 34: 获取该主机信息失败：无法访问实例

ui/lib/apps/ClusterInfo/translations/zh-CN.yaml
```diff
@@ -20,5 +20,17 @@ cluster_info:
         offline: 下线中
         unknown: 未知
         unreachable: 无法访问
+    host_table:
+      title: 节点
```
breeswish, line 24: 主机

ui/lib/apps/ClusterInfo/translations/zh-CN.yaml
```diff
@@ -20,5 +20,17 @@ cluster_info:
...
+      columns:
+        ip: 机器地址
```
breeswish, line 26: 主机地址

ui/lib/apps/ClusterInfo/components/HostTable.js
```diff
@@ -0,0 +1,219 @@
...
+      maxWidth: 150,
+      isResizable: true,
+      isCollapsible: true,
+      onRender: ({ ip, unavailable }) => {
+        if (unavailable) {
+          return (
+            <Tooltip
+              title={t('cluster_info.list.host_table.instanceUnavailable')}
+            >
+              <Text type="warning">{ip}</Text>
```
breeswish, line 38: `<Text type="warning"><WarningOutlined /> {ip}</Text>`

ui/lib/apps/ClusterInfo/components/HostTable.js
```diff
@@ -0,0 +1,219 @@
...
+          <>
+            <div>User: {getValueFormat('percentunit')(user)}</div>
+            <div>System: {getValueFormat('percentunit')(system)}</div>
+          </>
+        )
+        return (
+          <Tooltip title={title}>
+            <Bar
+              value={[user, system]}
+              colors={['#667cff', '#ff4d4f']}
```
breeswish, line 78: Avoid hard code colors. https://ant.design/docs/spec/colors#header
Also you can write `[null, '...']` so that the first segment uses default color

ui/lib/components/Bar/Bar.module.less
```diff
@@ -12,6 +12,7 @@
   flex-direction: row;
   align-items: center;
   min-height: 1em;
+  height: 100%;
```
breeswish, line 15: This might be a breaking change, since `height: 100%` varies depends on its parent container. Do we really need this style rule?

pkg/apiserver/clusterinfo/host.go
Deardrops: These two methods are same. I'd prefer this method.

breeswish: I'm worried about the correctness. For example, what's the purpose of `filepath.Rel(disk.Path, dataDir)` considering that `dataDir` may not be a directory inside `disk.Path`?

Issue #334, opened 2020-04-09
UCP: Use timestamp replace string as time parameters for statements APIs


baurine:
## Description

This task  is to use timestamp instead of string as time parameters when request statements APIs, for example, the original API is: `/dashboard/api/statements/overviews?schemas=&begin_time=2020-04-08%2021%3A30%3A00&end_time=2020-04-08%2022%3A00%3A00`, change it to `/dashboard/api/statements/overviews?schemas=&begin_time=1586352600&end_time=1586354400`.

## Score

* 100

## Mentor(s)

* @baurine 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Go
* JavaScript

PR #345, opened 2020-04-11
Use timestamp instead of string

Fullstop000:
UCP https://github.com/pingcap-incubator/tidb-dashboard/issues/334

### DIFF
--- a/pkg/apiserver/statement/models.go
+++ b/pkg/apiserver/statement/models.go
@@ -17,12 +17,6 @@ import (
 	"strings"
 )
 
-// TimeRange represents a range of time
-type TimeRangeModel struct {
-	BeginTime string `json:"begin_time"`
-	EndTime   string `json:"end_time"`
-}
-
 // TimeRangeUnix represents a range of time in Unix seconds
 type TimeRange struct {
 	BeginTime int64 `json:"begin_time"`
--- a/pkg/apiserver/statement/models.go
+++ b/pkg/apiserver/statement/models.go
@@ -13,12 +13,20 @@
 
 package statement
 
-import "strings"
+import (
+	"strings"
+)
 
 // TimeRange represents a range of time
-type TimeRange struct {
+type TimeRangeModel struct {
 	BeginTime string `json:"begin_time"`
 	EndTime   string `json:"end_time"`
+}
+
+// TimeRangeUnix represents a range of time in Unix seconds
+type TimeRange struct {
+	BeginTime int64 `json:"begin_time"`
+	EndTime   int64 `json:"end_time"`
 }
 
 // Overview represents the overview of a statement
--- a/pkg/apiserver/statement/models.go
+++ b/pkg/apiserver/statement/models.go
@@ -17,7 +17,7 @@ import (
 	"strings"
 )
 
-// TimeRangeUnix represents a range of time in Unix seconds
+// TimeRange represents a range of time
 type TimeRange struct {
 	BeginTime int64 `json:"begin_time"`
 	EndTime   int64 `json:"end_time"`
--- a/pkg/apiserver/statement/queries.go
+++ b/pkg/apiserver/statement/queries.go
@@ -38,7 +38,7 @@ func QuerySchemas(db *gorm.DB) ([]string, error) {
 	return schemas, nil
 }
 
-func QueryTimeRanges(db *gorm.DB) (result []*TimeRange, err error) {
+func QueryTimeRanges(db *gorm.DB) (result []*TimeRangeModel, err error) {
 	err = db.
 		Select(`
 			DISTINCT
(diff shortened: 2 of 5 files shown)


Fullstop000:
@baurine PTAL

pkg/apiserver/statement/statement.go
```diff
@@ -73,11 +77,29 @@ func (s *Service) schemasHandler(c *gin.Context) {
 func (s *Service) timeRangesHandler(c *gin.Context) {
 	db := utils.GetTiDBConnection(c)
 	timeRanges, err := QueryTimeRanges(db)
+	timeRangeResponse := []*TimeRange{}
+	for _, timeRange := range timeRanges {
+		beginTime, err := time.Parse(layout, timeRange.BeginTime)
```
breeswish, line 82: I think we should store timestamp as well. The human readable time will have problems with time zone.

pkg/apiserver/statement/models.go
```diff
@@ -13,14 +13,22 @@
...
+)
 
 // TimeRange represents a range of time
-type TimeRange struct {
+type TimeRangeModel struct {
 	BeginTime string `json:"begin_time"`
 	EndTime   string `json:"end_time"`
 }
 
+// TimeRangeUnix represents a range of time in Unix seconds
```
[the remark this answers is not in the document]
Fullstop000: Working on this. Can't change the inner column type because it's a system table (in `PERFORMANCE_SCHEMA`) in TiDB.

ui/packages/statement/src/components/StatementsOverview.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { useState, useReducer, useEffect, useContext } from 'react'
 import { Select, Button, Modal } from 'antd'
+import dayjs from 'dayjs'
```
baurine, line 3: If using dayjs, we need to declare it as a dependency in the statement/package.json else it increases the bundle size, considering statement has already depended on the moment.js, I prefer to use the moment.js currently (we will replace the moment.js by day.js later)

breeswish: actually day.js is already used @baurine

baurine: ok, got! so let's just add dayjs to the package.json dependencies field, @Fullstop000

pkg/apiserver/statement/queries.go
```diff
@@ -41,9 +41,8 @@ func QuerySchemas(db *gorm.DB) ([]string, error) {
 func QueryTimeRanges(db *gorm.DB) (result []*TimeRange, err error) {
 	err = db.
 		Select(`
-			DISTINCT
-			summary_begin_time AS begin_time,
-			summary_end_time AS end_time
+			DISTINCT CONVERT(UNIX_TIMESTAMP(summary_begin_time), UNSIGNED INTEGER) AS begin_time,
```
breeswish, line 44: I think use `FLOOR()` is better.

pkg/apiserver/statement/statement.go
```diff
@@ -125,10 +134,19 @@ func (s *Service) overviewsHandler(c *gin.Context) {
...
+		_ = c.Error(fmt.Errorf("invalid begin_time: %s", err))
+		return
+	}
+	endTime, err := strconv.Atoi(c.Query("end_time"))
+	if err != nil {
+		_ = c.Error(fmt.Errorf("invalid end_time: %s", err))
+		return
+	}
+
+	detail, err := QueryStatementDetail(db, schema, time.Unix(int64(beginTime), 0).Format(layout), time.Unix(int64(endTime), 0).Format(layout), digest)
```
breeswish, line 149: no need to format back again,  timestamp can be used all time

pkg/apiserver/statement/statement.go
```diff
@@ -150,10 +168,18 @@ func (s *Service) detailHandler(c *gin.Context) {
...
+	if err != nil {
+		_ = c.Error(fmt.Errorf("invalid begin_time: %s", err))
+		return
+	}
+	endTime, err := strconv.Atoi(c.Query("end_time"))
+	if err != nil {
+		_ = c.Error(fmt.Errorf("invalid end_time: %s", err))
+		return
+	}
+	nodes, err := QueryStatementNodes(db, schema, time.Unix(int64(beginTime), 0).Format(layout), time.Unix(int64(endTime), 0).Format(layout), digest)
```
breeswish, line 182: ditto.

pkg/apiserver/statement/statement.go
baurine: I tried to pass the beginTime and endTime as time.Time and int64 type, it seems they work neither...do we need to change any  for this condition?

```
Where("summary_begin_time >= ? AND summary_end_time <= ?", beginTime, endTime).
```

breeswish: I think either `UNIX_TIMESTAMP()` function or `FROM_UNIXTIME()` function is needed.

Fullstop000:
@baurine Thank you for addressing comments

PR #349, opened 2020-04-14
Add segment bar

breeswish:
Segment bars can be used in cluster info to display multiple segments.


ui/packages/dashboard_components/src/Bar/Bar.tsx
```diff
@@ -1,61 +1,111 @@
...
+
+  if (min) {
+    min = clamp(min, 0, valuesSum)
+    if ((valuesSum - min) / capacity < 0.01) {
+      min = undefined
+    }
   }
-  if (max !== undefined) {
-    max = clamp(max, value, capacity)
+  if (max) {
```
baurine, line 62: We need to consider when max is zero situation.

PR #359, opened 2020-04-16
Statements settings

baurine:
What did:

- support refresh
- support change statement settings
- support sticky the column headers by ScrollablePanel

image


pkg/apiserver/statement/queries.go
```diff
@@ -17,15 +17,93 @@ import (
...
+		config.HistorySize = historySize
+	}
+
+	return &config, err
+}
+
+func UpdateStmtConfig(db *gorm.DB, config *Config) (err error) {
+	var sql string
+	sql = fmt.Sprintf("SET GLOBAL %s = ?", stmtEnableVar)
+	err = db.Exec(sql, config.Enable).Error
```
baurine, line 92: Tried `db.Exec("SET GLOBAL ? = ?", stmtEnableVar, config.Enable).Error` doesn't work, so the `?` only can be show in value position?

breeswish: yes, I think this is expected.

pkg/apiserver/statement/queries.go
```diff
@@ -17,15 +17,93 @@ import (
...
+type sqlVariable struct {
+	Name  string `gorm:"column:Variable_name"`
+	Value string `gorm:"column:Value"`
+}
+
+// How to get sql variables by GORM
+// https://github.com/jinzhu/gorm/issues/2616
+func querySQLIntVariable(db *gorm.DB, name string) (int, error) {
+	var variables []sqlVariable
+	err := db.Raw("SHOW GLOBAL VARIABLES LIKE ?", name).Scan(&variables).Error
```
breeswish, line 42: better to use:

```go
db.Raw(`SELECT @@GLOBAL.tidb_enable_stmt_summary AS value`).Pluck("value", &....)
```

https://gorm.io/docs/query.html#Pluck

ui/lib/apps/Statement/components/StatementsOverview.tsx
```diff
@@ -205,13 +201,12 @@ export default function StatementsOverview({
...
+        const res = await onFetchConfig(state.curInstance)
         if (res !== undefined) {
-          // TODO: set on or off according res
-          // dispatch({
-          //   type: 'change_statement_status',
-          //   payload: 'on'
-          // })
+          dispatch({
+            type: 'change_statement_status',
+            payload: res.enable ? 'on' : 'off',
```
breeswish, line 208: we can use `true` / `false` now

ui/lib/apps/Statement/pages/StatementsOverviewPage.tsx
```diff
@@ -50,33 +44,21 @@ export default function StatementsOverviewPage({
...
   return (
     <StatementsOverview
       onFetchInstances={queryInstance}
       onFetchSchemas={querySchemas}
       onFetchTimeRanges={queryTimeRanges}
       onFetchStmtTypes={queryStmtTypes}
       onFetchStatements={queryStatements}
-      onGetStatementStatus={queryStatementStatus}
-      onSetStatementStatus={updateStatementStatus}
       onFetchConfig={queryConfig}
```
breeswish, line 62: looks like we don't need to repeat these functions again. It's fine to call api client directly in the `<overview>`?

ui/lib/apps/Statement/translations/en.yaml
```diff
@@ -31,3 +31,23 @@ statement:
...
+    close_statement: Close SQL statements
+    close_statement_warning: Are you sure to close this feature? All history records will be cleared after closing it!
+    switch: Open Statements
+    refresh_interval: Data Collect Interval
+    keep_duration: Data Keep Duration
+  actions:
+    save: Save
+    close: Close
+    cancel: Cancel
+  time:
```
breeswish, line 51: I have a lazy idea.. We can use `getValueFormat` / prettyMs to display time. We can avoid writing min / day translations again and again. Actually there is no need to translate the time unit to Chinese since everyone knows it.

ui/lib/apps/Statement/components/StatementSettingForm.tsx
```diff
@@ -0,0 +1,213 @@
...
+  onConfigUpdated: () => any
+}
+
+type InternalStatementConfig = StatementConfig & {
+  keep_duration: number
+  max_refresh_interval: number
+  max_keep_duration: number
+}
+
+const REFRESH_INTERVAL_MARKS = {
```
breeswish, line 32: `REFRESH_INTERVAL_MINUTES = [1, 5, ...]`

pkg/apiserver/statement/queries.go
baurine: Oh, good to know that we can use `SELECT @@GLOBAL.tidb_enable_stmt_summary ...` grammar to get the variable, I will try it. Have tried to use `Pluck` and `SHOW GLOBAL ...` together but failed before.

ui/lib/apps/Statement/components/StatementsOverview.tsx
baurine: make sense, will update it soon.

ui/lib/apps/Statement/pages/StatementsOverviewPage.tsx
baurine: Yep, will tidy them and remove all unnecessary codes in the next PR.

ui/lib/apps/Statement/translations/en.yaml
baurine: Ok, let me have a try.

pkg/apiserver/statement/queries.go
baurine: done!

ui/lib/apps/Statement/translations/en.yaml
baurine: `getValueFormat` converts "17  day" to "2 week" or "2.4 week", I don't think it is ok for me, and I think it is acceptable to fix the unit to "day" in this situation. I think it is fine to fix the unit to "min" and "day" in this situation.

ui/lib/apps/Statement/components/StatementSettingForm.tsx
baurine: marks prop for Slider  should be an object, not accept an array, so I add an util method to convert the arr to object:

```js
const convertArrToObj = (arr: number[]) =>
  arr.reduce((acc, cur) => {
    acc[cur] = cur
    return acc
  }, {})

const REFRESH_INTERVAL_MARKS = convertArrToObj([1, 5, 15, 30, 60])
const KEEP_DURATION_MARKS = convertArrToObj([1, 2, 5, 10, 20, 30]
```

pkg/apiserver/statement/queries.go
```diff
@@ -17,15 +17,89 @@ import (
...
+// How to get sql variables by GORM
+// https://github.com/jinzhu/gorm/issues/2616
+func querySQLIntVariable(db *gorm.DB, name string) (int, error) {
+	var values []string
+	sql := fmt.Sprintf("SELECT @@GLOBAL.%s as value", name) // nolints
+	err := db.Raw(sql).Pluck("value", &values).Error
+	if err != nil {
+		return 0, err
+	}
+	strVal := values[0]
```
breeswish, line 42: Maybe need to handle `len(values) == 0` to avoid panic

pkg/apiserver/statement/queries.go
```diff
@@ -17,15 +17,89 @@ import (
...
+}
+
+func QueryStmtConfig(db *gorm.DB) (*Config, error) {
+	config := Config{}
+
+	enable, err := querySQLIntVariable(db, stmtEnableVar)
+	if err != nil {
+		return nil, err
+	}
+	config.Enable = enable != 0
```
breeswish, line 60: better to check `enable == -1 || enable == 1`

breeswish:
The code needs some little refinement after Statement default values feature lands in TiDB.

PR #366, opened 2020-04-17
Filter columns

baurine:
What did:

- Support show or hide some columns
- Support show full or ellipsis SQL
- Refine tooltip for column headers
- Refine tooltip wording

Thanks for the great `CardTableV2.renderColumnVisibilitySelection()` method.

More metrics will add in the later PR.

image


ui/lib/components/CardTableV2/index.tsx
```diff
@@ -70,28 +70,22 @@ function renderColumnVisibilitySelection(
...
+        }
+        onChange({
+          ...visibleColumnKeys!,
+          [column.key]: e.target.checked,
+        })
+      }}
+    >
+      {column.name}
+    </Checkbox>
+  ))
```
baurine, line 88: The original implementation can't apply the correct style inside the `<Menu>`, even change the `<div key={column.key}>` to `<Menu.Item .../>`. the padding or margin will be lost, not sure why.

PR #378, opened 2020-04-20
statement: Add new detail backends

breeswish:

Support most of the statement aggregation fields.


ui/lib/apps/Statement/components/StatementsTable.tsx
```diff
@@ -190,9 +107,9 @@ export default function StatementsTable({
   const { t } = useTranslation()
   const navigate = useNavigate()
   const [items, setItems] = useState(statements)
-  const maxs = useMax(statements)
-  const [columns, setColumns] = useState(() =>
-    tableColumns(t, concise || false, maxs, onColumnClick)
+  // const maxs = useMax(statements)
+  const [columns, setColumns] = useState(
+    tableColumns(t, concise || false, statements, onColumnClick)
```
baurine, line 112: I think `tableColumns()` is a little expensive, so use `useState(()=> ...)` can avoid call `tableColumns()` in every render.

breeswish: Indeed! This can be optimized. Feel free to open a PR later.

breeswish:
@baurine Nice catch! I was developing the detail page locally so that I was not aware of this broken issue. I will fix it in later PRs before releasing new version.

PR #357, opened 2020-04-16
slow_query: add slow query detail page

Deardrops:
Preview:

Detail page:

PR #398, opened 2020-04-21
slow_query: add Slow Query List and Detail Page

baurine:
what did:

- Show slow query list
- Support sort by quert_time and memory_max columsn
- Support refresh
- Support filter by time range / schemas / search text

Some code is not clean, will be refined later.

Remain: 

- [x] Add limit param
- [x] Save context after jumping to the detail page
- [ ] i18n

@Deardrops 

Related PR: #357 

Detail page preview:


pkg/apiserver/slowquery/queries.go
```diff
@@ -0,0 +1,157 @@
...
+	if params.DB != "" {
+		dbs := strings.Split(params.DB, ",")
+		tx = tx.Where("DB IN (?)", dbs)
+	}
+
+	order := params.OrderBy
+	if params.DESC {
+		// FIXME
+		// but grom.Expr("? DESC", order) doesn't work
+		tx = tx.Order(fmt.Sprintf("%s desc", order))
```
breeswish, line 131: we need to verify order is one of the field. Otherwise there will be SQL injection problems. You can use reflection as in statement app.

pkg/apiserver/slowquery/queries.go
```diff
@@ -0,0 +1,157 @@
...
+	if !b.Time.IsZero() {
+		b.Timestamp = float64(b.Time.UnixNano()) / 1e9
+	}
+	return
+}
+
+type QueryRequestParam struct {
+	LogStartTS int64  `json:"logStartTS" form:"logStartTS"`
+	LogEndTS   int64  `json:"logEndTS" form:"logEndTS"`
+	DB         string `json:"db" form:"db"`
```
breeswish, line 103: `string[]` could work as well

pkg/apiserver/slowquery/queries.go
```diff
@@ -0,0 +1,157 @@
...
+		// FIXME
+		// but grom.Expr("? DESC", order) doesn't work
+		tx = tx.Order(fmt.Sprintf("%s desc", order))
+	} else {
+		tx = tx.Order(fmt.Sprintf("%s asc", order))
+	}
+
+	var results []Base
+
+	err := tx.Limit(params.Limit).Find(&results).Error
```
breeswish, line 138: We need to set a maximum limit (i.e. 1000) to avoid searching too many slow logs.

ui/lib/apps/SlowQuery/components/List.tsx
```diff
@@ -0,0 +1,209 @@
...
+        setSlowQueryList(res.data || [])
+      }
+    }
+    getSlowQueryList()
+  }, [curTimeRange, curSchemas, orderBy, desc, searchText, limit, refreshTimes])
+
+  // TODO: refine
+  const location = useLocation()
+  useEffect(() => {
+    if (location.search === '?from=detail') {
```
breeswish, line 103: when will we need it?

ui/lib/apps/SlowQuery/components/List.tsx
```diff
@@ -0,0 +1,209 @@
...
+                  {item}
+                </Option>
+              ))}
+            </Select>
+            <Search
+              placeholder="包含文本"
+              value={searchText}
+              onChange={(e) => setSearchText(e.target.value)}
+            />
+            <InputNumber
```
breeswish, line 181: prefer to be a dropdown select providing some common selections like `[100, 200, 500, 1000]` because accurate number of slow log limit is not very useful

ui/lib/apps/SlowQuery/components/TimeRangeSelector.module.less
```diff
@@ -0,0 +1,41 @@
+@import '~antd/es/style/themes/default.less';
+
+.dropdown_content_container {
+  background-color: @select-dropdown-bg;
```
breeswish, line 4: 👍

pkg/apiserver/slowquery/queries.go
```diff
@@ -0,0 +1,157 @@
...
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/jinzhu/gorm"
+)
+
+const (
+	MysqlTimeLayout = "2006-01-02 15:04:05.999999"
```
breeswish, line 25: maybe no need any more

pkg/apiserver/slowquery/queries.go
```diff
@@ -0,0 +1,157 @@
...
+	RequestCount uint   `gorm:"column:Request_count" json:"request_count"`
+	ProcessKeys  uint   `gorm:"column:Process_keys" json:"process_keys"`
+	TotalKeys    uint   `gorm:"column:Total_keys" json:"total_keys"`
+	CopProcAddr  string `gorm:"column:Cop_proc_addr" json:"cop_proc_addr"`
+	CopWaitAddr  string `gorm:"column:Cop_wait_addr" json:"cop_wait_addr"`
+}
+
+func (b *Base) AfterFind() (err error) {
+	if !b.Time.IsZero() {
+		b.Timestamp = float64(b.Time.UnixNano()) / 1e9
```
breeswish, line 95: you can `select *, unix_timestamp(time) as timestamp` instead

pkg/apiserver/slowquery/service.go
```diff
@@ -0,0 +1,123 @@
...
+	db            *dbstore.DB
+	tidbForwarder *tidb.Forwarder
+}
+
+func NewService(config *config.Config, tidbForwarder *tidb.Forwarder, db *dbstore.DB) *Service {
+	return &Service{config: config, db: db, tidbForwarder: tidbForwarder}
+}
+
+func Register(r *gin.RouterGroup, auth *user.AuthService, s *Service) {
+	endpoint := r.Group("/slowquery")
```
breeswish, line 42: slow_query

pkg/apiserver/slowquery/service.go
```diff
@@ -0,0 +1,123 @@
...
+	endpoint.GET("/list", s.listHandler)
+	endpoint.GET("/detail", s.detailhandler)
+}
+
+type InfoResponse struct { //nolint:golint
+	Version    utils2.VersionInfo `json:"version"`
+	PDEndPoint string             `json:"pd_end_point"`
+}
+
+type DatabaseResponse = []string
```
breeswish, line 54: maybe not needed?

ui/lib/apps/SlowQuery/components/List.tsx
baurine: back from the detail page, we need to recover the previous search options, we can use Context as in the statements app, but it needs more code, so I want to try another simpler way.

breeswish: We can save options in the URL, as it also suffers from page refresh.

baurine: Got what you mean, just want to try it, and should we do like this or store in session storage as you mentioned in #411

baurine: I prefer to use session storage, so it also can work when switch apps, how do you think about it?

PR #405, opened 2020-04-22
Add translations to slow query

breeswish:


ui/lib/apps/SlowQuery/components/Detail.tsx
```diff
@@ -136,30 +131,32 @@ function DetailPage() {
...
+              >
+                <TabCopr data={data} />
+              </CardTabs.TabPane>
+              <CardTabs.TabPane tab={t('slow_query.detail.tabs.txn')} key="txn">
+                <TabTxn data={data} />
+              </CardTabs.TabPane>
+            </CardTabs>
           </>
         )}
       </Head>
```
Deardrops, line 159: All components in `<Head>` element?

breeswish: It depends on whether or not there should be paddings.

PR #408, opened 2020-04-22
slow_query: improvement slow query page

Deardrops:
- [x] Order by `Timestamp` by default
- [x] Support search text case-insensitive
- [x] Auto do search when click search button or press `enter`
- [x] Move hook function to `render()` function
- [x] Increase upper bound for avoiding missing rounded timestamp.
- [x] Store all params in router query

### DIFF
--- a/ui/lib/apps/SlowQuery/components/List.tsx
+++ b/ui/lib/apps/SlowQuery/components/List.tsx
@@ -101,9 +101,7 @@ function List() {
           searchText
         )
       setLoading(false)
-      if (res?.data) {
-        setSlowQueryList(res.data || [])
-      }
+      setSlowQueryList(res.data || [])
     }
     getSlowQueryList()
     const qs = List.buildQuery({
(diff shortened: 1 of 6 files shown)


ui/lib/apps/SlowQuery/components/List.tsx
```diff
@@ -172,10 +178,7 @@ export default function List() {
                 </Option>
               ))}
             </Select>
-            <Search
-              value={searchText}
-              onChange={(e) => setSearchText(e.target.value)}
-            />
+            <Search onSearch={handleSearch} />
```
baurine, line 181: I did this before, but this way can't recover the search text when comes back from detail page.

Deardrops: Now we store all params in router query, this issue resolved, PTAL

ui/lib/apps/SlowQuery/components/TimeRangeSelector.tsx
```diff
@@ -21,9 +21,10 @@ export interface TimeRange {
 
 export const DEF_TIME_RANGE: TimeRange = {
   recent: 30,
-  begin_time: 0,
-  end_time: 0,
+  begin_time: dayjs().unix() - 30 * 60,
+  end_time: dayjs().unix(),
 }
```
baurine, line 26: yep, this part logic has some problem, I will try to figure it out after this PR is merged.

PR #413, opened 2020-04-23
SlowQuery: use SessionStorage to save search options

### LINKED ISSUE #411: Remember Query Parameters by Session
## Feature Request

For query parameters, they can be remembered in Session Storage (instead of Local Storage), so that it suffers from refresh, as well as history go back.

image

It should not be stored in local storage, to avoid user being trapped (e.g. repeatly meeting errors) with some irregular settings. When storing in session storage, user can at least close, open browser again to restore.

baurine:
Fix #411 , will do it for the statements page as well later.

### DIFF
--- a/ui/lib/apps/SlowQuery/components/Detail.tsx
+++ b/ui/lib/apps/SlowQuery/components/Detail.tsx
@@ -53,7 +53,7 @@ function DetailPage() {
       <Head
         title={t('slow_query.detail.head.title')}
         back={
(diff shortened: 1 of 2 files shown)


baurine:
@breeswish @Deardrops  PTAL

baurine:
seems @Deardrops resolved it by another solution, so I will close this.

breeswish:
Indeed. @Deardrops 's solution is also very nice. We also need to do the same for statement.

PR #414, opened 2020-04-23
ui: Add Metrics

breeswish:
image

Some implementation is dirty hack. To be refined.


ui/lib/apps/Overview/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+                {
+                  query:
+                    'histogram_quantile(0.99, sum(rate(tidb_server_handle_query_duration_seconds_bucket[1m])) by (le))',
+                  name: '99%',
+                },
+                {
+                  query:
+                    'histogram_quantile(0.9, sum(rate(tidb_server_handle_query_duration_seconds_bucket[1m])) by (le))',
+                  name: '90%',
+                },
```
baurine, line 92: as the metrics get more and more, maybe we can extract all metric configs to an individual js or JSON file.

PR #402, opened 2020-04-22
enabled keyviz by default && bump ui

HunDunDM:
Signed-off-by: Zheng Xiangsheng <hundundm@gmail.com>

Since it is not yet possible to control the start of keyviz through the front end, keyviz is currently started by default.

PR #421, opened 2020-04-23
keyviz: support to modify the config through the front-end && bump ui

HunDunDM:
* revert #402 (so keyviz is disabled by default)
* support to modify the config through the front-end
* bump ui

### DIFF
--- a/pkg/config/dynamic_config.go
+++ b/pkg/config/dynamic_config.go
@@ -13,13 +13,8 @@
 
 package config
 
-const (
-	StatusEnabled  = "enabled"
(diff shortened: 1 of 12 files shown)


baurine:
LGTM, just one point, do we need to wrap every method by `useCallback()`, does it low the performance without `useCallback()`, I use `useCallback()` rarely for other pages.

And btw, it has many compile warnings for `React Hook useEffect has a missing dependency`.

breeswish:
@baurine Agree. We can do it in May.

PR #423, opened 2020-04-23
slow query: save search options by session storage

baurine:


baurine:
@breeswish @Deardrops PTAL.

Will apply for statements as well later.

PR #426, opened 2020-04-24
slow_query: columns selector

baurine:
What did:

- Extract ColumnsSelector
- Support show/hide some columns for slow query table

image

Remain:

- [x] i18n


breeswish:
We can display the following columns in extra (default not display):

- TiDB instance Address
- Is Success
- SQL Template ID
- Execution Database
- Parse Time
- Compile Time
- Coprocessor Process Time
- Start Timestamp

In future (GA), better to figure out a way to let user able to select any columns available (and maybe not pulling back unneeded columns as well)

baurine:
> We can display the following columns in extra (default not display):
> 
> * TiDB instance Address
> * Is Success
> * SQL Template ID
> * Execution Database
> * Parse Time
> * Compile Time
> * Coprocessor Process Time
> * Start Timestamp
> 
> In future (GA), better to figure out a way to let user able to select any columns available (and maybe not pulling back unneeded columns as well)

Ok, will add them later in today.

baurine:
> In future (GA), better to figure out a way to let user able to select any columns available (and maybe not pulling back unneeded columns as well)

just got what you mean, you mean in the backend, we just select the columns which be chosen to show, right? er, it seems we need GraphQL (just kidding) or add a more query parameter named fields (likes the Facebook API).

baurine:
WX20200424-221641@2x

done! @breeswish @Deardrops PTAL

not sure whether the order of the columns is ok.

ui/lib/components/ColumnsSelector/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+  )
+
+  return (
+    <Dropdown
+      placement="bottomRight"
+      visible={dropdownVisible}
+      onVisibleChange={setDropdownVisible}
+      overlay={dropdownMenus}
+    >
+      <div style={{ cursor: 'pointer' }}>
```
breeswish, line 99: maybe span is better

ui/lib/components/ColumnsSelector/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+                  })
+              }}
+            >
+              {column.name}
+            </Checkbox>
+          </Menu.Item>
+        ))}
+
+      {footExtra && <Menu.Divider />}
+      {footExtra && <Menu.Item key="foot">{footExtra}</Menu.Item>}
```
breeswish, line 78: We can follow a custom render style like `<Select>` component: https://codesandbox.io/s/tmo8v It should be more flexible (for supporting render arbitrary contents)

ui/lib/components/ColumnsSelector/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  const dropdownMenus = (
+    <Menu>
+      {headExtra && <Menu.Item key="head">{headExtra}</Menu.Item>}
+      {headExtra && <Menu.Divider />}
+
+      {columns
+        .filter((c) => c.key !== 'dummy')
+        .map((column) => (
+          <Menu.Item key={column.key}>
```
breeswish, line 61: BTW actually it might not be very good to use `<Menu.Item>` to render the column dropdown (or maybe the style of `Checkbox` need to be tuned), since there are many places that have hover actions but not clickable. 

click

ui/lib/components/ColumnsSelector/index.tsx
baurine: ok, let me try it.

ui/lib/components/ColumnsSelector/index.tsx
baurine: nice! try it soon!

ui/lib/components/ColumnsSelector/index.tsx
baurine: wow, so considerate, yep, if we can use `Select` to replace `Dropdown`. the Dropdown only can use `Menu` component as its overlay.

ui/lib/components/ColumnsSelector/index.tsx
baurine: or maybe we can try `Popover`, I found the `Select` not easy to control the down arrow icon position, the example in https://preview.pro.ant.design/list/table-list uses `Popover`.

breeswish: Good idea! we can just follow the ant design pro's way.

ui/lib/components/ColumnsSelector/index.tsx
breeswish: Oh, actually I didn't mean to use `<Select>`. I think Dropdown is enough. My previous comment is saying that `<Select>` provides a custom render property. We can follow the same idea, providing a custom render property as well to render the tail item. In this way, the tail item is not limited to only one menu item, but can be anything.

ui/lib/components/ColumnsSelector/index.module.less
```diff
@@ -0,0 +1,7 @@
+@import '~antd/es/style/themes/default.less';
+
+.foot_container {
```
breeswish, line 3: Maybe you can use https://ant.design/components/divider/

ui/lib/components/ColumnsSelector/index.tsx
```diff
@@ -51,54 +51,34 @@ export default function ColumnsSelector({
...
-                  onChange({
-                    ...visibleKeys,
-                    [column.key]: e.target.checked,
-                  })
-              }}
-            >
-              {column.name}
-            </Checkbox>
-          </Menu.Item>
+          <Checkbox
```
breeswish, line 58: It can be

```
Space
  Checkbox
  Checkbox
Footer
  Divider
  Space
    Checkbox
```

Then maybe no extra styles are needed

baurine: I tried this layout with Divider before, because all of them are inside the div with class `ant-popover-inner` which has default 16px padding, we still need to add extra style to Divider to dismiss its default padding.

Later I found the Popover native title part implement its bottom divider by border-bottom, so I just copy it...

baurine:
Support the select all and none, reset.

ui/lib/components/ColumnsSelector/index.tsx
baurine: it works!

ui/lib/apps/SlowQuery/components/List.tsx
```diff
@@ -41,6 +44,13 @@ export function getDefSearchOptions(): ISearchOptions {
   }
 }
 
+const defColumnKeys = {
```
Deardrops, line 47: Declare the type of `defColumnKeys` before it?

Deardrops, line 47: Declare the type of `defColumnKeys` before it?

ui/lib/apps/Statement/components/StatementsOverview.tsx
```diff
@@ -23,18 +10,28 @@ import {
...
 import styles from './styles.module.less'
 
 const { Option } = Select
 
-const VISIBLE_COLUMN_KEYS = 'statement_overview_visible_column_keys'
-const SHOW_FULL_SQL = 'statement_overview_show_full_sql'
+const VISIBLE_COLUMN_KEYS = 'statement_visible_column_keys'
+const SHOW_FULL_SQL = 'statement_show_full_sql'
+
+const defColumnKeys = {
```
Deardrops, line 27: ditto

ui/lib/apps/SlowQuery/components/List.tsx
baurine: done!

ui/lib/apps/Statement/components/StatementsOverview.tsx
baurine: done!

PR #428, opened 2020-04-26
slow_query: fix bugs in slow query page

Deardrops:
Bugfix:
- [x] Fix bug in case-insensitive keyword search
- [x] Fix bug in time range selector, now we use recent 30 minutes time range instead of **fixed** 30 minutes time range.

Improvement:
- [x] After pingcap/tidb#16806 merged, now we can check if `cluster_slow_query.time` is equal to given time string by `=` operator.

### DIFF
--- a/ui/lib/apps/SlowQuery/components/List.tsx
+++ b/ui/lib/apps/SlowQuery/components/List.tsx
@@ -67,7 +67,8 @@ function List() {
       setLoading(true)
       const recentMins = searchOptions.timeRange.recent
       if (recentMins > 0) {
-        // beginTime & endTime may outdated, update to recent time range
+        // beginTime & endTime is fixed value,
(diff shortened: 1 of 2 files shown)


ui/lib/apps/SlowQuery/components/List.tsx
```diff
@@ -64,6 +65,18 @@ function List() {
...
+        // beginTime & endTime is fixed value,
+        // so update them to the time range from now
+        const now = dayjs().unix()
+        const beginTime = now - recentMins * 60
+        searchOptions.timeRange = {
+          recent: recentMins,
+          begin_time: beginTime,
+          end_time: now,
+        }
+      }
```
baurine, line 79: I will move this part logic to TimeSelector and supply a util method likes `updateTimeRange()` later.

PR #429, opened 2020-04-26
Extract Toolbar for KeyViz / SlowQuery / Statement

baurine:
What did:

- Extract Toolbar component
- Use Toolbar for KeyViz / SlowQuery / Statements

Before:

WX20200426-171548@2x

After:

WX20200426-171851@2x


baurine:
@breeswish @HunDunDM PTAL

breeswish:
I see that we have the similar pattern in multiple places, like statements, slow query and keyviz as well. How about introducing our own `<Toolbar>` component? (we can make `size="medium" by default, as well as introducing margin-bottom for all)

baurine:
> I see that we have the similar pattern in multiple places, like statements, slow query and keyviz as well. How about introducing our own `<Toolbar>` component? (we can make `size="medium" by default, as well as introducing margin-bottom for all)

Good advice! will update it later!

baurine:
hi @breeswish , extract the Toolbar component, PTAL!

cc @Deardrops @HunDunDM

ui/lib/components/Toolbar/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+  React.Children.forEach(children, (child) => {
+    if (!React.isValidElement(child) || child.type !== Space) {
+      console.error('Toolbar children only can be Space component')
+    }
+  })
+
+  return (
+    <div className={c} {...rest}>
+      {React.Children.map(children, (child, idx) => {
+        // https://stackoverflow.com/questions/42261783/how-to-assign-the-correct-typing-to-react-cloneelement-when-giving-properties-to
```
Deardrops, line 21:
```suggestion
        // https://stackoverflow.com/questions/42261783
```

ui/lib/components/Toolbar/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+import cx from 'classnames'
+import { Space } from 'antd'
+
+import styles from './index.module.less'
+
+export default function Toolbar(props: React.HTMLAttributes<HTMLDivElement>) {
+  const { className, children, ...rest } = props
+  const c = cx(className, styles.toolbar_container)
+
+  // https://stackoverflow.com/questions/27366077/only-allow-children-of-a-specific-type-in-a-react-component
```
Deardrops, line 11: ditto

ui/lib/components/Toolbar/index.tsx
baurine: nice!

PR #430, opened 2020-04-26
time range selector: move to lib/components and refine

baurine:
What did:

- Move TimeRangeSelector to lib/components
- Refine TimeRangeSelector


baurine:
@breeswish @Deardrops PTAL

ui/lib/components/TimeRangeSelector/index.tsx
```diff
@@ -1,60 +1,69 @@
...
-import { Moment } from 'moment'
+import styles from './index.module.less'
 
 const { RangePicker } = DatePicker
 
 const RECENT_MINS = [30, 60, 3 * 60, 6 * 60, 12 * 60, 24 * 60]
 
 export interface TimeRange {
-  recent: number
+  recent: number // 0 means absolute time range, else means relative time range
```
breeswish, line 17: prefer to add unit to the property name

ui/lib/components/TimeRangeSelector/index.tsx
```diff
@@ -1,60 +1,69 @@
...
   begin_time: number
```
breeswish, line 18: prefer to use camelCase

ui/lib/components/TimeRangeSelector/index.tsx
```diff
@@ -1,60 +1,69 @@
...
+    begin_time: now - mins * 60,
     end_time: now,
   }
 }
 
+export function getDefTimeRange(): TimeRange {
+  return calcTimeRange(30)
+}
+
+export function refreshTimeRange(timeRange: TimeRange): TimeRange {
```
breeswish, line 35: I prefer not to provide `beginTime` and `endTime` at all when this function is not called, so that developer will *never* remember to call this transformer function (since no absolute time will be provided). i.e. When a recent time is selected, there is no way to access the absolute time from the value directly. An absolute time can be only accessed via this transform function. Additionally, since they will be two different TypeScript types, we can even ensure the correctness at compile time.

ui/lib/components/TimeRangeSelector/index.tsx
```diff
@@ -1,60 +1,69 @@
...
   return {
-    recent: 30,
-    begin_time: now - 30 * 60,
+    recent: mins,
+    begin_time: now - mins * 60,
     end_time: now,
   }
 }
 
+export function getDefTimeRange(): TimeRange {
```
breeswish, line 31: Looks like normally components do not provide this kind of function. How about using `value == null` to denote a default selection?

ui/lib/components/TimeRangeSelector/index.tsx
baurine: It is! I was tangled with it. will update it after finishing the statement & slow query integration issue.

Deardrops: Agree with it.

baurine:
@breeswish @Deardrops , address the comments, PTAL

ui/lib/components/TimeRangeSelector/index.tsx
```diff
@@ -1,30 +1,48 @@
...
 
-export function getDefTimeRange(): TimeRange {
-  const now = dayjs().unix()
-  return {
-    recent: 30,
-    begin_time: now - 30 * 60,
-    end_time: now,
+export interface RangeTime {
+  type: 'absolute'
+  value: [number, number] // unit: seconds
```
Deardrops, line 30: Can we use type `TimeRange` from 'antd' here?

baurine: Does 'antd' have? I don't find it ...

Deardrops: In antd 4.0, it changed to type `moment[]`, so this is ok here.

PR #433, opened 2020-04-27
cluster_info: fix UI bug in cluster_info

Deardrops:
Bugfix:
- [x] Remove extra space on the right of page in cluster info page.

image


baurine:
How does it look like before and after? can you attach some screenshots?

Deardrops:
> 
> 
> How does it look like before and after? can you attach some screenshots?

@baurine updated

PR #435, opened 2020-04-27
Integrate statement and slow query

baurine:
What did:

- Show slow query list in the statement detail page
- Extract useSlowQuery()

image


baurine:
@breeswish @Deardrops PTAL.

this PR merges to `choose-columns-component` temporary for easy review, it will change to merge to `master` after `choose-columns-component` is merged.

PR #437, opened 2020-04-28
Integrate statement and slow query

baurine:
What did:

- Show slow query list in the statement detail page
- Extract useSlowQuery()

image


baurine:
@breeswish @Deardrops PTAL.

PR #443, opened 2020-05-06
use antd Result component to show message for feature closed status

### LINKED ISSUE #422: Use <Result> Component
## Feature Request

https://ant.design/components/result/

We can use the Result component comes from Ant Design in KeyViz and Statement when corresponding switch is closed

baurine:
close #422 

screenshots:

企业微信20200506115255

it seems the Result component recommends to put all contents of the subtitle in one line, not sure whether it is ok.


ui/lib/apps/KeyViz/components/KeyViz.tsx
```diff
@@ -220,16 +220,18 @@ const KeyViz = (props) => {
...
-        <p>{t('keyviz.settings.disabled_desc_line_2')}</p>
-      </div>
-      <Button type="primary" onClick={openSettings}>
-        {t('keyviz.settings.open_setting')}
-      </Button>
-    </div>
+    <Result
+      title={t('keyviz.settings.disabled_desc_title')}
+      subTitle={
+        t('keyviz.settings.disabled_desc_line_1') +
```
breeswish, line 226: Seems that we can merge them together, to be something like:

```yaml
keyviz:
  settings:
    disabled_result:
      title: xxx
      subTitle: xxx
```

In this way it matches the `<Result>` better.

baurine: Good point! will update it soon!

PR #448, opened 2020-05-06
Remember columns width

### LINKED ISSUE #444: The resized width of columns will be reset after sorting
## Bug Report

The resized width of columns will be reset after sorting, we should remember their width.

baurine:
resolve #444 

What did:

- Use ShimmeredDetailsList to represent loading status for not re-mount the DatailsList when reloading data
- Keep changed columns width when reloading data in the current page
- Show the existed old data when reloading

Failed to do:

- Persist the width of the columns.

  According to the issue https://github.com/microsoft/fluentui/issues/9287 , this feature isn't supported by DatailsList.

Do you guys have any idea?

### DIFF
--- a/ui/lib/apps/Statement/pages/List/index.tsx
+++ b/ui/lib/apps/Statement/pages/List/index.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react'
 import { Select, Space, Tooltip, Drawer, Button, Checkbox, Result } from 'antd'
-import { useLocalStorageState } from '@umijs/hooks'
+import { useLocalStorageState, useDebounceFn } from '@umijs/hooks'
 import { SettingOutlined, ReloadOutlined } from '@ant-design/icons'
 import { ScrollablePane } from 'office-ui-fabric-react/lib/ScrollablePane'
 import { IColumn } from 'office-ui-fabric-react/lib/DetailsList'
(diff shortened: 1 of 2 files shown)


breeswish:
Actually if you only want to prevent the re-mount, maybe consider `display: none`?

baurine:
> Actually if you only want to prevent the re-mount, maybe consider `display: none`?

Yep, that's a way! but I also prefer to show loading status inside the table a bit.

ui/lib/apps/Statement/pages/List/index.tsx
```diff
@@ -52,6 +54,22 @@ export default function StatementsOverview() {
...
+    newWidth?: number,
+    _columnIndex?: number
+  ) {
+    setColumnsWidth({
+      ...columnsWidth,
+      [column!.key]: newWidth,
+    })
+  }
+  const { run } = useDebounceFn(handleColumnResize, 500)
+
```
baurine, line 72: These don't work code will be reverted in the next PR.

ui/lib/components/CardTableV2/index.tsx
```diff
@@ -78,12 +79,30 @@ function CardTableV2(props: ICardTableV2Props) {
...
+                width: `${columnsWidth[c.key]}px`,
+              }, // doesn't work
+              currentWidth: columnsWidth[c.key], // doesn't work
+              calculatedWidth: columnsWidth[c.key], // doesn't work
+            }
+          : c
+      )
+    }
+    console.log(newColumns)
+    return newColumns
```
baurine, line 104: These don't work code will be reverted in the next PR.

PR #452, opened 2020-05-07
Refine order and desc

### LINKED ISSUE #447: Move orderBy and desc logic down to CardTableV2
## Feature Request

baurine:
resolve #447 

What did:

- Revert doesn't work code for persisting columns width
- Move all logic about sorting down to CardTableV2 to reduce duplicated logic
- Fix the wrong usage for hooks so now we can cache the tableColumns
- Fix the TimeRangeSelector doesn't work as expected

### DIFF
--- a/ui/lib/apps/SlowQuery/components/SlowQueriesTable.tsx
+++ b/ui/lib/apps/SlowQuery/components/SlowQueriesTable.tsx
@@ -1,50 +1,26 @@
-import React, { useEffect } from 'react'
+import React, { useMemo } from 'react'
 import { useNavigate } from 'react-router-dom'
+import { IColumn } from 'office-ui-fabric-react/lib/DetailsList'
 import { CardTableV2, ICardTableV2Props } from '@lib/components'
 import { SlowqueryBase } from '@lib/client'
-import { IColumn } from 'office-ui-fabric-react/lib/DetailsList'
 import * as useColumn from '@lib/utils/useColumn'
 
 import * as useSlowQueryColumn from '../utils/useColumn'
 import DetailPage from '../pages/Detail'
 
-function tableColumns(
-  rows: SlowqueryBase[],
-  onColumnClick: (ev: React.MouseEvent<HTMLElement>, column: IColumn) => void,
-  orderBy: string,
-  desc: boolean,
-  showFullSQL?: boolean
-): IColumn[] {
+function tableColumns(rows: SlowqueryBase[], showFullSQL?: boolean): IColumn[] {
   return [
     useSlowQueryColumn.useSqlColumn(rows, showFullSQL),
     useSlowQueryColumn.useDigestColumn(rows),
     useSlowQueryColumn.useInstanceColumn(rows),
     useSlowQueryColumn.useDBColumn(rows),
     useSlowQueryColumn.useSuccessColumn(rows),
-    {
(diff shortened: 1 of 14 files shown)


ui/lib/apps/SlowQuery/utils/useColumn.tsx
```diff
@@ -16,6 +16,13 @@ import { getValueFormat } from '@baurine/grafana-value-formats'
 import { max } from 'lodash'
 import { useTranslation } from 'react-i18next'
 
+function SuccessStatusBadge({ status }: { status: 'success' | 'error' }) {
```
baurine, line 19: Should change a name

baurine:
@breeswish @Deardrops PTAL, thanks!

ui/lib/apps/SlowQuery/components/SlowQueriesTable.tsx
```diff
@@ -53,49 +29,22 @@ function tableColumns(
...
   showFullSQL,
 
-  onChangeSort,
-  onGetColumns,
   ...restProps
 }: Props) {
   const navigate = useNavigate()
 
-  const columns = tableColumns(
+  const columns = useMemo(() => tableColumns(slowQueries, showFullSQL), [
```
breeswish, line 44: I suspect this violates the react hook principle since `tableColumns` is in fact a hook (or let's say, it contains hooks), and `useMemo()` causes hook calls being conditional

baurine: I doubt it as well, and in fact, it is true that we can't use `tableColumns()` inside the `useMemo()`, `useState(()=> {...})`, `useEffect()` before, else it will crash and warns us we can't use hooks inside them, that's because we call the `useTranslations()` directly inside a `useSuccessColumn()` method to make it become the hook, after moving the `useTranslation()` to the `ResultStatusBadge` function component, the `useSuccessColumn()` isn't the hook any more.

And although the name field returns a component `TextWithInfo.TransKey` which uses `useTranslations()` inside, but it won't make the `useXxxColumn()` method become the hook, this refreshed my understanding.

ui/lib/apps/SlowQuery/utils/useColumn.tsx
```diff
@@ -127,7 +134,9 @@ export function useDBColumn(
 export function useSuccessColumn(
   _rows?: { success?: number }[] // used for type check only
 ): IColumn {
-  const { t } = useTranslation()
+  // !! Wrong usage, hooks only can be used inside the Function Component
+  // !! Can't be used in normal function
```
baurine, line 138: This wording is not correct exactly, hooks can be used in the custom hook as well. we need to avoid `useXxxColumn()` to become the custom hook.

ui/lib/apps/SlowQuery/components/SlowQueriesTable.tsx
breeswish: Cool! The idea is brilliant. I suggest to remove the `useXxx` prefix to indicate that they are no longer hooks.

baurine: Agree, will update it soon!

baurine: There are too many `useXxxColumn()` methods, so I will merge this PR first, will tidy all `useXxxColumn()` in the next PR.

Issue #203, opened 2020-03-11
UCP: Support Case Insensitive Log Searching


breeswish:
## Description

Currently users input case sensitive regular expressions in log searching. We need to support case insensitive for easier use.

You need to change the code of TiDB, TiKV and PD.

## Score

* 300

## Mentor(s)

* @Deardrops 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Go
* Rust

SSebo:
Like add a checkbox on UI, and a param  indicate case on API ?  User can choose whether or not searcn case sentive? or just ignore the case?

breeswish:
@SSebo Just ignore the case.

weihanglo:
As far as I know, the current definition of log searching is upon [`Diagnostics` gRPC server interface](https://github.com/pingcap/kvproto/blob/0f5ffe4596777255953000469d6e2a1b9b84d991/proto/diagnosticspb.proto#L15-L20). And the corresponding [service implementations in TiKV](https://github.com/tikv/tikv/blob/cee0f390c18a6145d19f26b4f6b97d16d38dae7e/src/server/service/diagnostics.rs#L1135) and in PD/TiDB (both [rely on `sysutil`](https://github.com/pingcap/sysutil/blob/7480a118e997b97dfd9975db60d5e8cc0e1beffd/service.go#L61)) all accept arbitrary regular expression patterns, which means if we add prefix `(?i)` for both TiKV and `sysutil` regex patterns in `search_log` rpc call, all these services act as what we expect.

Not sure whether this is a suitable solution. @Deardrops what's your opinions?

Deardrops:
@weihanglo Nice idea, the 2nd way is more correct, because we do not change the origin semantic of `SearchLog` gRPC API. 
Have you tested this way locally?

weihanglo:
> @weihanglo Nice idea, the 2nd way is more correct, because we do not change the origin semantic of `SearchLog` gRPC API.
> Have you tested this way locally?

Tested with `tiup` and it just works. I'll send a PR later. Thanks!

PR #453, opened 2020-05-07
Support case-insensitive log searching

weihanglo:
UCP #203 

## What?

Support Case Insensitive Log Searching

## How

By modifying regexp patterns with an `(?i)` ignore-case flag before sending `SearchLogRequest` to gRPC service.

## Compatibility

- [x] Golang `regexp` package, which uses [`re2`](https://github.com/google/re2/wiki/Syntax) underneath the hood, supports inline `i` flag
- [x] Rust `regex` crate also [supports `i` flag](https://docs.rs/regex/1.3.7/regex/#grouping-and-flags)

## Screenshot

image

### DIFF
--- a/pkg/apiserver/logsearch/task.go
+++ b/pkg/apiserver/logsearch/task.go
@@ -187,7 +187,13 @@ func (t *Task) searchLog(client diagnosticspb.DiagnosticsClient, targetType diag
 	if t.model.Error != nil {
 		return
 	}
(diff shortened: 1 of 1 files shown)


weihanglo:
@Deardrops 
Also upload a screenshot. PTAL.

weihanglo:
> Could you also provide a screen shot for the TiKV?

@breeswish Here you are.

TiDB

image

TiKV

image

breeswish:
@all-contributors please add @weihanglo   for code

PR #454, opened 2020-05-07
instance_profiling: fix detail page crash

baurine:
Reason: row may be undefined

image

### DIFF
--- a/ui/lib/apps/InstanceProfiling/pages/Detail.js
+++ b/ui/lib/apps/InstanceProfiling/pages/Detail.js
@@ -134,7 +134,6 @@ export default function Page() {
         loading={isLoading && !data}
         columns={columns}
         items={data?.tasks_status || []}
-        getKey={(row) => row.id}
       />
     </div>
   )
--- a/ui/lib/apps/SlowQuery/components/SlowQueriesTable.tsx
+++ b/ui/lib/apps/SlowQuery/components/SlowQueriesTable.tsx
@@ -62,6 +62,7 @@ export default function SlowQueriesTable({
       columns={columns}
       items={slowQueries}
       onRowClicked={handleRowClick}
+      getKey={(row) => row?.digest}
     />
   )
 }
--- a/ui/lib/apps/Statement/components/StatementsTable.tsx
+++ b/ui/lib/apps/Statement/components/StatementsTable.tsx
@@ -66,6 +66,7 @@ export default function StatementsTable({
       columns={columns}
       items={statements}
       onRowClicked={handleRowClick}
+      getKey={(row) => row?.digest}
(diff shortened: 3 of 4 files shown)


ui/lib/apps/InstanceProfiling/pages/Detail.js
```diff
@@ -134,7 +134,6 @@ export default function Page() {
         loading={isLoading && !data}
         columns={columns}
         items={data?.tasks_status || []}
-        getKey={(row) => row.id}
```
baurine, line 137: `row => row?.id` works as well, but I prefer to just remove it, there is no warning after removing it in the console, even there are multiple rows.

baurine:
> I don't know why it works :)

Let me check the document deeper, actually, we don't add the getKey prop in other tables.

baurine:
According to the source code: https://github.com/microsoft/fluentui/pull/1347/files/c1fc1449ae064802e967b36022349179dbabb8b2#diff-92617c3c9fecec754c52f44e7204cec2R321 , if we don't supply the getKey prop, it will use index as the row key, but it is still weird if the row is undefined, the `getKey` shouldn't be called.

So `getKey` should be necessary when the rows need to sort, I will add it for StatementsTable and SlowQueryTable later.

breeswish:
Maybe the core problem is why `row` can be undefined?

baurine:
> Maybe the core problem is why `row` can be undefined?

Oh! good hint! I think it must the ShimmeredDetailsList create those empty rows, and it proved!

https://github.com/microsoft/fluentui/blob/master/packages/office-ui-fabric-react/src/components/DetailsList/ShimmeredDetailsList.base.tsx#L24

```
export class ShimmeredDetailsListBase extends React.Component<IShimmeredDetailsListProps, {}> {
  private _shimmerItems: null[];
  private _classNames: IProcessedStyleSet<IShimmeredDetailsListStyles>;

  constructor(props: IShimmeredDetailsListProps) {
    super(props);

    this._shimmerItems = props.shimmerLines ? new Array(props.shimmerLines) : new Array(SHIMMER_INITIAL_ITEMS);
  }
```

This line code doesn't judge the item whether null (they ever did, but don't know why it is removed in the master branch):

https://github.com/microsoft/fluentui/blob/master/packages/office-ui-fabric-react/src/components/List/List.tsx#L549

```
let itemKey = this.props.getKey ? this.props.getKey(item, index) : item && (item as any).key;
```

I guess the workaround is to use `row => row?.xxx`.

breeswish:
Considering current situation, I have some ideas about improving the table, as well as resolving some `<Skeleton>` related display problems. I will open a PR for this :)

PR #457, opened 2020-05-08
avoid duplicate keys for table rows

baurine:
Hi @breeswish , I forgot the slow queries may have the same digest, using `getKey={row => row?.digest}` will cause duplicate keys for slow query table, so I add more fields to avoid this. Although `row?.digest` works fine for statement table, I add more as well just in case.

Or we can wait for your PR which mentioned in https://github.com/pingcap-incubator/tidb-dashboard/pull/454#issuecomment-625242923 if it relates to this.

### DIFF
--- a/ui/lib/apps/SlowQuery/components/SlowQueriesTable.tsx
+++ b/ui/lib/apps/SlowQuery/components/SlowQueriesTable.tsx
@@ -62,7 +62,9 @@ export default function SlowQueriesTable({
       columns={columns}
       items={slowQueries}
       onRowClicked={handleRowClick}
-      getKey={(row) => row?.digest}
+      getKey={(row, idx) =>
+        `${row?.digest}_${row?.timestamp}_${row?.txn_start_ts}_${idx}`
+      }
     />
   )
 }
--- a/ui/lib/apps/SlowQuery/components/SlowQueriesTable.tsx
+++ b/ui/lib/apps/SlowQuery/components/SlowQueriesTable.tsx
@@ -62,9 +62,7 @@ export default function SlowQueriesTable({
       columns={columns}
       items={slowQueries}
       onRowClicked={handleRowClick}
-      getKey={(row, idx) =>
-        `${row?.digest}_${row?.timestamp}_${row?.txn_start_ts}_${idx}`
-      }
+      getKey={(row) => `${row?.digest}_${row?.timestamp}`}
     />
   )
 }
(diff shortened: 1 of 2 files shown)


ui/lib/apps/Statement/components/StatementsTable.tsx
```diff
@@ -66,7 +66,9 @@ export default function StatementsTable({
       columns={columns}
       items={statements}
       onRowClicked={handleRowClick}
-      getKey={(row) => row?.digest}
+      getKey={(row, idx) =>
+        `${row?.digest}_${row?.sum_latency}_${row?.max_latency}_${idx}`
```
breeswish, line 70: For statement table, the key might be digest + schema_name (not related_schemas)

breeswish, line 70: For statement table, the key should be digest + schema_name (not related_schemas)

ui/lib/apps/SlowQuery/components/SlowQueriesTable.tsx
```diff
@@ -62,7 +62,9 @@ export default function SlowQueriesTable({
       columns={columns}
       items={slowQueries}
       onRowClicked={handleRowClick}
-      getKey={(row) => row?.digest}
+      getKey={(row, idx) =>
+        `${row?.digest}_${row?.timestamp}_${row?.txn_start_ts}_${idx}`
```
breeswish, line 66: When there is re-order, the key is changed and it becomes no better than key=idx. I prefer to use digest+timestamp

baurine: Make sense, will update it soon.

ui/lib/apps/SlowQuery/components/SlowQueriesTable.tsx
```diff
@@ -62,7 +62,7 @@ export default function SlowQueriesTable({
       columns={columns}
       items={slowQueries}
       onRowClicked={handleRowClick}
-      getKey={(row) => row?.digest}
+      getKey={(row) => row && `${row.digest}_${row.timestamp}`}
```
baurine, line 65: `(row) => `${row?.digest}_${row?.timestamp}` will warning duplicated keys when loading because all keys are `undefined_undefined`, so change to `row && `${row.digest}_${row.timestamp}`

PR #465, opened 2020-05-09
UI refinements

breeswish:
1. Introduced AnimatedSkeleton as a replacement for Skeleton for better transition animation.

   Notice that Skeleton should only be used for **initial** loading. See https://ant.design/components/skeleton/#When-To-Use

2. Add memorization for HighlightSQL. Now the performance is acceptable even when we enable HighlightSQL for all DetailList rows:

   image

   The memorization is implemented using https://www.npmjs.com/package/moize. The difference between moize and React.memo is that, moize performs LRU cache for all instances, while React.memo only cache last version per instance.

3. Make whole table header sticky

   TODO: Currently there is still some lags caused by `Card` component having margins. Will be fixed in later PR.

   image

4. Add some memorization for functions to avoid unnecessary re-render.

5. When loading data in Statement / Slow Query list, the refresh button will be substituted by a spinner icon, to indicate data loading. This is also effective when user sorts the column:

   image

6. Resolve the `getKey` issue by using Skeleton again.

7. Introduced [why-did-you-render](https://github.com/welldone-software/why-did-you-render) in dev mode for easier tracking unnecessary re-render.

8. Add React.memo for some functional components that does not accept ReactNode (since ReactNode is not cacheable)


ui/lib/apps/SlowQuery/utils/useColumn.tsx
```diff
@@ -57,20 +56,21 @@ export function useSqlColumn(
...
+      ) : (
+        <Tooltip
+          title={<HighlightSQL sql={rec.query} theme="dark" />}
+          placement="right"
+        >
+          <TextWrap>
+            <HighlightSQL sql={rec.query} compact />
+          </TextWrap>
+        </Tooltip>
+      ),
```
baurine, line 73: Good point!

ui/lib/apps/Statement/components/StatementsTable.tsx
```diff
@@ -49,15 +50,17 @@ export default function StatementsTable({
...
       digest: rec.digest,
       schema: rec.schema_name,
       beginTime: timeRange.begin_time,
       endTime: timeRange.end_time,
     })
     navigate(`/statement/detail?${qs}`)
-  }
+  })
+
+  const getKey = useCallback((row) => `${row.digest}_${row.schema_name}`, [])
```
baurine, line 63: Just curious, what's the difference between usePersistFn and useCallback, seem useCallback also works for handleRowClick? (I guess the difference is caused by `navgiate`?)

breeswish: Same: Both return a function that will not be changed.

Difference:

- `a = useCallback(closure, dep)` is persisting the `closure` (and persist again when `dep` is changed), so that the variables captured in `closure` are not changed when `a` is called. This will cause problems if you reference something in the closure but not specified in `dep`. 

- `b = usePersistFn(closure)` is persisting another function that calls `closure` internally. This will not cause problems (reading stall data) when variables captured in `closure` is changed (since the closure is never persisted). `usePersistFn` is an alias to `useEventCallback` in the [React Hook FAQ](https://reactjs.org/docs/hooks-faq.html#how-to-read-an-often-changing-value-from-usecallback).

However as described by the React Hook FAQ, `usePersistFn` pattern is not recommended, so it is only used when `useCallback` is not suitable (i.e. changes too often).

PR #467, opened 2020-05-09
support open statement or slow query detail in a new tab or window

### LINKED ISSUE #374: command+click not working for clickable rows
## Bug Report

**What did you do?**

Command/Ctrl + Click row

Opened new window

**What did you see instead?**

Opened in current window

baurine:
close #374 

Screenshot:

I think it is fine to only support for statement and slow query table, if we need to support more tables, I will add them later.


ui/lib/utils/openLink.ts
```diff
@@ -0,0 +1,18 @@
...
+import React from 'react'
+
+export default function openLink(
+  url: string,
+  ev: React.MouseEvent<HTMLElement>,
+  navigate: NavigateFunction
+) {
+  if (ev.metaKey || ev.altKey || ev.ctrlKey) {
+    // open in a new tab
+    window.open(`/#${url}`, '_blank')
```
breeswish, line 11: This does not work in PD integration mode (the prefix is /dashboard)

baurine: Got it, let me think about it.

baurine:
address the comments.

PR #469, opened 2020-05-09
Refactor most remain JS code to TS code

baurine:
What did:

- Refactor most remain JS code to TS code, except the code inside dashboardApp, will do it later.
- Refactor remain class components to function components


ui/lib/utils/i18n.ts
```diff
@@ -1,11 +1,11 @@
-import i18next from 'i18next'
-import { initReactI18next } from 'react-i18next'
-import LanguageDetector from 'i18next-browser-languagedetector'
-
-import dayjs from 'dayjs'
 import 'dayjs/locale/en'
 import 'dayjs/locale/zh-cn'
 
+import dayjs from 'dayjs'
```
breeswish, line 4: do we need to import dayjs first?

baurine: not necessary, it is auto formatted by my vscode extension.

PR #474, opened 2020-05-11
refine format short value

### LINKED ISSUE #471: Execution Count displays incorrectly when the value is too large
## Bug Report

**What did you do?**

**What did you see instead?**

image

baurine:
close #471 

Screenshots:

image

image


baurine:
hi @breeswish , we discussed this problem several weeks ago, we said to hack the `getValueFormat()` method, but I think maybe it is better to wrap it instead of hacking it, how do you think about?

breeswish:
@baurine It is already recognized as a bug in grafana https://github.com/grafana/grafana/issues/23561 I think we can fix it by our own and not waiting for grafana to fix it, considering that the fix is extremely simple.

baurine:
Got, as it is recognized as a bug by grafana (I thought it is by design), it makes sense to fix it, and I think we should commit a PR for it (will do it later).

baurine:
I found this change changes the `scaledDecimals` original meaning (don't know what's its original meaning) totally and breaks too many upstream tests, so I release an incompatible version by changing version from 0.1.2 to 1.0.0 for grafana-value-formats lib.

PR #476, opened 2020-05-11
Refine instance host tables

baurine:
What did:

- Refine InstanceTable, add tooltip for version/deploy_path/git_hash columns, add dummyColumn
- Refine HostTable, add tooltip for disk column

Before:

企业微信20200511064550

After:

企业微信20200511062810

### DIFF
--- a/ui/lib/apps/ClusterInfo/components/InstanceTable.tsx
+++ b/ui/lib/apps/ClusterInfo/components/InstanceTable.tsx
@@ -13,6 +13,7 @@ import {
 import client from '@lib/client'
 import { CardTableV2 } from '@lib/components'
 import DateTime from '@lib/components/DateTime'
+import { dummyColumn } from '@lib/utils/tableColumns'
 import { useClientRequest } from '@lib/utils/useClientRequest'
 
 function useStatusColumnRender(handleHideTiDB) {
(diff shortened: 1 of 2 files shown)


ui/lib/apps/ClusterInfo/components/InstanceTable.tsx
```diff
@@ -153,21 +154,18 @@ export default function ListPage() {
...
-      onRender: (node) => (
-        <Tooltip title={`${node.ip}:${node.port}`}>
-          <span>
-            {node.ip}:{node.port}
-          </span>
-        </Tooltip>
-      ),
+      onRender: ({ ip, port }) => {
+        const fullName = `${ip}:${port}`
+        return <Tooltip title={fullName}>{fullName as any}</Tooltip>
```
breeswish, line 162: `<Tooltip>` needs some element node I believe, otherwise it won't take effect.

baurine: But actually it works, the tooltip can be displayed normally, just has type warning.

I just have a look at the Tooltip source code in https://github.com/ant-design/ant-design/blob/master/components/tooltip/index.tsx#L245 :

```js
    const child = getDisabledCompatibleChildren(
      React.isValidElement(children) ? children : <span>{children}</span>,
      prefixCls,
    );
```

So the Tooltip will wrap the string with `<span>`.

Anyway, it is better to pass the element node as children, I will update it soon.

breeswish: Good to know! I didn't know it before.

breeswish:
I have an idea. We can always append the dummy column in CardTableV2. (we can do it in another PR)

baurine:
> I have an idea. We can always append the dummy column in CardTableV2. (we can do it in another PR)

Right, brilliant!

PR #479, opened 2020-05-11
statement: fix refresh statements calls API twice

baurine:


ui/lib/apps/Statement/utils/useStatement.ts
```diff
@@ -125,7 +125,7 @@ export default function useStatement(
     }
 
     queryStatementList()
-  }, [queryOptions, allTimeRanges, validTimeRange, refreshTimes])
+  }, [queryOptions, allTimeRanges, validTimeRange])
```
baurine, line 128: the `refreshTimes` is already depended by another `useEffect()` which will update allTimeRanges and validTimaRange.

PR #481, opened 2020-05-12
profiling: support view and download a single profiling

baurine:
What did:

- Support view single profiling online
- Support download single profiling file
- Unify the get action token handler

Screenshots:


baurine:
> Cool! Actually I think we can simply remove the "single download" feature (since it is the same if we view + Ctrl+S). Then we can make "view" the default action for the row clicking.

Ok, let me do it.

baurine:
Final effect:

PR #482, opened 2020-05-12
Refine card table v2

baurine:
What did:

- Move isResizable / columnActionMode / dummyColumn down to CardTableV2


ui/lib/components/CardTableV2/index.tsx
```diff
@@ -132,10 +134,13 @@ function CardTableV2(props: ICardTableV2Props) {
     }
     newColumns = newColumns.map((c) => ({
       ...c,
+      isResizable: c.isResizable === false ? false : true,
       isSorted: c.key === orderBy,
       isSortedDescending: desc,
       onColumnClick,
+      columnActionsMode: c.columnActionsMode || ColumnActionsMode.disabled,
```
breeswish, line 141: disabled is set for non-sortable columns, and for sortable columns, the default is used.

baurine: Yes, because I found we need to write much `ColumnActionsMode.disabled` more than `ColumnActionsMode.clickable`

PR #484, opened 2020-05-13
Fix sticky table header can't scroll with its body in horizontal direction

### LINKED ISSUE #462: CardTableV2 sticky header can't scroll in horizontal direction
## Bug Report

**What did you do?**

**What did you see instead?**

51_origin

baurine:
fix #462 

What did:

- Fix the bug that sticky table header can't scroll with the body in the horizontal direction

The original situation:

51_origin

According to the issue: https://github.com/microsoft/fluentui/issues/7742 , the easiest way is to just set `constrainMode={ConstrainMode.unconstrained}` for DetailsList. This makes the sticky table header scrolls with its body but makes the sticky toolbar scrolls as well (same as the official demo), I don't think it is expected.

48_solution_1

So I only wrap the table instead of the root container by `<ScrollablePane>`.

49_solution_2

But the margin collapse doesn't work anymore, it makes the sticky table header jumping when it switches between sticky or not.

Adding another new props `noMarginTop` to remove the additional margin-top, maybe not very elegant, doesn't know whether it has some better ideas. (still a little jumping but much better than above)

Final effect:

52_solution_final

### DIFF
--- a/ui/lib/apps/SlowQuery/pages/List/index.tsx
+++ b/ui/lib/apps/SlowQuery/pages/List/index.tsx
@@ -16,7 +16,6 @@ import {
 import client from '@lib/client'
 import SlowQueriesTable from '../../components/SlowQueriesTable'
 import useSlowQuery from '../../utils/useSlowQuery'
-import { Sticky, StickyPositionType } from 'office-ui-fabric-react/lib/Sticky'
 
 const { Option } = Select
 const { Search } = Input
--- a/ui/lib/apps/SlowQuery/pages/List/index.tsx
+++ b/ui/lib/apps/SlowQuery/pages/List/index.tsx
@@ -67,98 +67,97 @@ function List() {
   }, [])
 
   return (
-    <ScrollablePane style={{ height: '100vh' }}>
-      <Sticky stickyPosition={StickyPositionType.Header} isScrollSynced>
-        <Card>
-          <Toolbar>
-            <Space>
-              <TimeRangeSelector
-                value={queryOptions.timeRange}
-                onChange={(timeRange) =>
-                  setQueryOptions({ ...queryOptions, timeRange })
-                }
-              />
-              <Select
-                value={queryOptions.schemas}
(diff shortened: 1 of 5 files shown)


ui/lib/apps/SlowQuery/pages/List/index.tsx
```diff
@@ -67,98 +66,97 @@ function List() {
...
-        slowQueries={slowQueries}
-        orderBy={orderOptions.orderBy}
-        desc={orderOptions.desc}
-        showFullSQL={showFullSQL}
-        visibleColumnKeys={visibleColumnKeys}
-        onGetColumns={setColumns}
-        onChangeOrder={changeOrder}
-      />
-    </ScrollablePane>
+      <div style={{ height: '100%', position: 'relative' }}>
```
breeswish, line 144: will this 100% lead to extra space in the vertical direction?

breeswish:
For the "This makes the sticky table header scrolls with its body but makes the sticky toolbar scrolls as well" issue, how about setting `isScrollSynced` to false (i.e. remove it)?

baurine:
> For the "This makes the sticky table header scrolls with its body but makes the sticky toolbar scrolls as well" issue, how about setting `isScrollSynced` to false (i.e. remove it)?

Tried but doesn't work.

ui/lib/apps/SlowQuery/pages/List/index.tsx
baurine: Tested but it does not. This div is required for `ScrollablePane`, it needs a relative position parent, and must set height or max-height for itself or its parent.

baurine: Let me have a test.

baurine: Yep, it will lead to some extra space on the page bottom if there are no much data, but I think it is acceptable currently.

image

Issue #487, opened 2020-05-13
UCP: Dashboard cannot run in Node 13+


breeswish:
## Description

Currently Dashboard cannot run in Node 13+

## Score

* 301

## Mentor(s)

* @baurine 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Nodejs

you06:
The score has been fixed.

Issue #485, opened 2020-05-13
Dashboard does not support Firefox


breeswish:
## Bug Report

![image](https://user-images.githubusercontent.com/1916485/81808845-ecbeb700-9552-11ea-9c68-493bdab759cc.png)

PR #489, opened 2020-05-13
fix build with node 14 && fix firefox

TennyZhuang:
UCP #485 
UCP #487 

upgrade react-scripts to fix build with node 14
fork [sql-formatter-plus](https://github.com/TennyZhuang/sql-formatter-plus/commit/d82c25f829d6767c6498d03984c8c20252457e26) to fix running with firefox


breeswish:
Further action item: https://github.com/kufii/sql-formatter-plus/pull/3

PR #500, opened 2020-05-15
instance_profiling: refine, remove Math.random() as key

baurine:
What did:

- A small refine, remove `Math.random()` as key for instance profiling pages


ui/lib/apps/InstanceProfiling/index.tsx
```diff
@@ -1,24 +1,26 @@
...
+  Routes,
+  useParams,
+} from 'react-router-dom'
 
 import { Root } from '@lib/components'
+import { Detail, List } from './pages'
 
-import DetailPage from './pages/Detail'
-import ListPage from './pages/List'
+function DetailPageWrapper() {
```
breeswish, line 12: Maybe this function can be a common pattern :) I believe for all pages using `useClientRequest` we'd better use a similar solution.

baurine: Make sense, let me try it.

ui/lib/components/ParamPageWrapper/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+export interface IParamPageWrapperProps {
+  paramName?: string
+  children: ReactNode
+}
+
+export default function ParamPageWrapper({
+  paramName = 'id',
+  children,
+}: IParamPageWrapperProps) {
+  const paramValue = useParams()[paramName]
```
breeswish, line 13: How about simply use the JSON stringify value of `useParams()` as the key? Then no need to specify the param name.

baurine: Good idea! Let me try it.

baurine: submit PR in #527

PR #505, opened 2020-05-18
ignore empty log searching results

### LINKED ISSUE #492: Do not produce slow log archive if there is no slow log
## Bug Report

image

Now we will have an empty slow log archive even if there is no slow log. This is pretty confusing.

Fullstop000:
UCP https://github.com/pingcap-incubator/tidb-dashboard/issues/507
Fix #492

Also, disable the tree node in log searching progress sidebar if the related task fails or has no searching results

### DIFF
--- a/pkg/apiserver/logsearch/task.go
+++ b/pkg/apiserver/logsearch/task.go
@@ -131,12 +131,12 @@ func (t *Task) setError(err error) {
 	t.model.Error = &errStr
 }
 
-func (t *Task) accumulateLogSize(dir *string) {
-	if dir != nil {
-		stat, err := os.Stat(*dir)
+func (t *Task) accumulateLogSize(path *string) {
+	if path != nil {
+		stat, err := os.Stat(*path)
 		if err != nil {
 			log.Warn("Can NOT fetch log file size for LogSearchTask",
-				zap.String("dir", *dir),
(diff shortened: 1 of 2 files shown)


ui/lib/apps/SearchLogs/components/SearchProgress.tsx
```diff
@@ -13,26 +13,16 @@ import styles from './Styles.module.css'
...
-      return {
-        icon: FailIcon,
-        disableCheckbox: true,
-      }
-    default:
-      break
+function leafNodeProps(task: LogsearchTaskModel) {
+  return {
+    icon: taskStateIcons[task.state || TaskState.Error],
+    disableCheckbox: task.size ? task.state !== TaskState.Finished : true,
```
breeswish, line 25: `disableCheckbox: !task.size || task.state !== TaskState.Finished`?

pkg/apiserver/logsearch/task.go
```diff
@@ -131,6 +131,21 @@ func (t *Task) setError(err error) {
 	t.model.Error = &errStr
 }
 
+func (t *Task) accumulateLogSize(dir *string) {
```
breeswish, line 134: maybe call it `path` is better.

PR #511, opened 2020-05-19
refine the use of backoff and fix tidb_forwarder bug

HunDunDM:
* *: refine the use of backoff with WithContext and Permanent
* tidb_forwarder: refine deduplication and sleep

### DIFF
--- a/pkg/tidb/proxy.go
+++ b/pkg/tidb/proxy.go
@@ -68,8 +68,8 @@ func (p *proxy) port() int {
 	return p.listener.Addr().(*net.TCPAddr).Port
 }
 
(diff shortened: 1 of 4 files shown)


pkg/tidb/proxy.go
```diff
@@ -189,27 +183,31 @@ func (p *proxy) doCheck(ctx context.Context) {
...
 			})
-		case <-ctx.Done():
-			return
 		}
 	}
 }
 
 func (p *proxy) run(ctx context.Context) {
-	endpoints := []string{}
+	var endpoints []string
```
breeswish, line 191: will this work for append?

Issue #508, opened 2020-05-19
UCP: Add gzip handler to static UI server


breeswish:
## Description

Now static UI server serves assets from go-bindata-assetfs. The assets served to the browser is not gzipped which will notably impact performance when used in Internet.

Goals:

- Add gzip to both uiserver in Dashboard and in PD.

## Score

* 605

## Mentor(s)

* @breeswish
* @HunDunDM 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Golang

PR #514, opened 2020-05-20
Add Gzip Support for uiserver

niedhui:
UCP #508

### DIFF
--- a/pkg/uiserver/uiserver.go
+++ b/pkg/uiserver/uiserver.go
@@ -14,8 +14,12 @@
 package uiserver
 
 import (
+	"compress/gzip"
 	"io"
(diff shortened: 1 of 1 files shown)


pkg/uiserver/uiserver.go
```diff
@@ -38,3 +42,52 @@ func Handler() http.Handler {
 func AssetFS() *assetfs.AssetFS {
 	return fs
 }
+
+type gzipHandler struct {
```
breeswish, line 46: Thanks! There is a similar server in PD as well. Could you figure out some way so that code can be reused? https://github.com/pingcap/pd/blob/master/pkg/dashboard/uiserver/uiserver.go

Issue #510, opened 2020-05-19
UCP: Avoid outputting raw log when generating profiles


breeswish:
## Description

When generating profiles, a raw log might be outputted like this:

```
Saved profile in /home/breezewish/pprof/pprof.tidb-server.samples.cpu.001.pb.gz
Generating report in dummy
Saved profile in /home/breezewish/pprof/pprof.tidb-server.samples.cpu.002.pb.gz
Generating report in dummy
```

This log does not conform our standard log format and is misleading. We need to remove them.

## Score

* 301

## Mentor(s)

* @rleungx 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Golang

PR #516, opened 2020-05-21
using a blank ui to eliminate the pprof logs

niedhui:
UCP #510


breeswish:
@all-contributors please add @niedhui   for code

Issue #515, opened 2020-05-21
UCP: Drop go-bindata totally


breeswish:
## Description

Currently we use go-bindata and go-bindata-assetfs to serve UI assets. There are some known limitations for go-bindata and we can switch to use vfsgen instead. For example, Prometheus pushgateway [switched](https://github.com/prometheus/alertmanager/pull/1532/files) from go-bindata to vfsgen as well. By using vfsgen, we can serve gzip contents more efficiently. Here are some detailed comparison: https://tech.townsourced.com/post/embedding-static-files-in-go/

There is a previous attempt of using vfsgen partially in TiDB Dashboard: https://github.com/pingcap-incubator/tidb-dashboard/pull/171 However now we don't have dynamic templates anymore so that the vfsgen introduced in that PR is not effective anymore.

## Score

* 605

## Mentor(s)

* @breeswish 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Golang

PR #504, opened 2020-05-17
Support dynamic prefix

TennyZhuang:
UCP #480

PR #521, opened 2020-05-23
replace go-bindata with vfsgen

niedhui:
UCP #515


pkg/apiserver/diagnose/diagnose.go
```diff
@@ -181,7 +181,13 @@ func (s *Service) reportHTMLHandler(c *gin.Context) {
 	}
 
 	// Serve report html directly from assets
-	d, err := s.uiAssetFS.Asset("build/diagnoseReport.html")
+	f, err := s.uiAssetFS.Open("diagnoseReport.html")
```
breeswish, line 184: maybe you can use [c.FileFromFS](https://github.com/gin-gonic/gin#serving-data-from-file)

scripts/embed_ui_assets.sh
```diff
@@ -27,15 +27,11 @@ if [ ! -d "ui/build" ]; then
...
-if [ "${NO_ASSET_BUILD_TAG:-}" = "1" ]; then
-  BUILD_TAG_PARAMETER=""
-else
-  BUILD_TAG_PARAMETER="-tags ${ASSET_BUILD_TAG:-ui_server}"
-fi
-
 echo "+ Embed UI assets"
 
-go-bindata-assetfs -pkg uiserver -prefix ui $BUILD_TAG_PARAMETER ui/build/...
+go build -o bin/assets_generate tools/assets_generate/main.go && bin/assets_generate
```
breeswish, line 32: How about using [go-generate](https://github.com/shurcooL/vfsgen#go-generate-usage)?

tools/assets_generate/main.go
```diff
@@ -0,0 +1,20 @@
...
+	"log"
+	"net/http"
+
+	"github.com/shurcooL/vfsgen"
+)
+
+func main() {
+	var fs http.FileSystem = http.Dir("ui/build")
+	err := vfsgen.Generate(fs, vfsgen.Options{
+		BuildTags:    "ui_server",
```
breeswish, line 13: There should be no build tags when it is built on CI to be served for PD (see github action file)

breeswish:
Also be aware of the dynamic modification introduced in https://github.com/pingcap-incubator/tidb-dashboard/pull/504/files

pkg/apiserver/diagnose/diagnose.go
niedhui: `c.FileFromFS` use `http.FileServer`, I just copy that code and replace it with `httpgzip.FileServer` 🤣 
BTW, why do we need to pass the assets file through to the service, not just use `uiserver#Handler` here?

tools/assets_generate/main.go
niedhui: Changed it to support passing build tags.

scripts/embed_ui_assets.sh
niedhui: Hi, I'm not quite familiar with  `go generate`, I'm wondering what's the benefit of using it here? 
Here is what I got after reading the doc:
1. Create a go file  including `//go:generate go run  assets_generate.go` comment
2. Export `BUILD_TAG_PARAMETER` in `embed_ui_assets.sh`
3. Fill `BuildTags` option of `vfsgen.Options` using `os.Getenv("BUILD_TAG_PARAMETER")`
4. Call `go generate` in `embed_ui_assets.sh`

Am I missing something?

niedhui:
> Also be aware of the dynamic modification introduced in https://github.com/pingcap-incubator/tidb-dashboard/pull/504/files

I'll adapt the 'InitAssetFS` after that PR merged

pkg/apiserver/diagnose/diagnose.go
breeswish: It it to serve the HTML file in a specific path ( so that `./data.js` will resolve to a correct location)

scripts/embed_ui_assets.sh
breeswish: Yes, looks good

pkg/apiserver/diagnose/diagnose.go
```diff
@@ -179,15 +179,12 @@ func (s *Service) reportHTMLHandler(c *gin.Context) {
...
-	// Serve report html directly from assets
-	d, err := s.uiAssetFS.Asset("build/diagnoseReport.html")
-	if err != nil {
-		c.Status(http.StatusNotFound)
-		return
-	}
-
-	c.Data(http.StatusOK, "text/html; charset=utf-8", d)
+	c.Request.URL.Path = "diagnoseReport.html"
+	httpgzip.FileServer(s.uiAssetFS, httpgzip.FileServerOptions{IndexHTML: true}).ServeHTTP(c.Writer, c.Request)
```
breeswish, line 187: can we access the content directly instead of constructing a FileServer each request?

pkg/uiserver/embedded_assets_rewriter.go
```diff
@@ -3,38 +3,41 @@
...
-		a.info = modifiedFileInfo{a.info, int64(len(a.bytes))}
-		_bindata[assetPath] = func() (*asset, error) {
-			return a, nil
+
+		var b bytes.Buffer
+		w := gzip.NewWriter(&b)
+		w.Write([]byte(updated))
+		w.Close()
+
+		fs := assets.(vfsgen۰FS)
```
breeswish, line 32: May be we can accept `asset` as an parameter so that this function can be reused later in PD.

pkg/uiserver/embedded_assets_rewriter.go
```diff
@@ -3,38 +3,41 @@
...
-
-func (f modifiedFileInfo) Size() int64 {
-	return f.size
-}
-
-func (f modifiedFileInfo) Sys() interface{} {
-	return nil
-}
-
 func InitAssetFS(prefix string) {
```
breeswish, line 13: let's rename it :) @TennyZhuang  is not good at giving names.

How about `RewriteAssetsPublicPath(publicPath string)`

pkg/uiserver/empty_assets_handler.go
```diff
@@ -15,11 +15,9 @@
 package uiserver
 
 import (
-	assetfs "github.com/elazarl/go-bindata-assetfs"
+	"net/http"
 )
 
+var assets http.FileSystem
```
breeswish, line 21: need a new line below

pkg/uiserver/embedded_assets_rewriter.go
```diff
@@ -3,38 +3,41 @@
 package uiserver
```
breeswish, line 3: could you also help supplying a license header? I missed this part previously..

pkg/apiserver/diagnose/diagnose.go
niedhui: Without the FileServer, we need to handle content-encoding by ourselves.

pkg/uiserver/embedded_assets_rewriter.go
niedhui: The  `vfsgen۰CompressedFileInfo`  is in the generated file, I extract most logic out, but remaining codes in `embedded_assets_rewriter.go` still need be copied to PD

pkg/uiserver/empty_assets_handler.go
niedhui: done

pkg/uiserver/embedded_assets_rewriter.go
niedhui: done

pkg/uiserver/embedded_assets_rewriter.go
niedhui: done

TennyZhuang: main package really don’t need to know what uiserver do with prefix, the InitAssetFs is more extensible, maybe we will use other implementation instead of rewrite.

I’m good at giving names, renaming here just “make reviewers happy”.

breeswish:
Hi, lint check didn't pass. Could you fix it?

niedhui:
backend lint fixed

breeswish:
LGTM, thanks! Could you also open the corresponding PD PR?

PR #587, opened 2020-06-01
ui: add progress bar when switching pages

### LINKED ISSUE #126: UCP: Show progress when switching frontend page
## Description

This task is to show progress by the progress bar when switch frontend pages.

## Score 

* 301

## Mentor(s)

* @baurine  

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* React

baurine:
resolve #126 

What did:

- Add progress bar in the top of page when switching pages

Just not sure whether this is a good way, because the [NProgess](https://github.com/rstacruz/nprogress) is a jQuery-style lib which operates the DOM directly.

And the progress bar color is not very explicit. If we want to change that color, we need to copy the nprogress.css from node_modules and edit it.

Demo:

### DIFF
--- a/ui/dashboardApp/layout/signin/index.js
+++ b/ui/dashboardApp/layout/signin/index.js
@@ -10,7 +10,7 @@ import {
 import { Form, Input, Button, message } from 'antd'
 import { motion } from 'framer-motion'
 import { useTranslation } from 'react-i18next'
-import LanguageDropdown from '@lib/components/LanguageDropdown'
+import { LanguageDropdown, TopLoadingBar } from '@lib/components'
 import client from '@lib/client'
 import * as auth from '@lib/utils/auth'
 
@@ -173,6 +173,7 @@ function App({ registry }) {
   return (
     <Root>
       <div className={styles.container}>
+        <TopLoadingBar />
         <div className={styles.dialogContainer}>
           <div className={styles.dialog}>
             <TiDBSignInForm registry={registry} />
--- a/ui/dashboardApp/layout/signin/index.js
+++ b/ui/dashboardApp/layout/signin/index.js
@@ -1,6 +1,7 @@
 import * as singleSpa from 'single-spa'
 import { Root } from '@lib/components'
 import React, { useState, useEffect, useRef } from 'react'
+import { HashRouter as Router } from 'react-router-dom'
 import {
   DownOutlined,
   GlobalOutlined,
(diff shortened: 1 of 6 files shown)


breeswish:
How about the solution react-top-loading-bar suggested in #126?

baurine:
> How about the solution react-top-loading-bar suggested in #126?

I glanced at its documents and example project - [react-moviews](https://github.com/klendi/react-movies), it seems need to change much code so I didn't try it. I can have a try today and see whether there are some other solutions.

baurine:
New solution by  react-top-loading-bar. But there is a small issue, the loading bar works for most pages except the sign-in page. I know the reason, because `single-spa:before-routing-event` event is emitted before this page is loaded. So I think it can't work really globally.

ui/dashboardApp/layout/signin/index.js
```diff
@@ -172,19 +173,22 @@ function TiDBSignInForm({ registry }) {
 function App({ registry }) {
   return (
     <Root>
-      <div className={styles.container}>
-        <div className={styles.dialogContainer}>
-          <div className={styles.dialog}>
-            <TiDBSignInForm registry={registry} />
+      <Router>
```
baurine, line 176: After testing, the TopLoadingBar can't receive the `single-spa:before-routing-event` event if it isn't wrapped by the Router component.

ui/dashboardApp/layout/signin/index.js
```diff
@@ -172,19 +173,22 @@ function TiDBSignInForm({ registry }) {
...
+        <TopLoadingBar />
```
breeswish, line 177: How about making it another SingleSPA app that always exists on the page, or simply put it outside everything mounted by SingleSPA. In this way, it can listen SingleSPA global events gracefully without worrying about its own lifecycle.

baurine: It may be a solution, let me have a try.

baurine: Implement it by this way in PR #661 , much more graceful! will close this PR.

baurine:
Implement it by PR #661 , so close this PR.

baurine:
Btw, this branch can be deleted.

PR #591, opened 2020-06-02
Refine ClusterInfo & Instance

breeswish:
1. Add `manualTestEnv` for easier testing. You can use Vagrant to establish the environment. See README in each directory for details.

2. Refined cluster info from getTopology to getTiDB, getStores, getPD, ...

3. Introduce `pd.Client` to backend, which gracefully handles context.

4. Share TiKV & TiFlash cluster info logic

5. Remove animations from sider menu to be efficient

6. Add a DebugPlayground App (TODO: Use Storybook instead @baurine )

7. Introduce `<BaseSelect>`, which is a `<Select>` like component with customization support for the drop content or displayed value

   image

8. Introduce `<InstanceSelect>` based on `<BaseSelect>` to provide a nice UI for selecting multiple instances

   image
   image
   image

It is compatible with Antd form validation:

   image

9. Now Overview is composed by self-managed components (except for Statement & Slow Query -- TODO: Refine them @baurine) instead of sharing some states. 

   This makes it possible to allow user customizing the Overview page in future.

10. Update Overview page UI:

   image

11. Introduce `utils/instanceTable` to contain everything common about instances (like instance names and variants)

12. Add more time slots to `<TimeRangeSelector>`

   image

13. Remove unused `<CardTable>` (TODO: Rename `<CardTableV2>` to `<CardTable>` @baurine)

14. Update `<CardTableV2>` to have a correct Checkbox display when there are group headers. (TODO: Drop implementation after https://github.com/microsoft/fluentui/issues/13144 is merged)

15. Slightly refine `<TimeRangeSelector>`

16. Refine LogSearch to use `<InstanceSelect>` and `utils/instanceTable` (and use `treeData` in `<Tree>` component)

17. Add `useCallback`, `useMemo`, `usePersistFn` to several places to reduce unnecessary re-render.

TODO:

- Test whether "Delete TiDB node" still works

Call for help:

- This PR has been proposed for a long time and I may made some mistakes (e.g. forget updating something). Please test / play with this PR as much as possible!

PR #602, opened 2020-06-04
ui: refine handle errors

baurine:
What did:

- Use key for the message box to avoid multiple same message boxes
- Only show the first error if has multiple errors
- Show "Network Error" in the Alert component as well
  
  企业微信20200604045627

  企业微信20200604045943

TODO:

- Move Alert out of CardTableV2 after `label` branch in PR #591 is merged. Currently moving out loses the margin.

  企业微信20200604045236

  企业微信20200604045251


baurine:
close this PR since it is replaced by PR #662, and this branch can be deleted.

PR #625, opened 2020-06-09
ui: remove deprecated <big> element

Rustin-Liu:
This feature is obsolete. Although it may still work in some browsers, its use is discouraged since it could be removed at any time. Try to avoid using it.


breeswish:
@all-contributors please add @Rustin-Liu for code

PR #607, opened 2020-06-08
integrate with dbass: support embed dasbhoard by iframe

baurine:
resolve #535 

What did:

- Support assign token/hideNav/lang params by URL

Demo:

Issue #535, opened 2020-05-26
Customization of dashboard when used as iframe


sivagao:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

use iframe' src url params to indicate : 

 - hide the side-menu(headless)
 - inject auth token to dashboard

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

baurine:
Maybe we can add a new entry for iframe usage likes `dashboardApp` and `diagnoseReportApp` which only includes statements and slow query features?

breeswish:
After discussion with @sivagao we have two solutions:
- Implement a portal page, which accept frontend configurations like apiPrefixPath, isHeadlessMode and so on. These configurations are then saved in localStorage.
- Add new configuration items in the backend and then passed to frontend. This is more complex but looks to be more secure.

breeswish:
See also https://github.com/pingcap-incubator/tidb-dashboard/issues/554

PR #628, opened 2020-06-10
DBass portal page

baurine:
Another implementation for issue #535 compared with PR #607 .

What did:

- Add a portal page for dbass to embed dashboard by iframe
- Move remain utils to @lib/utils
- Change all remain js files to ts files

For security consideration, I try to implement this feature in a new way, instead of attaching the token to the dashboard URL (in PR #607) which may access the dashboard without login dbass easily, now we pass the token to the dashboard by `postMessage()`.


ui/dbassPortal/index.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/////////////////////////////////////////////////
+
+let started = false
+auth.setStore(new auth.MemAuthTokenStore())
+window.addEventListener(
+  'message',
+  (event) => {
+    const appOptions = event.data
+    console.log('options:', appOptions)
```
baurine, line 72: To remove it later.

ui/lib/components/CardTableV2/index.tsx
```diff
@@ -18,8 +18,7 @@ import Card from '../Card'
 import styles from './index.module.less'
 
 export { AntCheckboxGroupHeader } from './GroupHeader'
-
-DetailsList.whyDidYouRender = {
+;(DetailsList as any).whyDidYouRender = {
```
baurine, line 21: fix it later

ui/dbassPortal/index.ts
```diff
@@ -0,0 +1,97 @@
+import '@lib/utils/wdyr'
```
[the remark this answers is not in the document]
baurine: Yep, agree with you about drafting a design before developing! And let me think about your solution.

ui/lib/utils/auth.ts
```diff
@@ -15,9 +19,49 @@ export function getAuthTokenAsBearer() {
...
+
+//////////////////////////////////////
+
+interface AuthTokenStore {
+  getAuthToken: () => string | null
+  setAuthToken: (token) => void
+  clearAuthToken: () => void
+}
+
+export class MemAuthTokenStore implements AuthTokenStore {
```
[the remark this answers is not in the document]
baurine: I tested it, it didn't lose.

ui/lib/components/CardTableV2/index.tsx
baurine: This is formatted by the prettier automatically.

baurine:
Hi, @breeswish , updated the implementation, and the main changes should be less than 50 LOC ^_^ except the auth.ts file.

I removed the real portal HTML file and replaced it by `/portal` virtual route.

Actually now this implementation is similar to the first try in PR #607 , but just change the way to pass parameters.

Please help take a look again when you're free, thanks in advance!

ui/lib/utils/auth.ts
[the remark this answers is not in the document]
breeswish: What if we open new pages?

ui/lib/components/CardTableV2/index.tsx
breeswish, line 21: Can we revert it now?

ui/dashboardApp/index.ts
```diff
@@ -23,8 +24,33 @@ import AppInstanceProfiling from '@lib/apps/InstanceProfiling/index.meta'
...
+async function main(appOptions: AppOptions = defOptions) {
+  // handle options
+  if (appOptions.isPortal) {
+    auth.setStore(new auth.MemAuthTokenStore())
+    auth.setAuthToken(appOptions.token)
+  } else {
+    auth.setStore(new auth.LocalStorageAuthTokenStore())
+  }
+  if (appOptions.lang) {
+    i18n.changeLang(appOptions.lang)
```
breeswish, line 50: We can use `i18next.changeLanguage` to be aligned with the rest of the code

ui/dashboardApp/index.ts
```diff
@@ -23,8 +24,33 @@ import AppInstanceProfiling from '@lib/apps/InstanceProfiling/index.meta'
...
+  lang: '',
+  hideNav: false,
+}
+
+async function main(appOptions: AppOptions = defOptions) {
+  // handle options
+  if (appOptions.isPortal) {
+    auth.setStore(new auth.MemAuthTokenStore())
+    auth.setAuthToken(appOptions.token)
+  } else {
```
breeswish, line 46: Maybe no need for these 4 lines

ui/dashboardApp/index.ts
```diff
@@ -23,8 +24,33 @@ import AppInstanceProfiling from '@lib/apps/InstanceProfiling/index.meta'
...
+  token?: string
+  lang?: string
+  hideNav?: boolean
+}
+
+const defOptions: AppOptions = {
+  isPortal: false,
+  token: '',
+  lang: '',
+  hideNav: false,
```
breeswish, line 38: We can treat `lang`, `hideNav` as a config, then store them in localStorage. When App is initialized, we read them from local storage. In this way, "Command+Click" will keep working.

ui/dashboardApp/index.ts
```diff
@@ -34,7 +60,10 @@ async function main() {
 
   singleSpa.registerApplication(
     'layout',
-    AppRegistry.newReactSpaApp(() => LayoutMain, 'root'),
+    AppRegistry.newReactSpaApp(
+      () => (appOptions.hideNav ? LayoutFull : LayoutMain),
+      'root'
```
breeswish, line 65: We can pass the config to each App (modify AppRegistry). Then we even don't need to introduce a new Layout.

baurine:
Address all comments, @breeswish PTAL, thanks!

ui/dashboardApp/index.ts
```diff
@@ -75,12 +84,60 @@ async function main() {
...
   singleSpa.start()
 }
 
-main()
+/////////////////////////////////////
+
+// check whether this is a portal page
+function checkPortal(): boolean {
+  return window.location.pathname.endsWith('/portal')
+}
```
breeswish, line 99: I prefer to put them in `routing` and use something like `routing.isLocationMatch`.

ui/dashboardApp/index.ts
```diff
@@ -75,12 +84,60 @@ async function main() {
...
+  }
+  let q = hash.substring(pos + 1)
+  const p = new URLSearchParams(q)
+  return p.get('iframe') !== null
+}
+
+function start() {
+  // non portal page
+  if (!checkPortal()) {
+    main({ hideNav: false, lang: '' })
```
breeswish, line 116: How about storing it in something like `defaultAppOptions`

ui/dashboardApp/index.ts
```diff
@@ -75,12 +84,60 @@ async function main() {
...
+    return
+  }
+
+  // use another auth token key for portal page to avoid affect the non portal page
+  auth.setTokenKey(auth.portalTokenKey)
+
+  // portal page runs in iframe
+  if (checkIframe()) {
```
breeswish, line 124: We can make life simple by stipulating that the portal page always accepts config (or reset config) from `message` event and the other pages always loads config from localStorage (or if non-exist, using default config). In this way, we don't need to check the existence of `iframe` as well as supporting multiple token keys.

ui/dashboardApp/layout/signin/index.tsx
```diff
@@ -33,8 +33,8 @@ function TiDBSignInForm({ registry }) {
   const [loading, setLoading] = useState(false)
   const [signInError, setSignInError] = useState(null)
 
-  const refForm = useRef()
-  const refPassword = useRef()
+  const [refForm] = Form.useForm()
```
breeswish, line 36: 👍

ui/dashboardApp/index.ts
baurine: Yep! will update it soon.

ui/dashboardApp/index.ts
baurine: Understand the whole logic finally and update it!

ui/dashboardApp/index.ts
```diff
@@ -75,12 +84,35 @@ async function main() {
...
+  // the portal page is only used to receive options
+  if (routing.isPortalPage()) {
+    function handleConfigEvent(event) {
+      const { token, lang, hideNav, app } = event.data
+      auth.setAuthToken(token)
+      saveAppOptions({ hideNav, lang })
+
+      // redirect
+      const { origin, pathname } = window.location
+      const homePath = pathname.replace('portal', '#/')
```
breeswish, line 106: user may specify `portal` in publicUrlPrefix

ui/lib/utils/routing.ts
```diff
@@ -10,6 +10,10 @@ export function isLocationMatch(s, matchPrefix) {
   }
 }
 
-export function isLocationMatchPrefix(s) {
+export function isLocationMatchPrefix(s): boolean {
   return isLocationMatch(s, true)
 }
+
+export function isPortalPage(): boolean {
+  return window.location.pathname.endsWith('/portal')
```
breeswish, line 18: I realized a problem. Since we support customizing publicPath, `/portal` is a valid publicPath that may be specified by user. Maybe we must put it after `#`.

baurine: Yes, it is, put it in the hash is better. Done!

ui/dashboardApp/index.ts
```diff
@@ -72,15 +81,38 @@ async function main() {
...
+  if (routing.isPortalPage()) {
+    function handleConfigEvent(event) {
+      const { token, lang, hideNav, app } = event.data
+      auth.setAuthToken(token)
+      saveAppOptions({ hideNav, lang })
+
+      // redirect
+      const { origin, pathname } = window.location
+      const appUrl = `${origin}${pathname}#/${app || 'statement'}`
+      window.location.href = appUrl
```
breeswish, line 107: how about `window.location.hash = `#${redirectPath}` directly

ui/dashboardApp/index.ts
```diff
@@ -72,15 +81,38 @@ async function main() {
...
 }
 
-main()
+/////////////////////////////////////
+
+function start() {
+  // the portal page is only used to receive options
+  if (routing.isPortalPage()) {
+    function handleConfigEvent(event) {
+      const { token, lang, hideNav, app } = event.data
```
breeswish, line 100: prefer to name it `redirectPath` so that it can be something complex (e.g. including the query parameter part).

ui/dashboardApp/index.ts
baurine: Don't know this way before, not sure whether it works, let me try it.

baurine: Awesome! works perfectly.

.github/workflows/master-release-ui-assets.yaml
```diff
@@ -3,7 +3,7 @@ name: Release UI Assets
 on:
   push:
     branches:
-      - master
+      - dbass-portal
```
breeswish, line 6: Cool, please revert this line and the rest LGTM

PR #634, opened 2020-06-15
check browers list

baurine:
What did: 

- Show a simple prompt message, which suggests user update browser, in the top of the page if the browser is not compatible with the browserslist

Principle: [browserslist-useragent-regexp](https://github.com/browserslist/browserslist-useragent-regexp) is a utility to compile browserslist query to a RegExp to test browser useragent.


ui/public/checkBrowser.js
```diff
@@ -0,0 +1,35 @@
+function checkBrowser() {
+  if (!window.__supported_browsers__.test(navigator.userAgent)) {
+    var text
+    if (navigator.language.indexOf('zh') === 0) {
+      text =
+        '您的浏览器版本已过期，推荐使用最新版本的 Chrome 或 Edge 浏览器以便获得更好的体验。'
+    } else {
+      text =
+        'Your browser version is outdated, recommend to use the latest Chrome or Edge to get better experience.'
+    }
```
baurine, line 10: Do a simple i18n to avoid to use i18next library.

ui/public/diagnoseReport.html
```diff
@@ -10,6 +10,8 @@
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
+    <script src="%PUBLIC_URL%/supportedBrowsers.js"></script>
+    <script src="%PUBLIC_URL%/checkBrowser.js"></script>
```
baurine, line 14: We can combine these 2 js together but that needs to use sed. As I know, the sed works differently between macOS and Linux.

ui/public/checkBrowser.js
```diff
@@ -0,0 +1,35 @@
...
+        <b>
+          ${text}
+          <a><span>X</span></a>
+        </b>
+      <div>
+    `
+    var d = document.createElement('div')
+    d.innerHTML = content
+    d.querySelector('a').onclick = function () {
+      d.querySelector('div').style.top = '-60px'
```
baurine, line 29: TODO: use old style `getElementByTag()` API to be as compatible as much.

ui/public/checkBrowser.js
```diff
@@ -0,0 +1,35 @@
...
+                  z-index: 4;
+                  padding: 8px;
+                  text-align: center;
+                  transition: top 2s;">
+        <b>
+          ${text}
+          <a><span>X</span></a>
+        </b>
+      <div>
+    `
```
baurine, line 25: TODO: not use string interpolator.

ui/public/supportedBrowsers.js
```diff
@@ -0,0 +1 @@
...1-9]\d|\d{3,})[_\.]\d+|11[_\.]3|11[_\.]([4-9]|\d{2,})|(1[2-9]|[2-9]\d|\d{3,})[_\.]\d+|12[_\.]0|12[_\.]([1-9]|\d{2,})|12[_\.]4|12[_\.]([5-9]|\d{2,})|(1[3-9]|[2-9]\d|\d{3,})[_\.]\d+|13[_\.]0|13[_\.]([1-9]|\d{2,})|13[_\.]3|13[_\.]([4-9]|\d{2,})|(1[4-9]|[2-9]\d|\d{3,})[_\.]\d+)(?:[_\.]\d+)?)|(CFNetwork\/8.* Darwin\/16\.5\.\d+)|(CFNetwork\/8.* Darwin\/16\.6\.\d+)|(CFNetwork\/8.* Darwin\/16\.7\.\d+)|(SamsungBrowser\/(10\.1|10\.([2-9]|\d{2,})|(1[1-9]|[2-9]\d|\d{3,})\.\d+))|(Edge\/(18(?:\.0)?|18(?:\.([1-9]|\d{2,}))?|(19|[2-9]\d|\d{3,})(?:\.\d+)?))|((Chromium|Chrome)\/(49\.0|49\.([1-9]|\d{2,})|([5-9]\d|\d{3,})\.\d+|79\.0|79\.([1-9]|\d{2,})|([8-9]\d|\d{3,})\.\d+)(?:\.\d+)?([\d.]+$|.*Safari\/(?![\d.]+ Edge\/[\d.]+$)))|(Version\/(12\.1|12\.([2-9]|\d{2,})|(1[3-9]|[2-9]\d|\d{3,})\.\d+|13\.0|13\.([1-9]|\d{2,})|(1[4-9]|[2-9]\d|\d{3,})\.\d+)(?:\.\d+)? Safari\/)|(Trident\/7\.0)|(Firefox\/(68\.0|68\.([1-9]|\d{2,})|(69|[7-9]\d|\d{3,})\.\d+|74\.0|74\.([1-9]|\d{2,})|(7[5-9]|[8-9]\d|\d{3,})\.\d+)\.\d+)|(Firefox\/(68\.0|68\.([1-9]|\d{2,})|(69|[7-9]\d|\d{3,})\.\d+|74\.0|74\.([1-9]|\d{2,})|(7[5-9]|[8-9]\d|\d{3,})\.\d+)(pre|[ab]\d+[a-z]*)?)|(([MS]?IE) (11\.0|11\.([1-9]|\d{2,})|(1[2-9]|[2-9]\d|\d{3,})\.\d+))/
```
breeswish, line 1: maybe we should ignore this file?

ui/public/checkBrowser.js
```diff
@@ -0,0 +1,27 @@
+function checkBrowser() {
+  if (!window.__supported_browsers__.test(navigator.userAgent)) {
+    var text
+    if (navigator.language.indexOf('zh') === 0) {
+      text =
+        '您的浏览器版本已过期，推荐使用最新版本的 Chrome 或 Edge 浏览器以便获得更好的体验。'
+    } else {
+      text =
+        'Your browser version is outdated, recommend to use the latest Chrome or Edge to get better experience.'
```
breeswish, line 9: Sometimes the page is simply broken, prefer to avoid the soft "recommend"

breeswish: We also support Firefox

ui/public/checkBrowser.js
```diff
@@ -0,0 +1,27 @@
+function checkBrowser() {
+  if (!window.__supported_browsers__.test(navigator.userAgent)) {
+    var text
+    if (navigator.language.indexOf('zh') === 0) {
```
breeswish, line 4: IE6 ~ 10 even do not support NavigatorLanguage API, and I'm afraid it will be broken in such cases.

ui/public/checkBrowser.js
```diff
@@ -0,0 +1,27 @@
...
+      text =
+        '您的浏览器版本已过期，推荐使用最新版本的 Chrome 或 Edge 浏览器以便获得更好的体验。'
+    } else {
+      text =
+        'Your browser version is outdated, recommend to use the latest Chrome or Edge to get better experience.'
+    }
+
+    const content =
+      '<div style="background: yellow; width: 100%; position: absolute; top: 0; z-index: 4; padding: 8px; text-align: center; transition: top 2s;">' +
```
breeswish, line 13: Transition needs later browser support and may not work in old browsers (please verify whether it will break things)

ui/public/checkBrowser.js
```diff
@@ -0,0 +1,27 @@
+function checkBrowser() {
+  if (!window.__supported_browsers__.test(navigator.userAgent)) {
+    var text
+    if (navigator.language.indexOf('zh') === 0) {
+      text =
+        '您的浏览器版本已过期，推荐使用最新版本的 Chrome 或 Edge 浏览器以便获得更好的体验。'
```
breeswish, line 6: UTF8 characters might cause problems in some old browsers I guess, please test it against IE6

ui/public/diagnoseReport.html
```diff
@@ -10,6 +10,8 @@
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
+    <script src="%PUBLIC_URL%/supportedBrowsers.js"></script>
```
breeswish, line 13: Can we ensure that these scripts will run before our main scripts?

ui/public/checkBrowser.js
breeswish, line 13: How about using red as the background to bring more attractions

ui/public/supportedBrowsers.js
baurine: Agree.

ui/public/checkBrowser.js
baurine: Ok, how about "latest modern browsers"?

ui/public/checkBrowser.js
baurine: Let me do some test.

ui/public/checkBrowser.js
breeswish: Modern browser is a word for developers and users don't know what it means precisely. I prefer to explicitly list out our supported browsers

ui/public/checkBrowser.js
baurine: So it just has no animation in the old browser, I guess, let me test it.

ui/public/checkBrowser.js
baurine: Really, just know it, 😂 Where I can find an IE6 now... anyway, I will try it, maybe I can test it in IE11 first. (ant design only support IE11 for IEs)

ui/public/diagnoseReport.html
baurine: Yes, according to the result I observed, I will do some more tests to confirm.

ui/public/checkBrowser.js
baurine: I am ok with it, just found some examples which use the yellow background, so do I.

image

image

ui/public/checkBrowser.js
breeswish: A note.. AFAIK TiDB Dashboard won't work on IE11 as well

ui/public/checkBrowser.js
breeswish: Cool, if it doesn't break things then it would be great

ui/public/checkBrowser.js
breeswish: IE6 has a wrong charset detecting strategy https://support.microsoft.com/zh-cn/help/928847/internet-explorer-uses-the-wrong-character-set-when-it-renders-an-html

ui/public/checkBrowser.js
breeswish: How about two policies:

- For browsers that functionality will totally break, must not continue, asking user to change browser, like:
  image
- For browsers that functionality works (but maybe not perfect), display a soft recommendation.

baurine: Compared following colors (choose from ant design colors):

red:
red

ff4d4f:
#ff4d4f

yellow:
yellow

ffc53d:
#ffc53d

fadb14:
#fadb14

Feel the last one makes me most comfortable.

baurine: > How about two policies:
> 
> * For browsers that functionality will totally break, must not continue, asking user to change browser, like:
>   image
> * For browsers that functionality works (but maybe not perfect), display a soft recommendation.

This would be better.

baurine:
Updated logic:

1. If the browser is not supported by our pages definitely, we show the following static page to ask the user to update their browser. The list of the unsupported browsers is here: https://browserl.ist/?q=dead%2C+IE+11

    企业微信20200616042540

    企业微信20200616042704

1. If the browser is supported but the version is too old, show a prompt message box to ask the user to upgrade their browser to the latest version as demonstrated above.

ui/gulpfile.esm.js
```diff
@@ -28,11 +28,40 @@ task(
...
+    'echo "window.__supported_browsers__ = $(browserslist-useragent-regexp --allowHigherVersions)" > ./public/supportedBrowsers.js'
+  )
+)
+
+task(
+  'unSupportedBrowsers',
+  shell.task(
+    'echo "window.__unsupported_browsers__ = $(browserslist-useragent-regexp "dead, IE 11" --allowHigherVersions)" > ./public/unSupportedBrowsers.js'
+  )
+)
```
baurine, line 43: The generated full list is here: https://browserl.ist/?q=dead%2C+IE+11

breeswish: It won't work in some old Chrome versions as well. The list needs some tuning..

ui/public/updateBrowser.html
```diff
@@ -0,0 +1,101 @@
...
+      .safari {
+        background-image: url(./assets/img/safari.png);
+      }
+    </style>
+  </head>
+  <body>
+    <div id="root">
+      <h1 id="title">Please update your browser</h1>
+      <p id="desc">
+        Your browser isn't supported anymore. Update it to get the best TiDB
```
breeswish, line 56: Notice that YouTube will display different messages when current browser is up-to-date while browsing the same page:
image

This avoid confusing when user switched the browser, and copy-paste the URL.

baurine:
I am working on another PR.

Issue #181, opened 2020-03-06
UCP: Replace momentjs with dayjs


breeswish:
## Description

Momentjs has some major problems like mutable API and heavy bundle size. We can switch to dayjs, which is also supported by ant design.

https://ant.design/docs/react/replace-moment
https://github.com/iamkun/dayjs

## Score

* 300

## Mentor(s)

* @baurine 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* React

gauss1314:
Hi, I find `Replace momentjs with dayjs` require antd 4.0, but we use antd 3.x.

Because when use `import generatePicker from 'antd/es/date-picker/generatePicker';`, `generatePicker` is added since 4.0.

breeswish:
@gauss1314 antd 3.x also supports dayjs and actually it is already using dayjs (see `config-overrides.js`, not verified).

breeswish:
I'm updating the score to 300 now, since recently we removed the [antd-dayjs-webpack-plugin](https://github.com/ant-design/antd-dayjs-webpack-plugin) due to its compatibility issue with existing dayjs instances. One need to add it back and resolve these compatibility issues to work on this PR.

Rustin-Liu:
I am working on this.

PR #637, opened 2020-06-15
ui: replace momentjs with dayjs

Rustin-Liu:
closes https://github.com/pingcap-incubator/tidb-dashboard/issues/181

### DIFF
--- a/ui/lib/apps/SearchLogs/components/SearchResult.tsx
+++ b/ui/lib/apps/SearchLogs/components/SearchResult.tsx
@@ -5,12 +5,12 @@ import { Alert } from 'antd'
 import React, { useEffect, useState, useMemo, useCallback } from 'react'
 import { useTranslation } from 'react-i18next'
 import { InstanceKindName } from '@lib/utils/instanceTable'
+import dayjs from 'dayjs'
 
 import { LogLevelText } from '../utils'
 import Log from './Log'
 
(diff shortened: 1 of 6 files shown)


ui/lib/components/DatePicker/index.tsx
```diff
@@ -0,0 +1,8 @@
+import { Dayjs } from 'dayjs'
```
breeswish, line 1: I think you also need to expose it in `components/index.ts`.

ui/lib/apps/SearchLogs/components/SearchResult.tsx
```diff
@@ -11,6 +10,7 @@ import { LogLevelText } from '../utils'
 import Log from './Log'
 
 import styles from './Styles.module.less'
+import dayjs from 'dayjs'
```
breeswish, line 13: prefer to put it before L9

ui/lib/components/DatePicker/index.tsx
Rustin-Liu: Added.

ui/lib/apps/SearchLogs/components/SearchResult.tsx
Rustin-Liu: Fixed.

PR #639, opened 2020-06-15
ui: add engine requirements and simplify format command

Rustin-Liu:


Rustin-Liu:
> Maybe you also need to change CI?

Fixed. Maybe a better way is add two commands:

"fmt": "prettier --check ."

"fmt:fix": "prettier --write ."


But it's  OK for me. What do you think?

breeswish:
@Rustin-Liu Oh, I have already merged it 🤣 Considering that `check` is only used in CI and no one else is likely to use it, I think we can keep the current name.

PR #644, opened 2020-06-17
Report data by mixpanel

baurine:


pkg/apiserver/info/info.go
```diff
@@ -45,8 +45,9 @@ func Register(r *gin.RouterGroup, auth *user.AuthService, s *Service) {
 }
 
 type InfoResponse struct { //nolint:golint
-	Version    pkgutils.VersionInfo `json:"version"`
-	PDEndPoint string               `json:"pd_end_point"`
+	Version      pkgutils.VersionInfo `json:"version"`
+	PDEndPoint   string               `json:"pd_end_point"`
+	EnableReport bool                 `json:"enable_report"`
```
breeswish, line 50: As a default opt-in option, prefer to call it `DisableXxx`, like `DisableTelemetry`.

pkg/config/config.go
```diff
@@ -38,6 +38,9 @@ type Config struct {
 
 	// TLS config for mTLS authentication between TiDB and MySQL client.
 	TiDBTLSConfig *tls.Config
+
+	// Enable client to report data for analysis
+	EnableReport bool
```
breeswish, line 43: `DisableTelemetry`

ui/dashboardApp/index.ts
```diff
@@ -38,6 +39,7 @@ async function main(options: AppOptions) {
   )
 
   apiClient.init()
+  reportData.init()
```
breeswish, line 42: `telemetry.init`

ui/dashboardApp/index.ts
```diff
@@ -88,6 +90,12 @@ async function main(options: AppOptions) {
     }
   })
 
+  window.addEventListener('single-spa:before-routing-event', () => {})
+
+  window.addEventListener('single-spa:routing-event', () => {
+    reportData.report('EventPageChange', {})
```
breeswish, line 96: `telemetry.track('PageChange', ...)`

ui/dashboardApp/index.ts
breeswish: Need to await?

ui/lib/utils/reportData.ts
```diff
@@ -0,0 +1,17 @@
+import mixpanel from 'mixpanel-browser'
+import client from '@lib/client'
+
+export async function init() {
+  mixpanel.init('cb1135f29a413c653332990f07b3586a', {
+    opt_out_tracking_by_default: true,
```
breeswish, line 6: we can enable `batch_requests` and `persistence = localStorage`.

ui/lib/utils/reportData.ts
```diff
@@ -0,0 +1,17 @@
...
+  })
+  // check option
+  const res = await client.getInstance().getInfo()
+  if (res?.data?.enable_report) {
+    mixpanel.opt_in_tracking()
+  }
+}
```
breeswish, line 13: **In next PR**, we need to further identify users by using https://developer.mixpanel.com/docs/javascript#managing-user-identity

Notice that the identity in our case should comes from unique cluster uuid + the hash of user id (since different clusters will have the same user, e.g. root, but they are different users)

ui/lib/utils/reportData.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function report(eventType: string, eventBody: object) {
+  mixpanel.track(eventType, eventBody)
+}
```
breeswish, line 17: I would prefer to not encapsulate the API so that we are free to use all Mixpanel APIs.

ui/lib/utils/reportData.ts
```diff
@@ -0,0 +1,17 @@
+import mixpanel from 'mixpanel-browser'
+import client from '@lib/client'
+
+export async function init() {
+  mixpanel.init('cb1135f29a413c653332990f07b3586a', {
```
breeswish, line 5: I opened a new project called `Dev` so that we can separate Dev data and production data. Then you can supply different token in different env by using `.env` file: https://create-react-app.dev/docs/adding-custom-environment-variables/

ui/dashboardApp/index.ts
baurine: Yep, you're right, else we need to return a "./config.js" in the html head.

ui/lib/utils/reportData.ts
baurine: So we will always use MixPanel, won't change it later?

ui/lib/utils/reportData.ts
baurine: Nice!

pkg/apiserver/info/info.go
```diff
@@ -38,15 +38,16 @@ func NewService(config *config.Config, tidbForwarder *tidb.Forwarder, db *dbstor
 
 func Register(r *gin.RouterGroup, auth *user.AuthService, s *Service) {
 	endpoint := r.Group("/info")
-	endpoint.Use(auth.MWAuthRequired())
 	endpoint.GET("/info", s.infoHandler)
```
baurine, line 41: `/info` API should not be auth required, else it has no normal response before login. And I think it is fine to non-auth-required.

ui/lib/utils/reportData.ts
breeswish: Nop, we definitely would change. But I think current abstraction is not enough for a painless vendor change.

breeswish:
Maybe we can use `[mixpanel.register](https://developer.mixpanel.com/docs/javascript-full-api-reference#mixpanelregister)` to override (whether it will indeed override needs further test) [`$current_url`](https://github.com/mixpanel/mixpanel-js/blob/bf905cfcda6fa1f8c29edfb857d7e28932a4fd4d/src/utils.js#L1617) when route is changed. Also I believe `[referrer](https://github.com/mixpanel/mixpanel-js/blob/bf905cfcda6fa1f8c29edfb857d7e28932a4fd4d/src/utils.js#L1613)` shall be removed for which could contain sensitive data.

ui/lib/utils/telemetry.ts
```diff
@@ -8,15 +8,36 @@ export async function init() {
...
   const res = await client.getInstance().getInfo()
   if (res?.data?.disable_telemetry === false) {
-    mixpanel.opt_in_tracking()
     // https://developer.mixpanel.com/docs/javascript-full-api-reference#mixpanelset_config
     mixpanel.set_config({
       batch_requests: true,
       persistence: 'localStorage',
     })
+    mixpanel.register({
+      // $initial_referrer and $initial_referring_domain overrides don't work in register()
```
breeswish: initial_referrer is registered using register_once, so that it cannot be overridden. You can use property blacklist: https://help.mixpanel.com/hc/en-us/articles/115004601523-Block-Default-Mixpanel-Properties

ui/lib/utils/telemetry.ts
```diff
@@ -8,15 +8,36 @@ export async function init() {
...
+  if (pos === -1) {
+    return fullUrlHash
+  }
+  return fullUrlHash.substring(0, pos)
+}
+
+export function track(eventType: string, eventBody: object = {}) {
+  mixpanel.track(eventType, {
+    ...eventBody,
+    $current_url: getPathInUrlHash(),
```
breeswish, line 40: I prefer to override `current_url` in route events, otherwise other mixpanel features eithere cannot be used (due to incorrect current_url) or need similar pattern again.

baurine: You mean, use mixpanel.register() before mixpanel.track(), right? Let me try it.

baurine:
Final effects:

企业微信20200618112115

企业微信20200618112057

baurine:
Found the test job in the CI reports the data to the production environment, try to resolve it.

.github/workflows/e2e-test.yaml
```diff
@@ -54,6 +54,7 @@ jobs:
         env:
           NO_MINIMIZE: true
           CI: true
+          REACT_APP_MIXPANEL_TOKEN: 6b8d3fb93ffd04f2e76325540c838624 # change mixpanel token to dev env
```
breeswish, line 57: This leaves two places for a single token, which is not a good practice considering maintenance. I think we can assign an empty token for it.

pkg/apiserver/info/info.go
```diff
@@ -38,15 +38,16 @@ func NewService(config *config.Config, tidbForwarder *tidb.Forwarder, db *dbstor
...
+	endpoint.Use(auth.MWAuthRequired())
 	endpoint.GET("/whoami", s.whoamiHandler)
 	endpoint.GET("/databases", utils.MWConnectTiDB(s.tidbForwarder), s.databasesHandler)
```
breeswish, line 44: For the /databases endpoint, it requires `MWAuthRequired`. However I guess this API is not used anywhere. You can verify it (and remove it)

baurine: Refine it by PR #646 , use this API to replace the `/statements/schemas` API.

PR #647, opened 2020-06-18
update telemetry endpoint

baurine:
Update the telemetry endpoint.


baurine:
@breeswish @HunDunDM PTAL, thanks!

ui/.env.development
```diff
@@ -1 +1,2 @@
 REACT_APP_MIXPANEL_TOKEN=6b8d3fb93ffd04f2e76325540c838624
+REACT_APP_MIXPANEL_HOST=http://127.0.0.1:8090/api/v1/dashboard/report
```
breeswish, line 2: Is it better to not provide a local endpoint in development? Otherwise developers need to set up a local server in order to see what kind of event is collected, instead of just using the Mixpanel UI.

baurine: Make sense, but this doesn't affect the 4.0.2 release. I will update it after finishing the telemetry-log-collector, and when developing, I can put it in the local .env file.

breeswish: Yup. I think so.

PR #648, opened 2020-06-18
ui: refine some code

Rustin-Liu:


ui/lib/apps/KeyViz/components/KeyViz.less
```diff
@@ -141,8 +141,7 @@
 
   .heatmap {
     flex-grow: 1;
-    margin: @padding-page;
-    margin-top: -32px;
+    margin: -32px @padding-page @padding-page;
```
baurine, line 144: In my personal opinion, I prefer the original expression, 3 values expression makes me confused.

ui/lib/components/Descriptions/index.module.less
```diff
@@ -7,13 +7,11 @@
     }
 
     .ant-descriptions-row > th {
-      padding: 0;
-      padding-bottom: @padding-xs;
+      padding: 0 0 @padding-xs;
```
baurine, line 10: ditto

ui/lib/components/Descriptions/index.module.less
```diff
@@ -7,13 +7,11 @@
...
     }
 
     .ant-descriptions-row > td {
-      padding: 0;
-      padding-bottom: @padding-md;
+      padding: 0 0 @padding-md;
```
baurine, line 14: ditto

ui/lib/apps/KeyViz/components/KeyViz.less
Rustin-Liu: I don't think so, but it's OK for me. I will revert it.

Rustin-Liu:
@baurine  Reverted, PTAL~  Thanks!

ui/lib/apps/SearchLogs/components/SearchHeader.tsx
```diff
@@ -119,18 +119,14 @@ export default function SearchHeader({ taskGroupID }: Props) {
...
-      try {
-        const result = await client.getInstance().logsTaskgroupPut(req)
-        const id = result.data.task_group?.id
-        if (!id) {
-          throw new Error('Invalid server response')
-        }
-        navigate(`/search_logs/detail/${id}`)
-      } catch (e) {
-        // FIXME
+      const result = await client.getInstance().logsTaskgroupPut(req)
```
breeswish, line 122: What about errors?

ui/lib/components/CardTableV2/index.tsx
```diff
@@ -168,7 +168,7 @@ function CardTableV2(props: ICardTableV2Props) {
     }
     newColumns = newColumns.map((c) => ({
       ...c,
-      isResizable: c.isResizable === false ? false : true,
+      isResizable: c.isResizable !== false,
```
breeswish, line 171: How about `c.isResizable ?? true`

ui/lib/apps/SearchLogs/components/SearchHeader.tsx
```diff
@@ -119,18 +119,14 @@ export default function SearchHeader({ taskGroupID }: Props) {
...
+      const id = result.data.task_group?.id
```
baurine, line 123: Maybe `result?.data?.task_group?.id` is better.

ui/lib/components/CardTableV2/index.tsx
Rustin-Liu: Refactored.

ui/lib/apps/SearchLogs/components/SearchHeader.tsx
Rustin-Liu: Refactored.

ui/lib/apps/SearchLogs/components/SearchHeader.tsx
Rustin-Liu: I think result is OK :smile:

breeswish: What if there are network errors or server side errors?

Rustin-Liu: Did you mean `const result = await client.getInstance().logsTaskgroupPut(req)` will throw Error?

I will check it tonight.

baurine: We have a global axios error handler to handle network and auth error (but only for these 2 kinds of errors). Ideally, we should use try...catch to wrap all API requests, but we don't do that for many API requests. (maybe we should add a wrapper first). I will continue to refine handle error in PR #602

Rustin-Liu: Fixed, PTAL~

ui/lib/apps/SearchLogs/components/SearchHeader.tsx
```diff
@@ -121,13 +121,15 @@ export default function SearchHeader({ taskGroupID }: Props) {
 
       try {
         const result = await client.getInstance().logsTaskgroupPut(req)
-        const id = result.data.task_group?.id
+        const id = result?.data?.task_group?.id
         if (!id) {
-          throw new Error('Invalid server response')
+          Modal.error({
+            content: 'Invalid server response',
+          })
```
breeswish, line 128: The original way is simpler, avoid writing `Modal.error` twice :)

Rustin-Liu: But I think use Error to control the dataflow  is  not  a good  idea.

Rustin-Liu: I search this issue and find an answer: https://stackoverflow.com/questions/47015693/how-to-fix-throw-of-exception-caught-locally

But both OK for me, If you don't want this. I will revert this change later.

Rustin-Liu: Reverted.

PR #657, opened 2020-06-23
fix e2e test

### LINKED ISSUE #642: Fix e2e test
The e2e test broke after the pages are changed.

baurine:
close #642

### DIFF
--- a/ui/tests/e2e/search_log.test.ts
+++ b/ui/tests/e2e/search_log.test.ts
@@ -78,6 +78,6 @@ describe('Search Logs', () => {
       expect(content).toContain('Welcome to TiDB')
       expect(content.includes('Welcome to TiKV')).toBe(false)
     },
-    30 * 1000
+    25 * 1000
   )
 })
--- a/ui/tests/e2e/search_log.test.ts
+++ b/ui/tests/e2e/search_log.test.ts
@@ -61,6 +61,6 @@ describe('Search Logs', () => {
       expect(content).toContain('Welcome to TiDB')
       expect(content.includes('Welcome to TiKV')).toBe(false)
     },
-    20 * 1000
+    30 * 1000
   )
 })
--- a/ui/tests/e2e/search_log.test.ts
+++ b/ui/tests/e2e/search_log.test.ts
@@ -56,6 +56,8 @@ describe('Search Logs', () => {
         'div[data-e2e="search-result"] div[role="presentation"]:last-child'
       )
       console.log('get the result')
+      const url = await page.url()
+      console.log('current url:', url)
       const content = await logsTable.evaluate((node) => node.innerText)
       expect(content).toContain('Welcome to TiDB.')
       expect(content.includes('Welcome to TiKV')).toBe(false)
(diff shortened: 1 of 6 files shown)


baurine:
Finally, it gets success. But it still fails randomly sometimes, time out or says get an empty logs search result. I have no idea.

ui/tests/e2e/search_log.test.ts
```diff
@@ -22,32 +22,62 @@ describe('Search Logs', () => {
...
 
       // find search form
       const searchForm = await page.waitForSelector('form#search_form')
 
+      // choose time range
+      await ppExpect(searchForm).toClick(
+        'button[data-e2e="timerange-selector"]'
+      )
+      await ppExpect(page).toClick(
+        'div[data-e2e="common-timeranges"] div:last-child'
```
breeswish, line 40: what does last-child mean?

ui/tests/e2e/search_log.test.ts
```diff
@@ -22,32 +22,62 @@ describe('Search Logs', () => {
...
       // select TiDB component
-      await ppExpect(searchForm).toClick('div[data-e2e="components_selector"]')
+      // https://stackoverflow.com/questions/59882543/how-to-wait-for-a-button-to-be-enabled-and-click-with-puppeteer
+      await page.waitForSelector('div#instances input:not([disabled])')
+      await ppExpect(searchForm).toClick('div#instances')
       // components selector dropdown is a DOM node with absolute position
       // and its parent is body, failed to add id or data-e2e to it
-      await ppExpect(page).toClick('span[title="TiDB"]')
+      // cancel select PD and TiKV, and only remain TiDB
+      await ppExpect(page).toClick('span', { text: 'PD' })
```
breeswish, line 57: a span with text 'PD' may occur very frequently. Can we make it more accurate?

ui/tests/e2e/search_log.test.ts
baurine: As its name, it means the element is its parent's last child.

image

ui/tests/e2e/search_log.test.ts
baurine: I think so, let me try it.

ui/tests/e2e/search_log.test.ts
breeswish: Ok, I got it. Maybe it's better to modify our time range picker and provide some better ids :) It will better adapt to structural or style changes.

ui/tests/e2e/search_log.test.ts
baurine: Not easy to add additional props to the `span` directly, so change to add `data-e2e="table-with-filter"` to its container. The matching changes to:

```
await ppExpect(page).toClick('div[data-e2e="table-with-filter"] span', {
    text: 'PD',
})
```

ui/tests/e2e/search_log.test.ts
baurine: Add ``data-e2e={`timerange-${seconds}`}`` to each item.

PR #662, opened 2020-06-30
ui: refine code and error handle

### LINKED ISSUE #594: ui: make StatementsTable and SlowQueriesTable self-managed in the overview page
related PR: #591 

move the `useStatement()` and `useSlowQuery()` methods into the component self.

### LINKED ISSUE #595: ui: rename CardTableV2 to CardTable
related PR: #591

### LINKED ISSUE #596: ui: improve error alerts
Move error alert out the table for Statement and SlowQuery.

baurine:
close #594, close #595, close #596, close #602 

What did:

- Extract StatementsTable and SlowQueriesTable from Overview page ( #594 )
- Rename CardTableV2 to CardTable ( #595 )
- Use key for the message box to avoid multiple same message boxes
- Only show the first error if has multiple errors at some cases
- Show "Network Error" in the Alert component as well

  企业微信20200604045627

  企业微信20200604045943

Some other screenshots:

企业微信20200630012826

企业微信20200630012854

I planned to move Alert out of CardTable, then I found it doesn't work because we need to show alert between the table title and table body. If we move the Alert out of the CardTable, then it looks like this:

企业微信20200630011837


ui/lib/components/CardTable/index.tsx
```diff
@@ -215,27 +216,23 @@ function CardTableV2(props: ICardTableV2Props) {
...
-              columns={finalColumns}
-              items={finalItems}
-              {...restProps}
-            />
-          </div>
-        )}
+      {errMessages.map(
+        (msg, idx) =>
+          msg && <Alert key={idx} message={msg} type="error" showIcon />
+      )}
```
baurine, line 222: I tried to move the Alert of the CardTable but found it doesn't work because the Alert displays out of the table title which is not expected.

ui/lib/apps/ClusterInfo/components/InstanceTable.tsx
```diff
@@ -78,6 +79,19 @@ export default function ListPage() {
   } = useClientRequest((cancelToken) =>
     client.getInstance().getPDTopology({ cancelToken })
   )
+  const errMessages = useMemo(() => {
+    let errMsgs: string[] = []
+    if (errTiDB) {
+      errMsgs.push('Load TiDB instances failed')
```
breeswish, line 85: We'd better provide detailed error information

ui/lib/components/CardTable/index.tsx
breeswish, line 222: you don't need to try to keep the same looking as the old one. You can refine and find a way that the code is simple and the user can know about the error.

breeswish:
> I planned to move Alert out of CardTable, then I found it doesn't work because we need to show alert between the table title and table body

How about not displaying the table totally (and substitute it with the error bar)?

baurine:
> Thanks! Mostly fine, is it possible to remove `apiErrorMsg` and encapsulate it in a component, something like
> 
> ```js-jsx
> <ErrorBar error={error} />
> ```
> 
> so that using it can be simple?

Good idea, will try it later.

ui/lib/apps/ClusterInfo/components/InstanceTable.tsx
baurine: I think so.

breeswish:
Is this PR reviewable now? If not, you can add a WIP :)

baurine:
Yep, just ready.

How about this:

- Use a list to show multiple errors (show at most 3 kinds of errors):

  企业微信20200713015141

- Use the original style for a single error:

  image

>  >  I planned to move Alert out of CardTable, then I found it doesn't work because we need to show alert between the table title and table body

> How about not displaying the table totally (and substitute it with the error bar)?

The key point is that in the overview page, the statements and slow queries table have a table title shows above the alert and table body:

企业微信20200713015341

If we hide the whole table, the title will be hidden as well, which looks not good. Except that we re-implement the titles by an extra Card component.

PR #663, opened 2020-06-30
remember the selected legends for metric chart

### LINKED ISSUE #598: metric chart: hidden item displays again after reloading data
## Bug Report

**What did you do?**

**What did you see instead?**

baurine:
close #598 

Before:

After:

- [x] TODO: fix the new warning when toggling the legend 

image


baurine:
After comparing the demo code provided by echarts-for-react, I found the simplest way is just changing the `noMerge` prop value to its default false, then echarts will keep the old state inside itself. This change won't affect the result.

PR #673, opened 2020-07-06
Docs : Add environment variable declare step in  Contributing

ericsyh:
## Why for change 
When following the contribute docs to set the development workspace, i found the step of environment variable declare is missed and get the warning that `Command 'tiup' not found`.

## What did i do
I update the contribute docs and add the required step. 

## Test
No code change 

## Impact
No impact

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,7 +31,7 @@ Declare the global environment variable:
    > After the installation, TiUP displays the absolute path of the corresponding `profile` file. You need to modify the following `source` command according to the path.
 
 ```bash
-source .bash_profile
+source ~/.bash_profile
 ```
 
 Start a local TiDB cluster:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -24,6 +24,16 @@ Download and install TiUP:
...
 ```
 
+Declare the global environment variable:
+
+   > **Note:**
+   >
+   > After the installation, TiUP displays the absolute path of the corresponding `profile` file. You need to modify the following `source` command according to the path.
+
+```bash
+source .bash_profile
```
breeswish, line 34: Good catch, how about:

```suggestion
source ~/.bash_profile
```

ericsyh:
@breeswish What's the `command` to trigger run test again? Seems all open PRs are blocked by e2e test.

baurine:
> @breeswish What's the `command` to trigger run test again? Seems all open PRs are blocked by e2e test.

e2e test is still in fixing, but it is not required.

Issue #675, opened 2020-07-08
i18n: the default language value is incorrect 


ericsyh:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

**What did you do?**

I setup the TiDB Dashboard environment then signed in and go to the Setting page. 

**What did you expect to see?**

The **Language & Location** value should be **简体中文** or **English**. 

**What did you see instead?**
I got **zh** instead. 

![image](https://user-images.githubusercontent.com/10498732/86870706-79c55d00-c10b-11ea-826b-f56a4d492399.png)

**What version of TiDB Dashboard are you using (`./tidb-dashboard --version`)?**

The latest version: 
```
Release Version: ui_release_20200619_3-14-g00bca6f
Git Commit Hash: 00bca6f05d8e3d348a2b530204d6c4929140e7b8
Git Branch: master
UTC Build Time:  2020-07-08 01:44:05
```

breeswish:
This is not an expected behaviour. @baurine PTAL if you are free

baurine:
Ahh, looks weird, I even can't find the `zh` keyword in the source code... What's your browser version and system? Does it display correctly after refreshing the browser?

baurine:
I guess the language code i18next get from the browser is 'zh' instead of 'zh-CN'. We need to refine it. @ericsyh Would you like to help check what's the value of `i18n.language` in the `ui/lib/apps/DashboardSettings`?

baurine:
I changed my browser language to 'zh-TW', get a 'zh-TW' option in language selector, and a whole page with the English language.

![image](https://user-images.githubusercontent.com/1284531/87000422-9cc53f00-c1e7-11ea-93e5-5589e10af03b.png)

PR #677, opened 2020-07-09
UI: Add a whitelist in i18next to fix the issue #675

ericsyh:
## Why for change
To fix https://github.com/pingcap-incubator/tidb-dashboard/issues/675.
The root cause of issue is we declare two languages as `zh-CN` and `en` in i18next while some situations browser will return `zh` (Under my case the browser is `Chrome 83` the OS is `Windows 10`) which is out of our language definition. 

## What did i do
I added a whitelist only for `zh-CN` and `en` in i18next to block other cases. With the whitelist i18next will fallback to `en` by default if browser return language out of our definition like `zh` or `en-US`.

## Test
Passed on my environment. Now i18next got `en` as `i18nextLng`

image


ericsyh:
@baurine PTAL :)

baurine:
So by this way, 'zh-HK' / 'zh-TW' / 'zh-SG' (https://www.andiamo.co.uk/resources/iso-language-codes/) will be fallback to 'en' as well, not sure whether i18next has a customized fallback strategy.

ericsyh:
Seems no customized fallback provided in i18next but has [custom Detector](https://github.com/i18next/i18next-browser-languageDetector#adding-own-detection-functionality).  I tried it but do not figure it out  :(

baurine:
According to the document: https://www.i18next.com/principles/fallback

1. Per default locals containing region or script will take a translation from the pure language file if not found. (https://www.i18next.com/principles/fallback#locals-resolving). It means 'zh-XX' will fallback to 'zh' automatically
1. We can define an object for `fallbackLng` (https://www.i18next.com/principles/fallback#fallback-language) like this:

    ```
    fallbackLng: { 
        'zh-HK': ['zh-CN', 'en'], 
        'zh-TW': ['zh-CN', 'en'],
        ...
    }
    ```

I prefer to use the way 1. We can change all places used by 'zh-CN' to 'zh'. (for example, zh-CN.yaml to zh.yaml), this can prevent 'zh' /  'zh-HK' / 'zh-TW' / 'zh-SG' / 'zh-XXX' except 'zh-CN' fallback to 'en' with no more configuration. (need test to confirm)

@breeswish , how do you think about it?

baurine:
Or I can try the custom detector later.

breeswish:
Yes, I also think solution 1 looks best 

https://github.com/pingcap-incubator/tidb-dashboard/pull/677#issuecomment-655906807

baurine:
Thanks for your contribution! LGTM, I will have a test soon.

ui/lib/apps/DashboardSettings/translations/zh.yaml
```diff
@@ -0,0 +1,5 @@
+dashboard_settings:
```
breeswish, line 1: We no longer have this file, you can remove it

ericsyh: I see, you removed dashboard setting to the user profile page in previous commit. I will remove it.

ericsyh: Push a new commit https://github.com/pingcap-incubator/tidb-dashboard/pull/677/commits/436817df5977987d29bbfb89da4c03801a5db880 to remove the dashboard setting page.

breeswish:
@all-contributors please add @ericsyh  for code

PR #689, opened 2020-07-14
forwarder: evict invalid current picked remote

Fullstop000:
This PR does fixes and enhancements below
- use `FetchTiDBTopology` to replace `getServerInfo`
- Evict invalid `p.current` when the remote is not contained in `remotes`


breeswish:
Verified it works!

PR #697, opened 2020-07-19
build(deps): sql-formatter-plus-plus -> sql-formatter-plus

TennyZhuang:
Revert parts of https://github.com/pingcap-incubator/tidb-dashboard/pull/489

Latest Firefox stable release has supported [Unicode property escapes](https://caniuse.com/#search=Unicode%20property%20escapes)

Signed-off-by: TennyZhuang <zty0826@gmail.com>

### DIFF
--- a/ui/lib/utils/formatSql.ts
+++ b/ui/lib/utils/formatSql.ts
@@ -1,4 +1,4 @@
-import sqlFormatter from 'sql-formatter-plus-plus'
+import sqlFormatter from 'sql-formatter-plus'
 
 export default function formatSql(sql?: string): string {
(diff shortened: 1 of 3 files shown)


kennytm:
I agree with the discussion in the Slack channel that

> it's too new. Maybe we can remove it after several months

We should at least wait until Firefox 68 ESR went end-of-life on September 22nd.

breeswish:
Thanks for the contribution! I would like to hold this PR until our supported browsers all have this feature.

breeswish:
Considering that this may further break our compatibility promise, I think this PR may not be accepted in recent time.

PR #700, opened 2020-07-20
Refine diagnosis

### LINKED ISSUE #694: Refine system diagnose page
## Feature Request

- N/A

### the cluster-diagnose page such as below:

image

### the cluster-report page such as below (almost no change)

image

baurine:
resolve #694


baurine:
Almost done, current effect:



@breeswish @crazycs520 

Remain work:

- show subvalues with collapse mode default
- expand row and subvalues

baurine:
The functions are almost done:

Overview:



Support auto calculates the start time:



Support to start all diagnosis or just a single diagnosis:



Support show or hide sub-values, and expand or collapse a row:



Please help review first, @crazycs520 @breeswish  thanks! 

Some wording should be adjusted for the system report later.

This feature cost much longer much than I thought, some times the thread is not very clear when developing.

baurine:
image

differ the sub value row with a different color (maybe we need to use a better color)

breeswish:
Some immature ideas:

#### A. Maybe we can further simplify the start process for "current" cluster, to better encourage using it

360 and 腾讯电脑管家 has great examples about how to do diagnostics for the current system: provide a one-click button

image

We can preserve "specified time diagnostics" as an advanced feature. However maybe it is also better to remove diagnosics items that based on current status for "specified time diagnostics", like config diagnostics? @crazycs520 

#### B. Maybe it is a good idea to display "success" diagnostics items as well.

This gives user a blueprint of the overall diagnostics rules, i.e. capability.

Example: Azure, 360, QQ, AlertManager

(note: green items are generally not expanded by default)

image
image
image

#### C. Make it more clear for diagnostics items

Currently we simply list all diagnostics results. Maybe better to:

1. Use different colors to mark different severity items
2. Give some more human readable descriptions and better group some items, e.g. 
   - ✔️ Check swap is not enabled
   - ✖ Check swap is not enabled
      - Swap is enabled for some hosts. Recommend to disable swap to improve performance. [Link: How to fix]
      - ✖Host 127.0.0.1 (swap bytes = ..., expect 0)
      - ✖Host 127.0.0.2 (swap bytes = ..., expect 0)
      - ✔️Host 127.0.0.3 (swap bytes = 0)

   Instead of `node-load      swap-memory-used   node   127.0.0.1   11714 ...  0 ....`

3. We can move some rules to alert manager?

Again, I would recommend the 360 and QQ :) Good examples of simple and easy to use UI.

image

baurine:
@breeswish @crazycs520 update the wordings, all is done, PTAL, thanks!

crazycs520 (commented):
LGTM

ti-srebot (approved):

PR #707, opened 2020-07-29
Update router style

### LINKED ISSUE #701: Move query parameters to URL query part
## Feature Request

Now we have telemetry feature, which records the URL part without `?xxx`. However since some query parameters are placed in paths, like `/search_logs/details/:id`, this makes it not possible to group such URLs together.

Move all query parameters to the URL query part.

baurine:
close #701 : Move query parameters to URL query part for diagnose / search logs / instance profiling apps.

image


baurine:
> How about creating a
> 
> ```
> function useQueryParams(..) {
> 
> }
> ```
> 
> Then `const { id } = useParams()` can be changed to `const { id } = useQueryParams()`

I thought that, but `new URLSearchParams(search)` returns a class instead of an object, we need to use `get()` method to get the value of a key, else we parse it manually to return an object.

Even we return an object, we may not promise the id has value, we can't give it a default value inside the `useQueryParams()`. (maybe we can return a `any` type)

PR #713, opened 2020-08-03
feature: Query Editor

breeswish:
This PR impls a new feature Query Editor, as below:

image

image


ui/lib/apps/QueryEditor/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+      })
+      setResults(resp.data)
+    } catch (ex) {
+      Modal.error({
+        content: ex.message,
+      })
+    }
+    setRunning(false)
+    editor.current?.editor.focus()
+  }, [])
```
baurine, line 48: Not sure whether we can use `useClientRequest()` to simplify the code like this:

```
const {data, isLoading, error, sendRequest: handleRun} = useClientRequest((cancelToken) => 
    client.getInstance().queryEditorRun({
        max_rows: MAX_DISPLAY_ROWS,
        statements: editor.current?.editor.getValue(),
    }, {cancelToken}), 
    { immediate: false, afterRequest: () => editor.current?.editor.focus() }
)
```

Then we don't need `results` and `isRunning` anymore.

baurine: Let me try it.

breeswish: Unlike queries, for operational errors mostly there will be a modal dialog to stop further user actions. Displaying modal dialog using this pattern can be pretty hard.

PR #723, opened 2020-08-06
[WIP] *: initial plugin skeleton

kennytm:


breeswish:
Plugin is not our P1 tasks, close for now. It will be picked up again later, maybe in 2021 Q1.

PR #724, opened 2020-08-06
return zip instead of tar or gzip

### LINKED ISSUE #718: Support .zip format for profile
## Feature Request

Because Windows machine cannot use tar.gz

baurine:
resolve #718


breeswish:
Instance profiling may need to be updated as well (which is what issue targets at)

baurine:
> Instance profiling may need to be updated as well (which is what issue targets at)

Yep, working on it now.

baurine:
TODO: find a windows environment to test it.

pkg/apiserver/logsearch/pack.go
```diff
@@ -14,101 +14,57 @@
...
-	if err != nil {
-		return err
+		c.Status(http.StatusInternalServerError)
+		_ = c.Error(err)
+		return
 	}
+	defer temp.Close()
 
-	_, err = io.Copy(tw, f)
+	err = utils.CreateZipPack(temp, filePaths)
```
breeswish, line 62: Will it be too slow to zip again when there are a lot of logs? (for example, 5 GB)

Note that previously it is a tar stream on the fly using io.Copy, which is roughly a memory copy and does not need to wait for everything to finish as well.

baurine: Yep, this may be a problem, let me check it.

baurine: zip provide an option to allow us not to compress the files, ref: https://stackoverflow.com/a/23717581/2998877 , has updated the code.

baurine:
Hi @breeswish , now the zip pack file is streamed, PTAL, thanks! ref: https://stackoverflow.com/questions/57429256/how-to-generate-zip-7z-archive-on-the-fly-in-a-http-server-using-gin/57434338#57434338

pkg/apiserver/logsearch/pack.go
baurine: stream the zip pack, ref: https://stackoverflow.com/a/57434338/2998877

breeswish:
Good job! Mostly lgtm. The only thing I'm worrying is that you are using defer in loops, which usually behaves like something you are not expecting. Would you like to move it to a standalone function so that defer will run immediately after each loop iteration?

baurine:
> Good job! Mostly lgtm. The only thing I'm worrying is that you are using defer in loops, which usually behaves like something you are not expecting. Would you like to move it to a standalone function so that defer will run immediately after each loop iteration?

Ok, let me try.

breeswish (approved):

ti-srebot (approved):

PR #732, opened 2020-08-10
Make e2e test not often fail

baurine:
- Add an extra step to debug TiUP
- Search logs by keyword "welcome" to reduce the result size

  After this change, the "Test / UI" job has success 5 times continuously. (Although I don't know why the big result size causes the job not stable)

  image


baurine:
@breeswish , PTAL, thanks!

PR #741, opened 2020-08-24
ui: Support sharing session

breeswish:
TiDB Dashboard requires the knowledge of a high privilege account to work. It may not be friendly for 3rd-party (like PingCAP) remote access.

Sharing session allows a signed-in user to share its session to other people, so that other people can access TiDB Dashboard without knowing the credentials. Sharing session may extend the lifetime of the current session, so that a shared session is not allowed to be shared again.

image
image
image
image
image
image
image


baurine (approved):
LGTM!

ti-srebot (approved):

PR #752, opened 2020-09-08
Fix ui test of ci

baurine:


baurine:
Get `fail reason: Sign in failed: Failed to connect to TiDB` log, seem referees to issue: https://github.com/pingcap/tidb/issues/19731 . Try to test v4.0.5

.github/workflows/e2e-test.yaml
```diff
@@ -46,23 +46,30 @@ jobs:
...
         run: |
           make ui
         env:
           NO_MINIMIZE: true
           CI: true
           REACT_APP_MIXPANEL_TOKEN: ""
+      # - name: Wait TiUP Playground
+      #   run: |
+      #     chmod u+x scripts/wait_tiup_playground.sh
+      #     scripts/wait_tiup_playground.sh 15 20
```
baurine, line 65: Will uncomment it when tiup release its new version which fixed the bug that `tiup playground display` command doesn't work.

.github/workflows/e2e-test.yaml
```diff
@@ -46,23 +46,30 @@ jobs:
         run: |
           curl --proto '=https' --tlsv1.2 -sSf https://tiup-mirrors.pingcap.com/install.sh | sh
           source /home/runner/.profile
-          tiup update --nightly
-          tiup playground nightly --tiflash=0 &
+          tiup install tidb:v4.0.5 tikv pd prometheus
+          tiup install playground
```
baurine, line 50: Pre-install these components in the foreground to make the playground run asap in the background.

baurine:
TiDB v4.0.5 works fine, so it can be confirmed the root cause is pingcap/tidb#19731 .  Should we use the stable TiDB instead of the nightly version?

.github/workflows/e2e-test.yaml
```diff
@@ -46,23 +46,30 @@ jobs:
         run: |
           curl --proto '=https' --tlsv1.2 -sSf https://tiup-mirrors.pingcap.com/install.sh | sh
           source /home/runner/.profile
-          tiup update --nightly
-          tiup playground nightly --tiflash=0 &
+          tiup install tidb:v4.0.5 tikv pd prometheus
```
breeswish, line 49: It can be simply

```
tiup playground v4.0.5 --tiflash=0
```

.github/workflows/e2e-test.yaml
```diff
@@ -46,23 +46,30 @@ jobs:
...
+          tiup install playground
+          tiup update --all
+          tiup update --self
+          source /home/runner/.profile
```
breeswish, line 53: After using a fixed version maybe we no longer need any `tiup update`

breeswish (commented):
The rest LGTM

.github/workflows/e2e-test.yaml
baurine: It may block the jobs, let me have a try.

.github/workflows/e2e-test.yaml
baurine: Yep, will remove it.

.github/workflows/e2e-test.yaml
breeswish: Right. Maybe we need something to wait TiDB started successfully, like https://github.com/tikv/copr-test/blob/master/push-down-test/run-tests.sh#L283

baurine: I write a shell to wait for it by check the` tiup playground display` command, but the tiup 1.1.1 version has a bug that `tiup playground display` doesn't work normally... will use it when tiup release 1.1.2

ui/tests/e2e/login.test.ts
```diff
@@ -20,7 +20,13 @@ describe('Login', () => {
 
       await ppExpect(page).toFill('input#tidb_signin_password', 'any')
       await ppExpect(page).toClick('button#signin_btn')
-      await ppExpect(page).toMatch('TiDB authentication failed')
+
+      const failReason = await page.waitForSelector(
+        'form#tidb_signin div[data-e2e="password"] div:last-child'
```
breeswish, line 25: This `div:last-child` is too magical.. Can we await for pages to contain some string?

baurine: It seems no way to add the extra flag to the error message self except to its parent container likes `data-e2e="password"`.

baurine: Or we can simply remove the `div:last-child`.

baurine: Just got what you mean. I print the failed reason content to help debug what the exact failed reason. It may not be "TiDB authentication failed".

breeswish: Oh, got it!

PR #768, opened 2020-10-13
ui: unify naming style in *.module.less

unbyte:
- add css-loader option so can use css module with camelCase style
- unify naming style in *.module.less to hyphen delimited style

### DIFF
--- a/ui/dashboardApp/layout/signin/index.module.less
+++ b/ui/dashboardApp/layout/signin/index.module.less
@@ -14,14 +14,14 @@
   align-items: stretch;
 }
 
-.contantContainer {
+.content-container {
   min-width: @content-width;
   width: 38%;
   background: #fff;
   position: relative;
 }
 
-.dialogContainer {
+.dialog-container {
   height: 100%;
   display: flex;
   align-items: center;
@@ -49,13 +49,13 @@
   margin-bottom: 80px;
 }
 
-.signInButton {
+.sign-in-button {
   margin-top: 10px;
   margin-bottom: 20px;
   font-size: 0.9rem;
 }
 
-.extraLink {
+.extra-link {
   font-size: 0.8rem;
   margin: 15px 0;
   a {
@@ -72,7 +72,7 @@
   }
 }
 
-.alternativeFormLayer {
+.alternative-form-layer {
   position: absolute;
   left: 0;
   top: 0;
@@ -82,7 +82,7 @@
   z-index: 2;
 }
 
-.alternativeCloseButton {
+.alternative-close-button {
   font-size: 1rem;
   border: 0;
   background: #fff;
@@ -99,7 +99,7 @@
   }
 }
 
-.alternativeButton {
+.alternative-button {
   .btn;
   .btn-default;
 
@@ -155,12 +155,12 @@
 }
 
 .container :global {
-  .formAnimation {
+  .form-animation {
     animation: 0.2s @ease-out-circ 0.5s antZoomBigIn;
     animation-fill-mode: both;
     animation-iteration-count: 1;
   }
-  .landingAnimation {
+  .landing-animation {
     animation: 0.5s linear 0 antFadeIn;
     animation-fill-mode: both;
     animation-iteration-count: 1;
--- a/ui/lib/components/CardTable/index.module.less
+++ b/ui/lib/components/CardTable/index.module.less
@@ -1,6 +1,6 @@
 .card-table {
   :global {
-    .ms-details-row {
+    .ms-detailsRow {
       > :first-child.ms-DetailsRow-fields {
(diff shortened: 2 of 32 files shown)


ui/lib/components/CardTable/index.module.less
```diff
@@ -1,6 +1,6 @@
-.cardTable {
+.card-table {
   :global {
-    .ms-DetailsRow {
+    .ms-details-row {
```
baurine, line 3: Not sure whether ".ms-DetailsRow" is a class for the Microsoft Fluent DetailsList component, need to check it.

baurine (approved):
most LGTM, just need to check more for the changes for ":global" classes, maybe it is a class name for some third-party components.

@breeswish how do you think about it?

ti-srebot (approved):

ui/dashboardApp/layout/signin/index.module.less
```diff
@@ -155,12 +155,12 @@
...
 
 .container :global {
-  .formAnimation {
+  .form-animation {
     animation: 0.2s @ease-out-circ 0.5s antZoomBigIn;
     animation-fill-mode: both;
     animation-iteration-count: 1;
   }
-  .landingAnimation {
+  .landing-animation {
```
baurine, line 163: These 2 class names are used in:

```
        <AppearAnimate
          className={styles.landing}
          motionName="landingAnimation"
        />
```

Not sure whether this breaks the behavior, need to check.

ui/lib/components/CardTable/index.module.less
```diff
@@ -1,6 +1,6 @@
...
       > :first-child.ms-DetailsRow-fields {
```
baurine, line 4: So does for ".ms-DetailsRow-fields".

breeswish (commented):
How about using camelCase in both CSS and JS? In this way we can just copy-paste style names, no need to transform by hand.

unbyte:
In fact, the naming style depends on the convention of the project.

However, good naming convention should always be followed in projects. There are already many naming conventions for CSS. Here is a more detailed one: https://cssguidelin.es/

*what's more, CSS is case insensitive. Using camel case may lead to unpredictable bug codes*

breeswish:
This is a good chance for us to establish a true naming convention, so that I think we can choose the best one for us.

We are using css modules so that I don't get the point of the defects of camelCases. These css style names are referenced using JavaScript symbols, in a case sensitive way.

unbyte:
Actually [CSS Module](https://github.com/css-modules/css-modules) recommends the use of camelCase in CSS, and the reason is https://github.com/css-modules/css-modules/pull/222#issuecomment-291013528 .

However, hyphen delimited style in CSS is still a mainstream and good convention so css loader provides option to auto camelCase css in js. And it's the same as the mental model of how [JSX](https://reactjs.org/docs/dom-elements.html#style) processes CSS, i.e. write `what-what` in css and map into `whatWhat` in js.

baurine:
Just thought of a small concern, when searching class name in the code for debugging, it needs to search twice if we use different styles between the CSS file and js file, that's may not very convenient.

unbyte:
intelligent IDEs can jump to css from `styles.xxxXXX` in js and jump to js from `xxx-xxx` in css

baurine:
> intelligent IDEs can jump to css from `styles.xxxXXX` in js and jump to js from `xxx-xxx` in css

Does that work for VS Code? I just tried it seems not to work... (maybe I miss some extensions?)

breeswish:
> > intelligent IDEs can jump to css from `styles.xxxXXX` in js and jump to js from `xxx-xxx` in css
> 
> Does that work for VS Code? I just tried it seems not to work... (maybe I miss some extensions?)

My dev environment (VsCode) does not support this operation as well. This will not be a problem if style names are camelCase.

> However, hyphen delimited style in CSS is still a mainstream and good convention so css loader provides option to auto camelCase css in js. And it's the same as the mental model of how JSX processes CSS, i.e. write what-what in css and map into whatWhat in js.

I think the main point is that, the mapping need to be performed manually by us during developing, if you want to jump or find symbols between CSS and JS, i.e. write xxx-yyy in CSS while search for both xxx-yyy and xxxYyy in order to find where it is, vice versa.

unbyte:
yeah, in view of the fact that the react community does not have a unified consensus, and the project has been used to using camel case in CSS, so I decided to close this PR, or could I change this PR to unifying naming convention to camel case?

baurine:
> yeah, in view of the fact that the react community does not have a unified consensus, and the project has been used to using camel case in CSS, so I'm going to close this PR, or could I change this PR to unifying naming style to camel case?

Prefer to unify the naming style to camel case.

PR #769, opened 2020-10-15
ui: display error message instead of stack trace on connect failure

kennytm:
The stack trace is useless without source map, and also it is very hard to understand what error is going on without the error message.

(Using `''+e` instead of `e.message` to be compatible with non-Exception errors)


breeswish (approved):
Good catch!

ti-srebot (approved):

PR #771, opened 2020-10-17
ui: break loop dependencies

### LINKED ISSUE #764: Remove the dependency loop hack for useStatementTableController
Related PR: https://github.com/pingcap-incubator/tidb-dashboard/pull/749#issuecomment-699729457

Do so for useSlowQueryTableContoller as well.

Code:

```ts
  // Notice: statements, tableColumns, selectedFields make loop dependencies
  const tableColumns = useMemo(
    () => statementColumns(statements, showFullSQL),
    [statements, showFullSQL]
  )
  // make selectedFields as a string instead of an array to avoid infinite loop
  // I have verified that it will cause infinite loop if we return selectedFields as an array
  // so it is better to use the basic type (string, number...) instead of object as the dependency
  const selectedFields = useMemo(
    () => getSelectedFields(visibleColumnKeys, tableColumns).join(','),
    [visibleColumnKeys, tableColumns]
  )
```

unbyte:
calculating selectedFields no longer need tableColumns

close #764

### DIFF
--- a/ui/lib/apps/Statement/utils/tableColumns.tsx
+++ b/ui/lib/apps/Statement/utils/tableColumns.tsx
@@ -12,7 +12,6 @@ import { Bar, Pre } from '@lib/components'
 import {
   formatVal,
   genDerivedBarSources,
-  IColumnWithSourceFields,
   TableColumnFactory,
 } from '@lib/utils/tableColumnFactory'
 
@@ -90,7 +89,7 @@ function avgMinMaxLatencyColumn(
 function errorsWarningsColumn(
   tcf: TableColumnFactory,
   rows?: { sum_errors?: number; sum_warnings?: number }[]
-): IColumnWithSourceFields {
+): IColumn {
   const capacity = rows
     ? max(rows.map((v) => v.sum_errors! + v.sum_warnings!)) ?? 0
     : 0
@@ -99,7 +98,6 @@ function errorsWarningsColumn(
     name: tcf.columnName('errors_warnings'),
     key,
     fieldName: key,
-    sourceFields: derivedFields.sum_errors,
     minWidth: 140,
     maxWidth: 200,
     columnActionsMode: ColumnActionsMode.clickable,
@@ -149,7 +147,7 @@ function avgMaxColumn<T>(
 export function statementColumns(
   rows: StatementModel[],
   showFullSQL?: boolean
-): IColumnWithSourceFields[] {
+): IColumn[] {
   const tcf = new TableColumnFactory(TRANS_KEY_PREFIX)
 
   return [
@@ -207,7 +205,6 @@ export function statementColumns(
       ...tcf.textWithTooltip('related_schemas', rows),
       minWidth: 160,
       maxWidth: 240,
-      sourceFields: derivedFields.related_schemas,
     },
   ]
 }
--- a/ui/lib/apps/Statement/utils/tableColumns.tsx
+++ b/ui/lib/apps/Statement/utils/tableColumns.tsx
@@ -11,7 +11,7 @@ import { StatementModel } from '@lib/client'
 import { Bar, Pre } from '@lib/components'
 import {
   formatVal,
-  genDerivedBar,
+  genDerivedBarSources,
   IColumnWithSourceFields,
   TableColumnFactory,
 } from '@lib/utils/tableColumnFactory'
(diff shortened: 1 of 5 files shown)


unbyte:
still need test

ui/lib/apps/Statement/utils/tableColumns.tsx
```diff
@@ -49,7 +99,7 @@ function errorsWarningsColumn(
     name: tcf.columnName('errors_warnings'),
     key,
     fieldName: key,
-    sourceFields: ['sum_errors', 'sum_warnings'],
+    sourceFields: derivedFields.sum_errors,
```
baurine, line 102: If we don't `getSelectedFields()` from `sourceFields` anymore, all places of `sourceFields` can be removed.

ui/lib/apps/Statement/utils/tableColumns.tsx
```diff
@@ -292,7 +207,7 @@ export function statementColumns(
       ...tcf.textWithTooltip('related_schemas', rows),
       minWidth: 160,
       maxWidth: 240,
-      sourceFields: ['table_names'],
+      sourceFields: derivedFields.related_schemas,
```
baurine, line 210: ditto.

ui/lib/utils/tableColumnFactory.tsx
```diff
@@ -229,26 +225,64 @@ export class BarColumn {
...
 ) {
   let fields: string[] = []
-  columns.forEach((c) => {
-    if (visibleColumnKeys[c.key] === true) {
-      if (c.sourceFields !== undefined) {
-        fields = fields.concat(c.sourceFields)
+  let sources: DerivedFields[keyof DerivedFields]
+  for (const columnKey in visibleColumnKeys) {
+    if (visibleColumnKeys[columnKey]) {
+      if ((sources = derivedFields[columnKey])) {
```
baurine, line 276: Curious what's this kind of usage? `((...))`

ui/lib/apps/Statement/utils/tableColumns.tsx
```diff
@@ -10,31 +10,81 @@ import { orange, red } from '@ant-design/colors'
...
+  avg_write_size: genDerivedBarSources('avg_write_size', 'max_write_size'),
+  avg_prewrite_regions: genDerivedBarSources(
+    'avg_prewrite_regions',
+    'max_prewrite_regions'
+  ),
+  avg_txn_retry: genDerivedBarSources('avg_txn_retry', 'max_txn_retry'),
+  avg_mem: genDerivedBarSources('avg_mem', 'max_mem'),
+  sum_errors: ['sum_errors', 'sum_warnings'],
+  related_schemas: ['table_names'],
+}
```
baurine, line 79: In fact, this solution is a bit similar to the old version that was replaced by bringing `sourceFields` and `IColumnWithSourceFields`, aka put the ref columns in an individual object, but this is better than before by reusing it with `statementColumns()` or `slowQueryColumns()`. So I think it's fine.

@breeswish how do you think about it?

ui/lib/utils/tableColumnFactory.tsx
```diff
@@ -97,46 +105,34 @@ export class TableColumnFactory {
...
     if (min) {
       sourceFields.push(min.fieldName)
     }
+
     return {
-      ...this.columnFromField(avg.fieldName as string),
-      name: this.columnName((displayTransKey || avg.fieldName) as string),
+      ...this.columnFromField(avg.fieldName),
+      name: this.columnName(displayTransKey || avg.fieldName),
       sourceFields,
```
baurine, line 136: ditto

ui/lib/utils/tableColumnFactory.tsx
unbyte: just to avoid lint reporting warnings

ui/lib/apps/Statement/utils/tableColumns.tsx
unbyte: IIRC, it's used by render function

baurine: > IIRC, it's used by render function

Nope, it should be only used for `getSelectedFields()`.

ui/lib/utils/tableColumnFactory.tsx
```diff
@@ -17,16 +17,20 @@ import {
   IColumnKeys,
 } from '@lib/components'
 
-type Bar<T> = { [key: string]: keyof T }
-type BarsConfig<T> = {
-  displayTransKey?: string // it is same as avg field name default
-  bars: [Bar<T>, Bar<T>, Bar<T>?] // [avg, max, min?]
+type Bar = {
```
baurine, line 20: Just found this type's name is the same as the imported component, can we change a name for it, for example: `interface IBar`. others LGTM.

baurine (approved):
LGTM

ti-srebot (approved):

PR #772, opened 2020-10-19
Refine store location, add zoom and pan

### LINKED ISSUE #737: Add zoom and pan to store topology
## Feature Request

Current UI does not look pretty when user has large screen:

image

While it is also hard to read when user screen is small:

image

Add support for zoom and pan for the graph. Initially, the visualization should display as 1:1 (the same font-size as the page) all nodes can be displayed, leaving some blanks. If not all nodes can be displayed, the graph is initially scaled-*out* to display a full view by default.

baurine:
close #737 

What did:

- Disable auto-scale when resizing store location tree component (cancel use viewbox)
- Add zoom and pan support for store location tree component

Screenshot:

Zoom and pan effect on the original component (auto-scale when resizing), just for comparison:

TODO:

- [x] add reset button


baurine:
@breeswish PTAL when you're free, thanks!

breeswish (approved):

ti-srebot (approved):

breeswish:
Looks like there is a bug that the canvas is not extended to the full screen:

image

Also maybe it is better to hint user to use "Ctrl+Scroll"?

breeswish:
BTW the refresh button doesn't refresh. Instead it *reset* positions. Maybe we can add reload functionality into it? (I didn't find a proper icon for reset but not reload..)

baurine:
> BTW the refresh button doesn't refresh. Instead it _reset_ positions. Maybe we can add reload functionality into it? (I didn't find a proper icon for reset but not reload..)

It's supposed to a reset button... but it is true we can add a reload function as well, will do it.

breeswish:
@baurine Yes, it's a reset button. But the icon is exactly the same as a refresh button. User won't know it's reset instead of refresh.

breeswish:
To avoid losing track, I opened https://github.com/pingcap/tidb-dashboard/issues/801 for issues described in https://github.com/pingcap/tidb-dashboard/pull/772#issuecomment-717208128

baurine:
> To avoid losing track, I opened #801 for issues described in [#772 (comment)](https://github.com/pingcap/tidb-dashboard/pull/772#issuecomment-717208128)

Got, thank you!

PR #776, opened 2020-10-22
ui: remove experimental API in checkBrowser.js

### LINKED ISSUE #770: Check browerlist whether really works
## Bug Report

**What did you do?**

Access dashboard in Chrome 5x version.

Work normally.

**What did you see instead?**

It crashes and says `Object.entries()` is not a function.

(This version of Chrome is supported in the browser list, so the checkBrowser.js doesn't show a prompt to say this browser version is too low.)

4.0.7

unbyte:
image

---

close #770 

when testing the related issue on the low version of the Firefox, we found that alert from checkBrowser.js can't appear for the experimental API checkBrowser.js uses is not supported by these low version browsers.

### DIFF
--- a/ui/public/checkBrowser.js
+++ b/ui/public/checkBrowser.js
@@ -40,7 +40,7 @@ function checkBrowser() {
     d.getElementsByTagName('a')[0].onclick = function () {
       d.getElementsByTagName('div')[0].style.top = '-60px'
     }
-    document.body.prepend ? document.body.prepend(d) : document.body.insertBefore(d, document.body.firstChild)
+    document.body.insertBefore(d, document.body.firstChild)
   }
 }
 
--- a/ui/public/checkBrowser.js
+++ b/ui/public/checkBrowser.js
@@ -40,7 +40,7 @@ function checkBrowser() {
     d.getElementsByTagName('a')[0].onclick = function () {
       d.getElementsByTagName('div')[0].style.top = '-60px'
     }
-    document.body.prepend(d)
+    document.body.prepend ? document.body.prepend(d) : document.body.insertBefore(d, document.body.firstChild)
   }
 }
 
--- a/ui/gulpfile.esm.js
+++ b/ui/gulpfile.esm.js
@@ -59,7 +59,7 @@ function updateBrowserList() {
 }
 
 task('gen:browserlist', () => {
-  return src('public/checkBrowser.js')
+  return src('public/compat.js')
     .pipe(updateBrowserList())
     .pipe(dest('public', { overwrite: true }))
 })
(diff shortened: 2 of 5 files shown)


breeswish (approved):

ti-srebot (approved):

ui/public/checkBrowser.js
```diff
@@ -40,7 +40,7 @@ function checkBrowser() {
     d.getElementsByTagName('a')[0].onclick = function () {
       d.getElementsByTagName('div')[0].style.top = '-60px'
     }
-    document.body.prepend(d)
+    document.body.prepend ? document.body.prepend(d) : document.body.insertBefore(d, document.body.firstChild)
```
baurine, line 43: I prefer to directly use `document.body.insertBefore(d, document.body.firstChild)`.

We can use 1 PR to resolve the problems of #770 together (will discuss it later).

baurine:
https://developer.mozilla.org/zh-CN/docs/Web/API/ParentNode/prepend

prepend can be used in chrome 54+, edge 17+, Firefox 49+

ui/public/checkBrowser.js
unbyte: yeah, however, the support for `firstChild` in chromes <=85 is unknown as caniuse shows

baurine: It doesn't matter, according to this doc: https://developer.mozilla.org/zh-CN/docs/Web/API/Node/insertBefore , even if firstChild exist or not, insertBefore will work fine.

kennytm: how about `childNodes[0]`.

but i highly doubt Chrome 85- *not* supporting `firstChild` which is in DOM 1 released back in 1997.

unbyte: just curious about why support is unknown. maybe there is a special reason?

baurine: I think it is fine, according to this doc: https://developer.mozilla.org/zh-CN/docs/Web/API/ParentNode/prepend , `prepend` polyfill is implemented by `insertBefore` and `firstChild`, even if `document.body.firstChild` is undefined, it works as expected, or we can consider using `appendChild`?

unbyte: `unknow` comes from no real data yet. see https://github.com/mdn/browser-compat-data/issues/6369
so yeah we can use `insertBefore` directly

unbyte:
since Firefox 46 is not in our support list, dashboard still can't work in it. However, alert works well now.
image

baurine (approved):
LGTM

PR #778, opened 2020-10-22
export statements

### LINKED ISSUE #565: Requirement Request: Data Export in statement
## Description

 The data of Statement UI is a complex aggregation of fields in statement tables. Since the query is usually hard to write manually, users are not easy to get these data. Thus for data in UI it should be allowed to be exported.

## Category

* Feature
* Usability

## Value

- Score: 3

## Workload Estimation

10 Person Day

baurine:
close #565 

What did:

- Export the queried statements result as .csv format file

Screenshots:

(mocked time delays)

WeCom20201026-110427@2x

WeCom20201026-110456@2x

### DIFF
--- a/pkg/apiserver/statement/statement.go
+++ b/pkg/apiserver/statement/statement.go
@@ -14,11 +14,10 @@
 package statement
 
 import (
-	"bufio"
-	"bytes"
 	"encoding/csv"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
@@ -33,6 +32,7 @@ import (
 	"go.uber.org/fx"
 	"go.uber.org/zap"
 
+	aesctr "github.com/Xeoncross/go-aesctr-with-hmac"
 	"github.com/pingcap-incubator/tidb-dashboard/pkg/apiserver/user"
 	"github.com/pingcap-incubator/tidb-dashboard/pkg/apiserver/utils"
 	"github.com/pingcap-incubator/tidb-dashboard/pkg/tidb"
@@ -304,31 +304,23 @@ func (s *Service) downloadTokenHandler(c *gin.Context) {
 		return
 	}
 	defer csvFile.Close()
-	// generate buf writer for tmpfile
-	csvBuf := bufio.NewWriter(csvFile)
-	// generate encrypted key
+
+	// generate encryption key
 	secretKey := cryptopasta.NewEncryptionKey()
 
-	rowBuf := bytes.NewBuffer(nil)
-	csvwriter := csv.NewWriter(rowBuf)
-	for _, csvRow := range csvData {
-		_ = csvwriter.Write(csvRow)
-		csvwriter.Flush()
-		row := make([]byte, len(rowBuf.String()))
-		_, _ = rowBuf.Read(row)
-
-		encrypted, err := cryptopasta.Encrypt(row, secretKey)
-		if err != nil {
-			_ = c.Error(err)
-			return
-		}
-		csvBuf.Write(encrypted)
+	pr, pw := io.Pipe()
+	go func() {
+		csvwriter := csv.NewWriter(pw)
+		_ = csvwriter.WriteAll(csvData)
+		pw.Close()
+	}()
+	err = aesctr.Encrypt(pr, csvFile, (*secretKey)[0:16], (*secretKey)[16:])
+	if err != nil {
+		_ = c.Error(err)
+		return
 	}
-	csvBuf.Flush()
 
-	// zip file and encrypt
-
-	// generate token by filepath
+	// generate token by filepath and secretKey
 	token, err := utils.NewJWTString("statements/download", csvFile.Name())
 	if err != nil {
 		_ = c.Error(err)
--- a/pkg/apiserver/statement/statement.go
+++ b/pkg/apiserver/statement/statement.go
@@ -14,8 +14,13 @@
 package statement
 
 import (
+	"encoding/csv"
 	"errors"
+	"fmt"
+	"io/ioutil"
 	"net/http"
+	"reflect"
+	"strconv"
 	"strings"
 
 	"github.com/gin-gonic/gin"
@@ -223,48 +228,16 @@ func (s *Service) getPlanDetailHandler(c *gin.Context) {
 // @Router /statements/download/acquire_token [get]
 // @Summary Generate a download token for downloading statements
 // @Produce plain
-// @Param time query string true "Query"
+// @Param q query GetStatementsRequest true "Query"
 // @Success 200 {string} string "xxx"
 // @Security JwtAuth
 // @Failure 401 {object} utils.APIError "Unauthorized failure"
 func (s *Service) downloadTokenHandler(c *gin.Context) {
-	time := c.Query("time")
-	token, err := utils.NewJWTString("statements/download", time)
-	if err != nil {
-		_ = c.Error(err)
-		return
-	}
-	c.String(http.StatusOK, token)
-}
-
-type DownloadStatementsRequest struct {
-	GetStatementsRequest
-	Token string `json:"token" form:"token"`
-}
-
-// @Router /statements/download [get]
-// @Summary Download statements
-// @Produce application/x-tar,application/zip
-// @Param q query DownloadStatementsRequest true "Query"
-// @Failure 400 {object} utils.APIError
-// @Failure 401 {object} utils.APIError "Unauthorized failure"
-func (s *Service) downloadHandler(c *gin.Context) {
-	var req DownloadStatementsRequest
+	var req GetStatementsRequest
 	if err := c.ShouldBindQuery(&req); err != nil {
 		utils.MakeInvalidRequestErrorFromError(c, err)
 		return
 	}
-	str, err := utils.ParseJWTString("statements/download", req.Token)
-	if err != nil {
-		utils.MakeInvalidRequestErrorFromError(c, err)
-		return
-	}
-	time := string(req.BeginTime) + "-" + string(req.EndTime)
-	if str != time {
-		utils.MakeInvalidRequestErrorFromError(c, errors.New("invalid token"))
-		return
-	}
-
 	db := utils.GetTiDBConnection(c)
 	fields := []string{}
 	if strings.TrimSpace(req.Fields) != "" {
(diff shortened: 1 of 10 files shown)


pkg/apiserver/statement/statement.go
```diff
@@ -210,3 +227,115 @@ func (s *Service) getPlanDetailHandler(c *gin.Context) {
...
+			default:
+				val = s.String()
+			}
+			row = append(row, val)
+		}
+		csvData = append(csvData, row)
+	}
+
+	// generate csv
+	tmpfile, err := ioutil.TempFile("", fmt.Sprintf("statements_%d_%d_*.csv", req.BeginTime, req.EndTime))
```
baurine, line 294: Maybe it's better to convert the timestamp to a human-friendly format.

ui/lib/apps/Statement/utils/useStatementTableController.ts
```diff
@@ -67,6 +67,9 @@ export interface IStatementTableController {
 
   tableColumns: IColumn[]
   visibleColumnKeys: IColumnKeys
+
+  getDownloadToken: () => Promise<any>
```
baurine, line 71: Promise<string>

pkg/apiserver/statement/statement.go
```diff
@@ -40,15 +50,23 @@ func NewService(p ServiceParams) *Service {
...
-	endpoint.Use(utils.MWConnectTiDB(s.params.TiDBClient))
-	endpoint.GET("/config", s.configHandler)
-	endpoint.POST("/config", s.modifyConfigHandler)
-	endpoint.GET("/time_ranges", s.timeRangesHandler)
-	endpoint.GET("/stmt_types", s.stmtTypesHandler)
-	endpoint.GET("/overviews", s.overviewsHandler)
-	endpoint.GET("/plans", s.getPlansHandler)
-	endpoint.GET("/plan/detail", s.getPlanDetailHandler)
+	{
+		endpoint.GET("/download", s.downloadHandler)
```
unbyte, line 54: maybe it's better to register it with other endpoints
```go
endpoint.GET("/plan/detail", s.getPlanDetailHandler)
endpoint.GET("/download", s.downloadHandler)
endpoint.GET("/download/acquire_token", s.downloadTokenHandler)
```

pkg/apiserver/statement/statement.go
```diff
@@ -210,3 +228,118 @@ func (s *Service) getPlanDetailHandler(c *gin.Context) {
...
+		_ = c.Error(err)
+		return
+	}
+
+	c.Writer.Header().Set("Content-type", "application/octet-stream")
+	c.Writer.Header().Set("Content-Disposition", "attachment; filename=\"statements.zip\"")
+	err = utils.StreamZipPack(c.Writer, []string{fielPath}, true)
+	if err != nil {
+		log.Error("Stream zip pack failed", zap.Error(err))
+	}
```
unbyte, line 344: I wonder how to deal with the temporary .csv files if don't delete them after each download (though this is not a good solution)?

pkg/apiserver/statement/statement.go
baurine: It is because the `/download` API doesn't and shouldn't need `auth.MWAuthRequired()` middleware.

pkg/apiserver/statement/statement.go
baurine: Because we put them in the system `/tmp` folder, so they will be cleaned up by the system, for example, they will be deleted automatically when rebooting.

pkg/apiserver/statement/statement.go
unbyte: got

pkg/apiserver/statement/statement.go
```diff
@@ -210,3 +228,118 @@ func (s *Service) getPlanDetailHandler(c *gin.Context) {
...
+			var val string
+			switch s.Interface().(type) {
+			case int:
+				val = strconv.FormatInt(s.Int(), 10)
+			default:
+				val = s.String()
+			}
+			row = append(row, val)
+		}
+		csvData = append(csvData, row)
```
breeswish, line 291: Do we have some 3rd-party library for easier use? This logic may be the same when exporting slow queries.

pkg/apiserver/statement/statement.go
```diff
@@ -210,3 +228,118 @@ func (s *Service) getPlanDetailHandler(c *gin.Context) {
 	}
 	c.JSON(http.StatusOK, result)
 }
+
+// @Router /statements/download/acquire_token [get]
+// @Summary Generate a download token for downloading statements
```
breeswish, line 233:
```suggestion
// @Summary Generate a download token for exported statements
```

pkg/apiserver/statement/statement.go
```diff
@@ -210,3 +228,118 @@ func (s *Service) getPlanDetailHandler(c *gin.Context) {
 	}
 	c.JSON(http.StatusOK, result)
 }
+
+// @Router /statements/download/acquire_token [get]
```
breeswish, line 232: This operation leaves something on the disk. Should it be PUT or POST instead?

pkg/apiserver/statement/statement.go
```diff
@@ -210,3 +228,118 @@ func (s *Service) getPlanDetailHandler(c *gin.Context) {
...
+// @Failure 401 {object} utils.APIError "Unauthorized failure"
+func (s *Service) downloadHandler(c *gin.Context) {
+	token := c.Query("token")
+	fielPath, err := utils.ParseJWTString("statements/download", token)
+	if err != nil {
+		utils.MakeInvalidRequestErrorFromError(c, err)
+		return
+	}
+
+	_, err = os.Stat(fielPath)
```
breeswish, line 333: typo: file

pkg/apiserver/statement/statement.go
```diff
@@ -210,3 +228,118 @@ func (s *Service) getPlanDetailHandler(c *gin.Context) {
...
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+
+	c.Writer.Header().Set("Content-type", "application/octet-stream")
+	c.Writer.Header().Set("Content-Disposition", "attachment; filename=\"statements.zip\"")
+	err = utils.StreamZipPack(c.Writer, []string{fielPath}, true)
```
breeswish, line 341: maybe we can just place zip on the disk when generating the CSV?

pkg/apiserver/statement/statement.go
baurine: Make sense, let me have a check.

ui/lib/apps/Statement/pages/List/index.tsx
```diff
@@ -47,8 +48,19 @@ export default function StatementsOverview() {
     allStmtTypes,
     loadingStatements,
     tableColumns,
+
+    getDownloadToken,
+    downloading,
   } = controller
 
+  async function exportCSV() {
+    const token = await getDownloadToken()
```
breeswish, line 57: I recommend this UI interact pattern:

1. Click export.
2. Show spinner / progress.
3. When successfully exported, display a Modal, with a button that links to the download link.
4. Clicking the button result in downloading the file.

This will bypass the browser popup interception when step 2 takes long time.

pkg/apiserver/statement/statement.go
```diff
@@ -210,3 +228,118 @@ func (s *Service) getPlanDetailHandler(c *gin.Context) {
...
+	timeLayout := "01021504"
+	beginTime := time.Unix(int64(req.BeginTime), 0).Format(timeLayout)
+	endTime := time.Unix(int64(req.EndTime), 0).Format(timeLayout)
+	tmpfile, err := ioutil.TempFile("", fmt.Sprintf("statements_%s_%s_*.csv", beginTime, endTime))
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+	defer tmpfile.Close()
+	csvwriter := csv.NewWriter(tmpfile)
```
breeswish, line 304: In v5.0 we have a new security requirement: anything on the disk should not leak user data. Statements contain user data, so that maybe you can follow this way: the data is stream-encrypted on the disk using a non-persisted key and stream-decrypted and served to user when later a download token is used.

ui/lib/apps/Statement/pages/List/index.tsx
```diff
@@ -106,6 +118,9 @@ export default function StatementsOverview() {
                 setQueryOptions({ ...queryOptions, searchText })
               }
             />
+            <Button type="primary" onClick={exportCSV} loading={downloading}>
```
breeswish, line 121: maybe we can put it in some extra menu in "..." or bread icon since it is not a frequently used feature. For the same reason, we can place Settings into this new menu.

pkg/apiserver/statement/statement.go
breeswish: Maybe deleting the file after request can be a good idea to minimize the temp file size..

ui/lib/apps/Statement/pages/List/index.tsx
baurine: Good idea, will update it.

pkg/apiserver/statement/statement.go
baurine: Yep, that's would be better.

pkg/apiserver/statement/statement.go
baurine: Got!

pkg/apiserver/statement/statement.go
```diff
@@ -291,24 +294,42 @@ func (s *Service) downloadTokenHandler(c *gin.Context) {
...
-	defer tmpfile.Close()
-	csvwriter := csv.NewWriter(tmpfile)
+	defer csvFile.Close()
+	// generate buf writer for tmpfile
+	csvBuf := bufio.NewWriter(csvFile)
+	// generate encrypted key
+	secretKey := cryptopasta.NewEncryptionKey()
+
+	rowBuf := bytes.NewBuffer(nil)
+	csvwriter := csv.NewWriter(rowBuf)
```
breeswish, line 313: This library implements an encrypted io.Reader / io.Writer, which could be a better choice. Maybe you can have a try: https://github.com/Xeoncross/go-aesctr-with-hmac

The encapsulation of the stream is file stream -> encrypted stream -> csv stream.

baurine: Ok, thank you! let me have a look. Since I found zip the encrypted file won't reduce the size much, so I think it doesn't need to zip file in local, right? but to still zip it when sending data back to the client after being decrypted.

unbyte: IIRC, we use gzip already. Is there any further benefit of packaging as a zip for transmission?

baurine: Make sense, let me focus on the encryption first.

baurine: It works fine! and the final flow is this:

image

baurine:
Encrypt and decrypt flow:

image

pkg/apiserver/statement/statement.go
baurine: Done.

pkg/apiserver/statement/statement.go
baurine: Done.

pkg/apiserver/statement/statement.go
baurine: Cancel explicit zip operation, reasons:

- zip encrypted data won't reduce the size
- the encrypted CSV file will be deleted after it's downloaded 
- we have already use gzip as middleware

pkg/apiserver/statement/statement.go
baurine: Done.

pkg/apiserver/statement/statement.go
baurine: Seems this lib can help us, let me try it.

pkg/apiserver/statement/statement.go
```diff
@@ -337,22 +339,41 @@ func (s *Service) downloadTokenHandler(c *gin.Context) {
...
+		_ = c.Error(err)
+		return
+	}
+	f, err := os.Open(filePath)
 	if err != nil {
 		_ = c.Error(err)
 		return
 	}
 
 	c.Writer.Header().Set("Content-type", "application/octet-stream")
```
unbyte, line 370: since transmit `.csv` directly now, `Content-Type` may can be `text/csv`

baurine: I just searched there isn't a `text/csv` kind Content-Type, and `application/octet-stream` is ok for any kind of content that is used to download.

baurine: Found it here: https://developer.mozilla.org/zh-CN/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types will change it.

ui/lib/apps/Statement/pages/List/index.tsx
baurine: After trying, it is a little annoying that needs 2 steps to download, so I change another way that wouldn't trigger browser  popup interception policy (have tested it):

```
function downloadByLink(url: string) {
  const downloadLink = document.createElement('a')
  downloadLink.href = url
  downloadLink.click()
}
```

ui/lib/apps/Statement/pages/List/index.tsx
baurine: Done.

baurine:
New effect:

(Mocked time delays)

unbyte:
improvement of UX:
```js
message.info({content, key, duration: 0}) // 0 means close manually
```
and close message before download
```js 
message.destory(key)
```

reference: https://ant.design/components/message-cn/

ui/lib/apps/Statement/pages/List/index.tsx
```diff
@@ -47,8 +65,44 @@ export default function StatementsOverview() {
...
+        break
+      case 'export':
+        exportCSV()
+        break
+    }
+  }
+
+  const dropdownMenu = (
+    <Menu onClick={menuItemClick}>
+      <Menu.Item key="settings">{t('statement.settings.title')}</Menu.Item>
```
breeswish, line 97: Recommend to add some icons to make it prettier :)

breeswish:
There is a spinner message: https://codesandbox.io/s/3yq1h

baurine:
> https://codesandbox.io/s/3yq1h

Cool!

baurine:
New effect:

ui/lib/apps/Statement/pages/List/index.tsx
baurine: Done.

ui/lib/apps/Statement/pages/List/index.tsx
baurine: It seems can be simplified by `window.location.href = url` as well.

baurine:
TODO: 
- [ ] export all fields instead of selected fields

breeswish:
@baurine Is this PR ready to merge?

baurine:
> @baurine Is this PR ready to merge?

Yep，I think so.

PR #793, opened 2020-11-11
Use string primitive type for some uint64 fields

### LINKED ISSUE #791: 64bit connection id may break slow query detail
## Bug Report

**What did you do?**

TiDB generates 64bit connection id since https://github.com/pingcap/tidb/pull/17649. In some cases, the slow query detail page cannot correctly handle this case due to JavaScript max int limitation.

Slow query detail work correctly.

**What did you see instead?**

Slow query detail may throw error.

breeswish:
This is a hack to fix #791 .


kennytm (commented):
lgtm

PR #794, opened 2020-11-11
ui: fix errors doesn't display

baurine:
Currently when we kill the TiDB, the dashboard doesn't display the error on the page when requesting API.

Reason: we store the error message in `err.message`, but sometimes when we access `err.message` later  we get undefined, still don't know the exact reason.

Fix: store the error message in `err.msg` field instead and it works fine.

After fixing:

WeCom20201111-154611@2x

### DIFF
--- a/ui/dashboardApp/layout/signin/index.tsx
+++ b/ui/dashboardApp/layout/signin/index.tsx
@@ -157,7 +157,7 @@ function useSignInSubmit(
       singleSpa.navigateToUrl(successRoute)
     } catch (e) {
       if (!e.handled) {
-        setError(t('signin.message.error', { msg: e.message }))
+        setError(t('signin.message.error', { msg: e.msg }))
         onFailure()
       }
     } finally {
--- a/ui/dashboardApp/layout/signin/index.tsx
+++ b/ui/dashboardApp/layout/signin/index.tsx
@@ -157,7 +157,7 @@ function useSignInSubmit(
       singleSpa.navigateToUrl(successRoute)
     } catch (e) {
       if (!e.handled) {
-        setError(t('signin.message.error', { msg: e.msg }))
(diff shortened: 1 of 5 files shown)


baurine:
@breeswish PTAL, thanks!

baurine:
It seems to break the CI, let me have a look.

ui/lib/client/index.tsx
```diff
@@ -59,7 +59,9 @@ function initAxios() {
       }
     }
     const content = i18next.t(errCode)
-    err.message = content
+    // err.message = content // sometimes we can't change err.message at all, still don't know the exact reason
```
breeswish, line 62: I encourage to dig deeper about this behaviour.

ui/dashboardApp/layout/signin/index.tsx
```diff
@@ -157,7 +157,7 @@ function useSignInSubmit(
       singleSpa.navigateToUrl(successRoute)
     } catch (e) {
       if (!e.handled) {
-        setError(t('signin.message.error', { msg: e.message }))
+        setError(t('signin.message.error', { msg: e.msg }))
```
breeswish, line 160: This only considers `e.msg`, will there be problems when `e.msg` is not available?

ui/lib/client/index.tsx
baurine: Ok!

baurine:
Get the root cause of this issue finally. It is because the `message` property of the `Error` object is a non-enumerable property, its value will miss when using spread syntax (I use it somewhere but not everywhere).

Ref: https://stackoverflow.com/a/59874261/2998877

baurine:
@breeswish Update it, PTAL, thanks!

breeswish (approved):
Nice dig!

PR #799, opened 2020-11-20
ui: fix the error message doesn't show correct

baurine:
Reason: 

`i18next.t()` method will truncate some contents and convert colons to dots if the content contains colons.

Example: 

```js
console.log(
  i18next.t('Test: error message with colons: colons will become dots') // error message with colons. colons will become dots
)
```

We also refine the logic a bit. If errCode hasn't a translation, we show the original detail error message instead.

### DIFF
--- a/ui/lib/client/index.tsx
+++ b/ui/lib/client/index.tsx
@@ -39,6 +39,7 @@ export enum ErrorStrategy {
   Default = 'default',
   Custom = 'custom',
 }
+const ERR_CODE_OTHER = 'error.api.other'
 
 function initAxios() {
   i18n.addTranslations(require.context('./translations/', false, /\.yaml$/))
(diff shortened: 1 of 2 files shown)


ui/lib/client/index.tsx
```diff
@@ -50,15 +51,18 @@ function initAxios() {
...
     if (err.message === 'Network Error') {
       errCode = 'error.network'
     } else {
-      errCode = response?.data?.code || 'error.api.other'
-      if (errCode === 'error.api.other') {
-        errCode = response?.data?.message || err.message
-      }
+      errCode = response?.data?.code
+    }
+    if (errCode !== ERR_CODE_OTHER && i18next.exists(errCode)) {
```
breeswish, line 60: Good find! Would you also like to change `TextWithInfo/index.tsx`, which can be simplified using `i18next.exists`?

Current code base:

```typescript
  const tooltip = t(`${transKey}_tooltip`, {
    defaultValue: '',
    fallbackLng: '_',
  })
```

baurine: Sure!

breeswish (approved):

breeswish:
Oh, sorry. I just realize that `i18n.exists` might not be exactly the same as the current implementation. Could you revert my proposal?

Consider that we have these translations:

EN: err.code.xxx = "Xxx description".
CN: (not exist).

When we receive "err.code.xxx" and current language is CN, its translation exists, which is the EN fallback translation.

However, in `TextWithInfo` case, it's not the same:

EN: abc = "Abc", abc_tooltip = "Abc Tooltip"
CN: abc = "Abc 内容", abc_tooltip not exist.

In this case, we would like to show only "Abc 内容" when language is CN, instead of showing "Abc 内容" with an english "Abc Tooltip" tooltip.

baurine:
Got it, will revert it

breeswish (approved):

PR #802, opened 2020-11-20
ui: copy original content instead of formatted content for CopyLink

### LINKED ISSUE #798: Digester produces invalid sql for selecting global variables
### 1. Minimal reproduce step (Required)

Linked [issue](https://github.com/pingcap/tidb/issues/18224).

The digester produces an invalid SQL for the following statement:

```sql
SELECT @@global.tidb_stmt_summary_refresh_interval AS value
```

This statement is [generated by the tidb-dashboard](https://github.com/pingcap-incubator/tidb-dashboard/blob/df682d9eb42a34bf53cd9aec113cc8bf4dbe8865/pkg/apiserver/statement/queries.go#L34-L50).

I have another digester bug in https://github.com/pingcap/tidb/issues/14241 - it is annoying when the syntax generated back is invalid.

### 2. What did you expect to see? (Required)

```sql
SELECT @@global.tidb_stmt_summary_refresh_interval AS value
```

### 3. What did you see instead (Required)

There is a space between the two @ symbols:

```sql
SELECT @ @global.tidb_stmt_summary_refresh_interval AS value
```

### 4. Affected version (Required)

```
mysql> select tidb_version()\G
*************************** 1. row ***************************
tidb_version(): Release Version: v4.0.0-beta.2-684-gf590d74e1
Edition: Community
Git Commit Hash: f590d74e158cf4e33555bfeb8f88971af5956697
Git Branch: master
UTC Build Time: 2020-06-27 02:29:48
GoVersion: go1.13.8
Race Enabled: false
TiKV Min Version: v3.0.0-60965b006877ca7234adaced7890d7b029ed1306
Check Table Before Drop: false
1 row in set (0.00 sec)
```

baurine:
fix #798 in step 1

### DIFF
--- a/ui/lib/apps/SlowQuery/pages/Detail/index.tsx
+++ b/ui/lib/apps/SlowQuery/pages/Detail/index.tsx
@@ -91,7 +91,14 @@ function DetailPage() {
                         expanded={detailExpand.query}
                         onClick={toggleQuery}
                       />
-                      <CopyLink data={formatSql(data.query!)} />
+                      <CopyLink
+                        textTransKey="copyFormatted"
+                        data={formatSql(data.query!)}
+                      />
+                      <CopyLink
+                        textTransKey="copyOriginal"
+                        data={data.query!}
+                      />
                     </Space>
                   }
                 >
@@ -117,7 +124,14 @@ function DetailPage() {
                               expanded={detailExpand.prev_query}
                               onClick={togglePrevQuery}
                             />
-                            <CopyLink data={formatSql(data.prev_stmt!)} />
+                            <CopyLink
+                              textTransKey="copyFormatted"
+                              data={formatSql(data.prev_stmt!)}
+                            />
+                            <CopyLink
+                              textTransKey="copyOriginal"
+                              data={data.prev_stmt!}
+                            />
                           </Space>
                         }
                       >
--- a/ui/lib/apps/SlowQuery/pages/Detail/index.tsx
+++ b/ui/lib/apps/SlowQuery/pages/Detail/index.tsx
@@ -8,7 +8,6 @@ import { useLocalStorageState } from '@umijs/hooks'
 import client from '@lib/client'
 import { useClientRequest } from '@lib/utils/useClientRequest'
 import { buildQueryFn, parseQueryFn } from '@lib/utils/query'
-import formatSql from '@lib/utils/formatSql'
 import {
   AnimatedSkeleton,
   CardTabs,
@@ -91,7 +90,7 @@ function DetailPage() {
                         expanded={detailExpand.query}
                         onClick={toggleQuery}
                       />
-                      <CopyLink data={formatSql(data.query!)} />
+                      <CopyLink data={data.query!} />
                     </Space>
                   }
                 >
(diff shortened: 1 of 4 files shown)


breeswish (commented):
I think we should just fix the incorrect format, instead of copying an unformatted SQL.. Since current copy problem only exists in very rare cases, while copying a formatted SQL is widely used.

baurine:
According to the comment in https://github.com/pingcap/tidb-dashboard/issues/798#issuecomment-731554967 , we can supply "Copy Original" and "Copy Formatted" buttons both.

The effects:

WeCom20201123-121229@2x

WeCom20201123-121300@2x

ui/lib/apps/SlowQuery/pages/Detail/index.tsx
```diff
@@ -116,7 +124,14 @@ function DetailPage() {
                               expanded={detailExpand.prev_query}
                               onClick={togglePrevQuery}
                             />
-                            <CopyLink data={data.prev_stmt!} />
+                            <CopyLink
+                              textTransKey="copyFormatted"
```
breeswish, line 128: maybe better to make it type safe, i.e. `<CopyLink displayVariant={undefined|"default"|"original_sql"|"formatted_sql"}>`

baurine: Make sense!

ui/lib/components/CopyLink/index.tsx
```diff
@@ -43,15 +58,20 @@ function CopyLink({ data }: ICopyLinkProps) {
...
   return (
     <span>
-      {!showCopied && (
-        <CopyToClipboard text={data} onCopy={handleCopy}>
-          <a>
-            {t('component.copyLink.text')} <CopyOutlined />
-          </a>
-        </CopyToClipboard>
+      {!showCopied && displayVariant === 'original_sql' && (
+        <Tooltip title={t('component.copyLink.tooltip')}>{copyBtn}</Tooltip>
```
breeswish, line 72: Maybe here is a mistake. The tooltip only exists specifically for "original_sql" but not "formatted_sql". Maybe you can just remove this tooltip.

ui/lib/components/CopyLink/index.tsx
```diff
@@ -1,24 +1,39 @@
...
     success: 'Copied',
+    tooltip: 'Copy by this button if you need to run it in SQL client',
   },
   zh: {
-    text: '复制',
+    copy: '复制',
+    copyOriginal: '复制原始 SQL',
+    copyFormatted: '复制格式化 SQL',
     success: '已复制',
+    tooltip: '如果你需要在 SQL 客户端执行此 SQL 语句，请使用这个按钮进行复制',
```
breeswish, line 36: This tooltip can cause misleading because it is describing a bug.

baurine: 😂

ui/lib/components/CopyLink/index.tsx
baurine: Done!

breeswish (approved):

PR #805, opened 2020-11-23
ui: add MySqlFormatter to customize the sql formatter

### LINKED ISSUE #798: Digester produces invalid sql for selecting global variables

baurine:
fix #798  in step 2

Add our customized SQL formatter.

Instead of contributing to the upstream, I choose to extend it in ourself repo temporarily (or permanently).

I planned to implement the MySqlFormatter by inheriting the StardardSqlFormatter and then change its tokenizer slightly but found it too hard. Also tried to hook the constructor of Formatter, but failed. The original design is hard to extend.

Not sure whether you guys can accept this solution.

TODO: 

- [ ] add some test

Effect:

image


ui/lib/utils/formatSql.ts
```diff
@@ -1,5 +1,7 @@
-import sqlFormatter from 'sql-formatter-plus-plus'
+import MySqlFormatter from './MySqlFormatter/MySqlFormatter'
```
breeswish, line 1: Maybe better to arrange formatSql and MySqlFormatter in the same directory since they are deeply linked. Also maybe we can name it TiDBSQLFormatter :)

breeswish (commented):
The rest LGTM

ui/lib/utils/formatSql.ts
baurine: Ok, nice naming!

ui/lib/utils/TiDBSQLFormatter/formatSql.ts
```diff
@@ -0,0 +1,7 @@
+import TiDBSQLFormatter from './TiDBSQLFormatter'
```
breeswish, line 1: Maybe the directory and file can be `utils/sqlFormatter/index.ts`

Then the usage can be as simple as: import formatSql from '@lib/utils/sqlFormatter`.

baurine: Yep, that's much better.

PR #806, opened 2020-11-23
*: fix query statement detail error cause by round

crazycs520:

When query statement detail, dashboard uses to use `ROUND` function to query, but it may cause the flowing error:

image

This is caused by the `ROUND` function result is `float64`, not `int`, here is an example.

```sql
test> create table t (a int, b double);
Query OK, 0 rows affected
Time: 0.044s
test> insert into t values (62695, 2240325.4838709678),(6, 0.0);
Query OK, 2 rows affected
Time: 0.042s
test> select round(sum(a*b)/sum(a)) as avg from t;
+-----------+
| avg       |
+-----------+
| 2240111.0 |
+-----------+
```

This PR use `CAST( ... AS SIGNED)` function instead of `ROUND` when query statement detail.


crazycs520:
@breeswish @baurine PTAL

breeswish (approved):

PR #816, opened 2020-11-26
ui: Open statement and slow log in new tab

breeswish:
This is a workaround for https://github.com/pingcap/tidb-dashboard/issues/641 to improve UX without changing too many codebase.

However most ideally we should implement page memorization and bring back "Back" again.

Also fixes https://github.com/pingcap/tidb-dashboard/issues/796


baurine (approved):
LGTM

baurine:
Maybe later we can try to use use-swr to help cache the API response.

unbyte:
how about using react-keep-alive?

breeswish:
> how about using react-activation?

Feel free to dig it!

PR #817, opened 2020-11-26
slow_log: Improve descriptions

breeswish:
Also, some descriptions are long so that a tooltip is added.


crazycs520 (approved):
LGTM

PR #822, opened 2020-12-08
ui: keep table components from being destroyed when viewing details

### LINKED ISSUE #641: Slow log search result should be cached
## Bug Report

**What did you do?**

Search slow log, and click one to see the details. After moving back, it starts search again.

Display the result immediately

**What did you see instead?**

Nothing shows up, and the refresh button begin spinning.

I'm using v4.0.0 GA, not sure what version is Dashboard.

unbyte:
preview:

known issue: 
- meaningless transition animation

close #641

### DIFF
--- a/ui/lib/apps/SlowQuery/index.tsx
+++ b/ui/lib/apps/SlowQuery/index.tsx
@@ -1,16 +1,18 @@
 import React from 'react'
 import { Root } from '@lib/components'
-import { HashRouter as Router, Route, Routes } from 'react-router-dom'
-import { List, Detail } from './pages'
+import { HashRouter as Router } from 'react-router-dom'
+import { Detail, List } from './pages'
+import ListAndDetail from '@lib/components/ListAndDetail'
 
 export default function () {
   return (
     <Root>
       <Router>
-        <Routes>
-          <Route path="/slow_query" element={<List />} />
-          <Route path="/slow_query/detail" element={<Detail />} />
-        </Routes>
+        <ListAndDetail
+          DetailComponent={Detail}
+          ListComponent={List}
+          detailPathMatcher={(path) => path.startsWith('/slow_query/detail')}
+        />
       </Router>
     </Root>
   )
--- a/ui/lib/components/ListAndDetail/index.tsx
+++ b/ui/lib/components/ListAndDetail/index.tsx
@@ -3,7 +3,7 @@ import React from 'react'
 
 export interface ListAndDetailProps {
   ListComponent: React.FunctionComponent
-  DetailComponent: React.FunctionComponent<{ style: React.CSSProperties }>
+  DetailComponent: React.FunctionComponent
   detailPathMatcher: (path: string) => boolean
 }
 
@@ -13,7 +13,6 @@ export default function ({
   detailPathMatcher,
 }: ListAndDetailProps) {
   const location = useLocation()
-  console.log('rendered')
   return (
     <div
       style={{
@@ -22,7 +21,7 @@ export default function ({
     >
       <ListComponent />
       {detailPathMatcher(location.pathname) && (
-        <DetailComponent
+        <div
           style={{
             zIndex: 99,
             position: 'absolute',
@@ -32,7 +31,9 @@ export default function ({
             width: '100%',
             transition: 'none',
           }}
-        />
+        >
+          <DetailComponent />
+        </div>
       )}
     </div>
   )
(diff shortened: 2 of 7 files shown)


ui/lib/components/ListAndDetail/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+    >
+      <ListComponent />
+      {detailPathMatcher(location.pathname) && (
+        <DetailComponent
+          style={{
+            zIndex: 99,
+            position: 'absolute',
+            top: 0,
+            left: 0,
+            background: '#fff',
```
unbyte, line 31: does the background color need to be determined by theme?

baurine:
I will have a look later today

ui/lib/components/ListAndDetail/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+  detailPathMatcher: (path: string) => boolean
+}
+
+export default function ({
+  ListComponent,
+  DetailComponent,
+  detailPathMatcher,
+}: ListAndDetailProps) {
+  const location = useLocation()
+  console.log('rendered')
```
baurine, line 16: It can be removed.

ui/lib/components/ListAndDetail/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+  return (
+    <div
+      style={{
+        position: 'relative',
+      }}
+    >
+      <ListComponent />
+      {detailPathMatcher(location.pathname) && (
+        <DetailComponent
```
baurine, line 25: How about use an extra `div` to wrap the `<DetailComponent/>`? for example:

```
<div className={xxx}>
  <DetailComponent/>
</div>
```

So we don't need to modify the DetailPage.

baurine:
Although show List and Detail inside one page is a kind of solution, I am more feel it as tricky. How about importing a small store (only stores the list data) for each of these 2 apps? Or, cache the list data in the useStatementTableController.ts and useSlowQueryTableController.ts. I prefer the latter.

For example:

```js
// useStatementTableController.ts

const cachedListData = {
  key: '',  // beginTime_endTime_fields_schemas_stmtTypes_searchText_refreshTimes
  listData: [] // StatementModel[]
}

export default function useStatementTableController() {
  //...

  useEffect(() => {
    async function queryStatementList() {
      // check cache
      const curKey = `${validTimeRange.begin_time}_${validTimeRange.end_time!}..._${refreshTimes}`
      if (curKey === cachedListData.key) {
        return cachedListData.listData
      }
     
      //...
    }

    queryStatementList()
  }, [queryOptions, allTimeRanges, validTimeRange, selectedFields])

```

How do you think about it?

unbyte:
> How about importing a small store (only stores the list data) for each of these 2 apps

Before this PR, I have made many attempts, like implementing component caching like https://github.com/CJY0208/react-activation, or implementing a simple `useCache` hook, judge whether there is a cached state when the component is mounted and use the cache state to render the component.

One disadvantage of the above is the introduction of a relatively large amount of code. The first one is self-evident. The second one needs to maintain a state machine of route inside cache store and modify the logic in the controllers for using cached state and caching state.

In fact, the solution in this PR is simple enough and has no performance issues, though the transition animation makes the appearance of the details page a little less smooth.

ui/lib/components/ListAndDetail/index.tsx
unbyte: Just worried that the introduction of a new div will lead to the failure of some styles related to the div level. And it seems there is no such problem at present,

ui/lib/components/ListAndDetail/index.tsx
unbyte: my fault😂

ui/lib/apps/SlowQuery/index.tsx
```diff
@@ -1,16 +1,18 @@
...
     <Root>
       <Router>
-        <Routes>
-          <Route path="/slow_query" element={<List />} />
-          <Route path="/slow_query/detail" element={<Detail />} />
-        </Routes>
+        <ListAndDetail
+          DetailComponent={Detail}
+          ListComponent={List}
+          detailPathMatcher={(path) => path.startsWith('/slow_query/detail')}
```
breeswish, line 14: This kind of manual path match is too hacky

breeswish:
IMO, simple is good but should not sacrifice maintainability, i.e. it should not be a simple "hack". It should be a simple "solution".

PR #827, opened 2020-12-17
fix issue of query topn statement get error

crazycs520:

image

It is caused by https://github.com/pingcap/tidb/pull/21286, which is compatible with MySQL 5.7. 

So I just change the query SQL.


crazycs520:
@breeswish PTAL

breeswish (approved):

PR #830, opened 2020-12-24
show indents for slow query detail time more elegantly

baurine:
Before:

WeCom20201224-161739@2x

After:

WeCom20201224-161807@2x

image

### DIFF
--- a/ui/lib/utils/tableColumns.tsx
+++ b/ui/lib/utils/tableColumns.tsx
@@ -60,12 +60,9 @@ function fieldsKeyColumn(transKeyPrefix: string): IColumn {
     minWidth: 150,
     maxWidth: 250,
     onRender: (rec) => {
-      const indent = Array(rec.indentLevel || 0)
-        .fill('\t')
-        .join('')
       return (
         <>
-          <span style={{ whiteSpace: 'pre' }}>{indent}</span>
+          <span style={{ paddingLeft: (rec.indentLevel || 0) * 24 }}></span>
           {rec.keyDisplay ?? (
             <TransText transKey={`${transKeyPrefix}${rec.key}`} />
           )}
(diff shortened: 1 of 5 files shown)


baurine:
Another more elegant way (I think):

1. remove all indents in the i18n files
2. add indentLevel for each row
3. show extra indent

still in trying

```js
    {
      key: 'parse_time',
      value: data.parse_time! * 10e8,
      indentLevel: 1,
    }
    
    
function fieldsKeyColumn(transKeyPrefix: string): IColumn {
  return {
    name: commonColumnName(TRANS_KEY_PREFIX, 'name'),
    key: 'key',
    minWidth: 150,
    maxWidth: 250,
    onRender: (rec) => {
      if (rec.keyDisplay) {
        return rec.keyDisplay
      }
      let indent = Array((rec.indentLevel || 0)*4)
        .fill(' ')
        .join('')
      return (
        <div>
          <span>{indent}</span>
          <TransText transKey={`${transKeyPrefix}${rec.key}`} />
        </div>
      )
    },
  }
}
```

breeswish (commented):
I prefer the indentLevel solution, `trim()` is too hacky

ui/lib/utils/tableColumns.tsx
```diff
@@ -60,10 +60,17 @@ function fieldsKeyColumn(transKeyPrefix: string): IColumn {
...
-      if (rec.keyDisplay) {
-        return rec.keyDisplay
-      }
-      return <TransText transKey={`${transKeyPrefix}${rec.key}`} />
+      const indent = Array(rec.indentLevel || 0)
+        .fill('\t')
+        .join('')
+      return (
+        <>
+          <span style={{ whiteSpace: 'pre' }}>{indent}</span>
```
breeswish, line 68: maybe better to be `<span style={{ width: indentLevel * xxx }} /> `

baurine: Good idea! will update it soon.

baurine: Done, use padding-left instead of width to fill the space, width doesn't work for span unless we set the display to inline-block.

breeswish (approved):

Issue #833, opened 2020-12-30
Differ TiKV and TiFlash nodes in store location topo


baurine:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

![image](https://user-images.githubusercontent.com/1284531/103338359-1cf2a000-4ab9-11eb-8521-d369652b5d7a.png)

In the above screenshot, we can't know which node is tikv and which node is tiflash, we can differ them by different colors or text.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

Only tiflash nodes have "engine: tiflash" label.

PR #834, opened 2020-12-31
Diff tikv tiflash

baurine:
close issue #833 

What did:

- Differ TiFlash and TiKV nodes by different text and color (maybe we can choose a better color)
- Make reload really work
- Change tooltip to the text which always shows

Effect:

WeCom20201231-161953@2x

### DIFF
--- a/ui/lib/apps/ClusterInfo/components/StoreLocation.tsx
+++ b/ui/lib/apps/ClusterInfo/components/StoreLocation.tsx
@@ -46,7 +46,12 @@ function buildTreeData(data: TopologyStoreLocation | undefined): TreeNode {
 }
 
 export default function StoreLocation() {
-  const { data, isLoading, error } = useClientRequest((reqConfig) =>
+  const {
+    data,
+    isLoading,
+    error,
+    sendRequest,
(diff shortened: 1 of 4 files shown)


baurine:
@breeswish @unbyte PTAL, thanks!

unbyte:
color differences between nodes doesn't seem to be obvious. how about changing edge color too?

breeswish:
> color differences between nodes doesn't seem to be obvious. how about changing edge color too?

+1. And maybe color can be changed as well, for example, a color which has more contrast: 

[Magenta-4](https://ant.design/docs/spec/colors) 

image

or [Volcano-4](https://ant.design/docs/spec/colors)

image

baurine:
How about this?

image

breeswish (approved):
Cool!

Just a guess. Maybe you can try with cyan[5], magenta[4], orange[4]. The color display may be better. Current TiKV / TiFlash colors seems to be a bit dark.

baurine:
Update colors by:

Line/Label: #ddd / grey[1]
TiKV: cyan[5]
TiFlash: magenta[4]

image

PR #835, opened 2021-01-04
optimize slow-query list performance

crazycs520:
Signed-off-by: crazycs520 <crazycs520@gmail.com>

convert slow query list SQL from:

```sql
SELECT Digest, ...... ORDER BY timestamp DESC LIMIT 100
```

to 

```sql
SELECT Digest, ...... ORDER BY Time DESC LIMIT 100
```

The related query plan change from 

```sql
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| tidb_decode_plan('sA3oMAkzXzcJMAkxMDAJaW5mb3JtYXRpb25fc2NoZW1hLmNsdXN0ZXJfc2xvd19xdWVyeS5kaWdlc3QsIGmSLgCIY29ubl9pZCwgcGx1cyhjYXN0KHVuaXhfdGltZXN0YW1wKGmSSAABMIApLCBkb3VibGUgQklOQVJZKSwgMCktPkNvbHVtbiM3MSyalAAQcXVlcnmyLQAFr54yABhtZW1fbWF4JVMBt9A6NDAxLjRtcy                                                                                                                                                                                                                                                                                                                                                        |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|     id                              task         estRows    operator info                                                                                                                                                                                                                                                                                                                                                                                 actRows    execution info                                                                                                                     memory     disk |
|     Projection_7                    root         100        information_schema.cluster_slow_query.digest, information_schema.cluster_slow_query.conn_id, plus(cast(unix_timestamp(information_schema.cluster_slow_query.time), double BINARY), 0)->Column#71, information_schema.cluster_slow_query.query, information_schema.cluster_slow_query.query_time, information_schema.cluster_slow_query.mem_max                                                100        time:401.4ms, loops:2, Concurrency:OFF                                                                                             19.6 KB    N/A  |
[...]

PR #838, opened 2021-01-05
fix slow query bug when order by other fields

crazycs520:
1. revert #835
2. Only convert to `ORDER BY TIME` when the original is `ORDER BY timestamp`

### DIFF
--- a/pkg/apiserver/slowquery/queries.go
+++ b/pkg/apiserver/slowquery/queries.go
@@ -234,6 +234,7 @@ func QuerySlowLogList(db *gorm.DB, req *GetListRequest) ([]SlowQuery, error) {
 		order[0] = req.OrderBy
 	}
 	if order[0] == "timestamp" {
+		// Order by column instead of expression, see related optimization in TiDB: https://github.com/pingcap/tidb/pull/20750
 		order[0] = "Time"
 	}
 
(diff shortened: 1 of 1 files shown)


crazycs520:
@breeswish PTAL

pkg/apiserver/slowquery/queries.go
```diff
@@ -219,10 +219,28 @@ func QuerySlowLogList(db *gorm.DB, req *GetListRequest) ([]SlowQuery, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	// to handle the special case: timestamp
+	// if req.OrderBy is "timestamp", then the order is "(unix_timestamp(Time) + 0E0) AS timestamp"
+	if strings.Contains(order[0], " AS ") {
+		order[0] = req.OrderBy
+	}
+	if order[0] == "timestamp" {
+		order[0] = "Time"
```
breeswish, line 237: Please add a comment for why do we need to do this.

breeswish (approved):
The rest LGTM. @baurine PTAL

pkg/apiserver/slowquery/queries.go
baurine: Same question, when does this condition happen?

baurine (approved):
LGTM

pkg/apiserver/slowquery/queries.go
crazycs520: done.

PR #839, opened 2021-01-05
Refine page change statistics

baurine:
When debugging, found some small issues:

1. There are several warnings about React usage in the console.
2. When login or logout, the single-spa emits the same route change event twice or 3 times, this will make the page change statistics not precise.

WeCom20210105-141946@2x

WeCom20210105-142008@2x

WeCom20210105-142341@2x

WeCom20210105-142137@2x


ui/lib/apps/ClusterInfo/pages/List.tsx
```diff
@@ -56,14 +56,20 @@ export default function ListPage() {
     },
   ]
 
+  function clickTab(tabKey) {
+    if (tabKey === 'instance') {
```
breeswish, line 60: This is too hacky, how about using a redirection (redirect `/ `to `/instance`)?

baurine: Yep, I do think so. It is redirection before.

```
<Route	       
            path="/cluster_info"
            element={<Navigate to="/cluster_info/instance" replace />}
```

I do this change because I don't want the page change event to happen when we are on the "/cluster_info/instance" page and clicking the "Cluster Info" link on the left side menu, to keep the same behavior as other apps.

Let me think it about more to find a more elegant solution.

breeswish: I think this behaviour is fine, since this page change event only generates a dummy "/cluster_info" telemetry information, which we can just ignored, considering that a `/cluster_info/instance` should be always followed and information is not lost?

baurine: Ok, I reverted this part code.

breeswish (approved):

PR #828, opened 2020-12-17
use cache for slow query and statements

baurine:
- [x] use cache for slow query
- [x] use cache for statements

another solution for PR #822

PR #843, opened 2021-01-06
remember the coarse scrolled position for statements and slow queries list

baurine:
Related PR: #828 

What did:

- Remember the last coarse scrolled position (in fact the clicked row) in the statements table, restore the position when coming back from the detail page

Effect:

image

How it implemented:

Because I didn't find a good way to remember and recover the exactly scrolled position for the statements table, I found the fluent UI DetailsList has a [initialFocusedIndex](https://developer.microsoft.com/en-us/fluentui#/controls/web/detailslist) props that can make the list scroll to the target row when initializing. So we can remember the row index when clicking it. (Later found the initialFocusedIndex doesn't work well when the index is large, it needs the item is already rendered before scrolling to. Replace it with `scrollToIndex()` method and works well.)

When coming back from the detail page, the clicked row may not be scrolled to the exact position when leave, it may scroll to the top, middle, or bottom of the list, but we can see it on the screen. 

To make the user can get the last clicked row as quickly as possible, we can add a highlight effect for it, for example, an extra border.

Not sure whether this solution is acceptable, if it does, I will apply it for the slow query table.

### DIFF
--- a/ui/lib/apps/SlowQuery/components/SlowQueriesTable.tsx
+++ b/ui/lib/apps/SlowQuery/components/SlowQueriesTable.tsx
@@ -19,11 +19,15 @@ function SlowQueriesTable({ controller, ...restProps }: Props) {
     changeOrder,
     errors,
     visibleColumnKeys,
+
+    saveClickedItemIndex,
+    getClickedItemIndex,
   } = controller
 
   const navigate = useNavigate()
   const handleRowClick = usePersistFn(
-    (rec, _idx, ev: React.MouseEvent<HTMLElement>) => {
+    (rec, idx, ev: React.MouseEvent<HTMLElement>) => {
+      saveClickedItemIndex(idx)
       const qs = DetailPage.buildQuery({
         digest: rec.digest,
         connectId: rec.connection_id,
@@ -47,6 +51,7 @@ function SlowQueriesTable({ controller, ...restProps }: Props) {
       errors={errors}
       visibleColumnKeys={visibleColumnKeys}
       onRowClicked={handleRowClick}
+      clickedRowIndex={getClickedItemIndex()}
       getKey={getKey}
     />
   )
--- a/ui/lib/apps/SlowQuery/index.tsx
+++ b/ui/lib/apps/SlowQuery/index.tsx
@@ -6,7 +6,7 @@ import useCache, { CacheContext } from '@lib/utils/useCache'
 import { List, Detail } from './pages'
 
 export default function () {
-  const slowQueryCacheMgr = useCache()
+  const slowQueryCacheMgr = useCache(2)
 
   return (
     <Root>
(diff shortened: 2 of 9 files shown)


baurine:
@breeswish @unbyte PTAL, thanks!

breeswish:
Can we reuse the focus style or implementation built into the DetailsList?

image

baurine:
> Can we reuse the focus style or implementation built into the DetailsList?
> 
> image

Did notice that before, let me try it.

baurine:
hi @breeswish , where does the above example come from? I didn't find it in its documents.

breeswish:
@baurine Access https://developer.microsoft.com/en-us/fluentui#/controls/web/detailslist and then use Tab key

baurine:
thanks @breeswish , I tried and it seems the effect is only can be triggered by the tab key, I didn't find a way to trigger it in other ways...  and when you click anywhere, the effect will disappear quickly.

breeswish:
@baurine Because it's the focus ring. I mean you can use the focus ring to display the "last active item" status, so that when user clicks outside, it disappears as normal

unbyte:
@baurine just memorize the last clicked item and add a focus class on it when back to the list page

and I wonder if we can reuse the default outline style

baurine:
> @baurine just memorize the last clicked item and add a focus class on it when back to the list page
> 
> and I wonder if we can reuse the default outline style

Yep, that would be better if we can reuse the default outline style, I just don't know how to figure it out...

breeswish:
According to https://github.com/microsoft/fluentui/issues/1112, you may try:

1. Use `scrollToIndex` to bypass the visualized issue: https://github.com/microsoft/fluentui/issues/8722
2. Give focus to the details list to show the focus ring given by `initialFocusedIndex`

baurine:
> According to [microsoft/fluentui#1112](https://github.com/microsoft/fluentui/issues/1112), you may try:
> 
> 1. Use `scrollToIndex` to bypass the visualized issue: [microsoft/fluentui#8722](https://github.com/microsoft/fluentui/issues/8722)
> 2. Give focus to the details list to show the focus ring given by `initialFocusedIndex`

Let me have a try, thank you!

baurine:
> According to [microsoft/fluentui#1112](https://github.com/microsoft/fluentui/issues/1112), you may try:
> 
> 1. Use `scrollToIndex` to bypass the visualized issue: [microsoft/fluentui#8722](https://github.com/microsoft/fluentui/issues/8722)
> 2. Give focus to the details list to show the focus ring given by `initialFocusedIndex`

So pity it doesn't work... the `initialFocusedIndex` won't give the target item focus ring.

baurine:
btw, I reduced the border width to 1px, now it looks like this:

image

breeswish (approved):

PR #850, opened 2021-02-05
Update github.com/oleiade/reflections to v1.0.1 to avoid checksum mismatch

dveeden:
See https://github.com/oleiade/reflections/issues/14 for details.


baurine:
hi @dveeden, thanks for your contribution, the CI doesn't pass, can you check it.

breeswish:
Thank you for your contribution! The CI failed due to some linter problems. I guess it is not related to your change and I will take a look.

breeswish (approved):

breeswish:
I managed to fix the CI build error after cleaning up the `go.sum`.

breeswish:
Note: for people that facing build failures due to reflection commit change, you are likely to use GOPPROXY=direct. In order to migrate the issue, you can also try with something like:

```go
GOPROXY=https://proxy.golang.org make
```

TisonKun:
> Note: for people that facing build failures due to reflection commit change, you are likely to use GOPPROXY=direct. In order to migrate the issue, you can also try with something like:
> 
> ```shell
> GOPROXY=https://proxy.golang.org make
> ```

Thanks. It should have saved hours of work.

Issue #686, opened 2020-07-14
Visualize Regions or Replicas


breeswish:
## Feature Request

**Is your feature request related to a problem? Please describe:**

We commonly face these issues:

- How many miss replicas are there? Do they distribute on a specific host?
- Is the leader distribution of a table balanced for each store?
- Is the read / write flow of a table balanced for each store?
- What's the proportion of the hot / cold data for each table / database?
- What is the distribution of replica sizes, or keys in replica? (so that one can decide a better region merge configuration)

**Describe the feature you'd like:**

Issues listed above can be all answered by the region info accessible via `region` command in pd-ctl. However pd-ctl only produce raw data. People usually need to rely on other tools (which uses region info data under the hood), like writing a script to analyze, or using SQL command to query distributions.

We can provide a visualization / analytics page in TiDB Dashboard for region info, so that it can be easy to know everything about region info. The detail functionality of the page is listed in this proposal: https://docs.google.com/document/d/1moQVhvIgqu_FWuv_UMB76AM5tETJkyeWnzI04wyevhk/edit#heading=h.8xipn1t7e2cb 

The page can be implemented by using the awesome visualization library [SandDance](https://github.com/microsoft/SandDance).

This feature is a subset of https://github.com/pingcap-incubator/tidb-dashboard/issues/548 

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

IANTHEREAL:
Description：Visualize Regions or Replicas
Category：Feature、Usability
Value： Simplify the debug method of replica/region and provide more fine-grained observability of TiDB
Value Score: 3

baurine:
The link about how to use [pd-ctl](https://docs.pingcap.com/zh/tidb/stable/pd-control#pd-control-%E4%BD%BF%E7%94%A8%E8%AF%B4%E6%98%8E)

PR #851, opened 2021-02-06
feat: data visualization

unbyte:
related: #686 

Based on the needs of TiDB-Dashboard, customized SandDance and introduced it as external tool for TiDB-Dashboard.

Known Issue:
- text data in csv format source would be incorrectly parsed as date data because of the bug of V8 if it's of format `alias:port`
- precision of legend
- data displayed in the exported snapshot html is incorrect (the data itself is correct)

See the video (in Chinese) for more information.

https://user-images.githubusercontent.com/5772358/107119671-1fb89000-68c4-11eb-82d3-df59a3c6b48b.mp4


breeswish:
Very awesome job!

I have several thoughts:

1. Could we extract and modify the user interface part of SandDance? The goal is to make sure that people without enough knowledge can start to use it pretty quickly. For example, we can provide rich presets so that for common tasks user only need to select ones, instead of selecting x axis, y axis, color, size, graph kind, etc. Also, many other features in the UI is not necessary in TiDB Dashboard, for example, we may not need raw data view, or snapshot, etc.
   > According to my previous research Sanddance contains a main canvas (the right part) that can be controlled from the source code, and a UI to control it. So that maybe we can do something similar, or just reuse its UI control code.


2. About the csv issue, is it possible to parse the csv by our own to bypass the issue? Also maybe we need a JSON import as well, considering that what replica information we can get from PD (pd-ctl) is in JSON format instead of CSV. No one will have a region information in CSV.

unbyte:
@breeswish 

> we can provide rich presets so that for common tasks user only need to select ones, instead of selecting x axis, y axis, color, size, graph kind, etc. 

ofc, in fact I'm designing the presets selection component

> According to my previous research Sanddance contains a main canvas (the right part) that can be controlled from the source code, and a UI to control it. So that maybe we can do something similar, or just reuse its UI control code.

until a week ago, I was working on using the antd components and SandDanceReact to partially implement features from SandDance-Explorer, and at last I realized that under the premise of keeping certain features and large enough data display area in dashboard, the interaction of the page is a mess. So I decide to rewrite to make this feature independent of the dashboard, so I used SandDance-Explorer directly and customized SandDance-App. 

_If don't need add many features to what SandDance provides, using it directly is nicer._

> Also, many other features in the UI is not necessary in TiDB Dashboard, for example, we may not need raw data view, or snapshot, etc.

since a high degree of customization is provided, it is not a bad thing to have snapshots feature. It can save the results (or intermediate products) of data visualization, which is particularly useful (eg, some people need to obtain different charts of the same regions data and save them in a markdown file as a report)

but I agree that the UI should be modified to make the primary and secondary features clear, so I will try to modify the explorer interface.

> About the csv issue, is it possible to parse the csv by our own to bypass the issue? 
no, raw text data is read and parsed and preproduced by vega, and vega infers types of columns just after raw text data is parsed to an array.


> Also maybe we need a JSON import as well, considering that what replica information we can get from PD (pd-ctl) is in JSON format instead of CSV. No one will have a region information in CSV.

considering that the actual amount of regions data can be 10,000+, the number of replications should be higher than the number of regions, so using JSON format will produce huge data redundancy. That's why I decide to use csv by default. 
> data of JSON format can be got at /api/topology/regions?type=regions&format=json

unbyte:
goals:
- [ ] support presets selection and custom presets
- [ ] modify UI

ngaut:
This is amazing, very helpful. Good job.

breeswish:
@unbyte 

> considering that the actual amount of regions data can be 10,000+, the number of replications should be higher than the number of regions, so using JSON format will produce huge data redundancy. That's why I decide to use csv by default.
and data of JSON format can be got at /api/topology/regions?type=regions&format=json

Cool. As long as there is an easy way to obtain the CSV, I'm fine. (Maybe better to document it somewhere as well!)

> When I decided to make this feature independent, I realized that other data needed for visualization in the future can also use this feature, so I was more determined to make it independent.

Yeah this is a nice consideration. Actually we can have both. How about being a React Component? Then it is "independent" at source code level, while providing very integrated UX for users. BTW the "presets" feature should be general enough so that we can customize different presets when this visualization is used in different places.

unbyte:
My old SSD is dead :(, so the system has been reinstalled and all the dev tools and environments are being rebuilt. I will complete the expected goals as soon as possible

breeswish:
@unbyte Don't worry, take your time.

PR #861, opened 2021-02-25
Redirect root path to UI handler at /dashboard

LINKIWI:
This change proposes sending the client an HTTP 302 to `/dashboard` when the root path `/` is requested. This is a convenience feature useful for standalone/non-embedded TiDB dashboard deployments.


baurine:
hi @LINKIWI thanks for your contribution, we will have a test and review soon.

baurine:
I tested it and it works as expected. hi @breeswish , how do you think about it?

breeswish (approved):
looks good!

Issue #852, opened 2021-02-17
Try replace the vsfgen by go 1.16 `//go:embed`


baurine:
Related article: https://blog.carlmjohnson.net/post/2021/how-to-use-go-embed/

PR #863, opened 2021-03-03
Feature/go 1.16 embed

9547:
implement the #852


baurine:
hi @9547 , thanks for your contribution, I will test whether it works fine with PD, PD is still in go 1.13.

breeswish:
Currently PD is still compiling using old Golang compiler. I think we can hold this PR for now :) It won't be too long to reach go 1.16 in our CICD.

PR #873, opened 2021-03-10
server: add --log-dir to specify logstore dir

9547:
The default logstore path is under `ioutil.TempFile`. In some cases, we need to store the searched logs in a specific place.


pkg/config/config.go
```diff
@@ -40,8 +45,13 @@ type Config struct {
 }
 
 func Default() *Config {
+	dir, err := ioutil.TempDir("", "dashboard-logs")
+	if err != nil {
+		log.Fatal("Failed to create directory for storing logs", zap.Error(err))
+	}
 	return &Config{
 		DataDir:            "/tmp/dashboard-data",
+		LogDir:             dir,
```
breeswish, line 54: Looks like we should assign `""` as default?

pkg/config/config.go
```diff
@@ -29,6 +33,7 @@ const (
 
 type Config struct {
 	DataDir          string
+	LogDir           string
```
breeswish, line 36: This name is misleading. Actually it is the directory to store searched log results. How about calling it `TempStoreDir` and use it to replace the `ioutil.TempDir`? (remember to change the cli flag name as well)

breeswish (commented):
Thanks a lot for the contribution! I have a few ideas about the name.

pkg/config/config.go
9547: Sorry for the later reply, I'll change it later today.

pkg/config/config.go
9547: Thanks for the advice.

9547:
@breeswish PTAL.

breeswish (approved):
Sorry for the late review! This PR looks good!

Issue #717, opened 2020-08-04
Support other types of profile and flame graph


SunRunAway:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**
1. dashboard can only output CPU profile, it should also support memory profile, goroutine, trace profile, etc...
2. not only output the flame graph, but also be able to download the original profile documents.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

SunRunAway:
@crazycs520 @breeswish PTAL

breeswish:
For "profile documents" in 2, could you explain a bit more about its scenario?

SunRunAway:
The tidb developers can profile it with their own go toolchain in their local machine.

coocood:
SVG file lost many information in the original profile samples.

youjiali1995:
What's the progress of this feature? I don't think the default format(graph) of the TiDB profile is easy to use... I prefer to flame graph.

breeswish:
@youjiali1995 @coocood   In latest TiDB Dashboard, there is a debug API feature, that you can download the raw profile data (and other type of pprof endpoints as well). You need to start `go tool pprof` locally to analyze the downloaded raw profile though. Does this fit your need?

breeswish:
@YiniXu9506 Would you like to work on this feature so that the easy-to-use profiling UI can provide more detailed information?

coocood:
> @youjiali1995 @coocood In latest TiDB Dashboard, there is a debug API feature, that you can download the raw profile data (and other type of pprof endpoints as well). You need to start `go tool pprof` locally to analyze the downloaded raw profile though. Does this fit your need?

I think this is sufficient, profile analysis is for advanced users, we don't need to invest too much resource to make the profile UI. The UI just needs to guide the user on how to use the debug.zip file.

breeswish:
This have been implemented in https://github.com/pingcap/tidb-dashboard/pull/1095. It will be available in v5.4.0.

PR #874, opened 2021-03-11
Refine profiling

shhdgit:
Some code refine about https://github.com/pingcap/tidb-dashboard/pull/859#discussion_r582532471


pkg/apiserver/profiling/fetcher/client.go
```diff
@@ -31,80 +30,83 @@ const (
...
+	return &f, nil
 }
 
-var newFetchers = fx.Provide(func(
-	tikvClient *tikv.Client,
-	tidbClient *tidb.Client,
-	pdClient *pd.Client,
-	tiflashClient *tiflash.Client,
+func NewClientMap(
+	tikvc *tikv.Client,
```
breeswish, line 54: we prefer `tikvClient` instead of `tikvc` for better readability :)

pkg/apiserver/profiling/module.go
```diff
@@ -1,5 +1,28 @@
...
+import (
+	"go.uber.org/fx"
+
+	"github.com/pingcap/tidb-dashboard/pkg/apiserver/profiling/fetcher"
+)
 
-var Module = fx.Options(newFetchers, newService)
+var Module = fx.Options(
+	fx.Provide(
+		fetcher.NewClientMap,
```
breeswish, line 24: This leaks the internal `clientMap` to the outside (which can be accessed by other modules now). This doesn't seem to be good.

pkg/apiserver/profiling/profile.go
```diff
@@ -16,21 +16,43 @@ package profiling
...
 	switch target.Kind {
-	case model.NodeKindTiKV:
-		return fetchFlameGraphSVG(&flameGraphOptions{duration: profileDurationSecs, fileNameWithoutExt: fileNameWithoutExt, target: target, fetcher: &fts.tikv})
-	case model.NodeKindTiFlash:
-		return fetchFlameGraphSVG(&flameGraphOptions{duration: profileDurationSecs, fileNameWithoutExt: fileNameWithoutExt, target: target, fetcher: &fts.tiflash})
-	case model.NodeKindTiDB:
-		return fetchPprofSVG(&pprofOptions{duration: profileDurationSecs, fileNameWithoutExt: fileNameWithoutExt, target: target, fetcher: &fts.tidb})
-	case model.NodeKindPD:
-		return fetchPprofSVG(&pprofOptions{duration: profileDurationSecs, fileNameWithoutExt: fileNameWithoutExt, target: target, fetcher: &fts.pd})
+	case model.NodeKindTiKV, model.NodeKindTiFlash:
```
breeswish, line 36: Seems that your separation of `xxxClient` and `target` is not effective: consider the scenario that you want to **add a new fetch target**. This process will be hard and will be very easy to make mistakes. For example, it is easy to miss updating something.. Compilers just won't warn you about these mistakes. You can only discover these mistakes at runtime. This usually indicate an imperfect design.

pkg/apiserver/profiling/writer.go
```diff
@@ -0,0 +1,83 @@
...
+	"sync"
+
+	"github.com/goccy/go-graphviz"
+)
+
+var mu sync.Mutex
+
+type Writer interface {
+	Write(p []byte) (string, error)
+}
```
breeswish, line 29: Good abstraction!

breeswish (commented):
I'm also curious about how https://github.com/pingcap/tidb-dashboard/issues/717 can be implemented :) In #717, people want raw perf data, and different perf types..

breeswish:
Would you like a complete refactoring for this part, taking #717 into consideration, instead of improving the current code base? I guess it would be better. Remember to write a design doc first if you want to do this.

shhdgit:
Thanks for such a helpful review!
Yes, I want to do this refactoring. I learned a lot from these missions.

breeswish:
> Thanks for such a helpful review!
> Yes, I want to do this refactoring. I learned a lot from these missions.

:) I usually find myself knowledge++ when throwing away code instead of producing code.

PR #879, opened 2021-03-25
increase timeout to 60s when connecting to tidb database

### LINKED ISSUE #878: Dashboard cannot use under heavy load
## Bug Report

**What did you do?**

When TiDB is under heavy load, for example, needs more than 1 second to establish a connection, the TiDB dashboard cannot use any more, and we can no longer troubleshoot..

Use TiDB Dashboard to diagnose problems.

**What did you see instead?**

TiDB Dashboard cannot sign in.

master

baurine:
close #878 

but in fact, we have kinds of timeout, not sure whether it is ok to only increase this timeout.

### DIFF
--- a/pkg/tidb/client.go
+++ b/pkg/tidb/client.go
@@ -120,7 +120,7 @@ func (c *Client) OpenSQLConn(user string, pass string) (*gorm.DB, error) {
 	dsnConfig.Addr = addr
 	dsnConfig.User = user
 	dsnConfig.Passwd = pass
-	dsnConfig.Timeout = time.Second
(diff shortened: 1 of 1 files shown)


breeswish:
@baurine We also have other kind of timeouts, like proxy timeout. Could you confirm this behaviour? (e.g. whether increasing the proxy timeout is necessary).

baurine:
> @baurine We also have other kind of timeouts, like proxy timeout. Could you confirm this behaviour? (e.g. whether increasing the proxy timeout is necessary).

ok, I will check it.

PR #885, opened 2021-04-09
feat: time range limit

shhdgit:
Sync time range limit from custom time range component to common time range component.

before:
image

after:
image


ui/lib/apps/Statement/pages/List/TimeRangeSelector.tsx
```diff
@@ -124,6 +124,44 @@ export function calcValidStatementTimeRange(
...
+  if (!maxServerDataTime) {
+    return RECENT_SECONDS.map((s) => ({ enabled: false, value: s }))
+  }
+
+  const calcInterval =
+    maxServerDataTime +
+    getLocalTimeInterval(maxServerDataTime) -
+    minServerDataTime
+
+  return RECENT_SECONDS.map((s) => ({ enabled: s <= calcInterval, value: s }))
```
baurine, line 140: In my opinion, the following code may be enough?

```js
const now = Date.now()
return RECENT_SECONDS.map(s => ({enabled: now-s >= minminServerDataTime, value:s}))
```

ui/lib/apps/Statement/pages/List/TimeRangeSelector.tsx
```diff
@@ -124,6 +124,44 @@ export function calcValidStatementTimeRange(
...
+// ----first time open drop down----------...---------every time open drop down----
+//                                (time interval)
+// -----------Date.now()------------------...------------------Date.now()----------
+let getLocalTimeInterval = (time: number): number => {
+  let firstTime: number
+  if (!!time) {
+    firstTime = calcAbsIntervalFromNow(time)
+  }
+
+  getLocalTimeInterval = (time: number): number => {
```
baurine, line 154: This usage makes me a little confused, change the function definition inside itself?

baurine:
others lgtm!

ui/lib/apps/Statement/pages/List/TimeRangeSelector.tsx
shhdgit: I am afraid that the time zone of the database is different. IDK the exact scenario for this.

shhdgit: Different from the client maybe.

shhdgit: I'm afraid that the time setting on the client is wrong. This can be avoided by using time intervals.

ui/lib/apps/Statement/pages/List/TimeRangeSelector.tsx
shhdgit: Oh, this is a buggy impl. Let me re-impl it.

breeswish:
can we also restrict oldest record according to max history size?

shhdgit:
Is time range already query from record that restrict by max history size?

ui/lib/apps/Statement/pages/List/TimeRangeSelector.tsx
```diff
@@ -124,6 +124,27 @@ export function calcValidStatementTimeRange(
...
+
+  return RECENT_SECONDS.map((s) => ({ enabled: s <= validTimeRange, value: s }))
+}
+
+// use time interval to calculate valid time range when dropdown change visible state,
+// avoid error time setting on the client
+const getLocalTimeInterval = (() => {
+  const firstTimestamp = Date.now()
+  return (): number => Date.now() - firstTimestamp
+})()
```
baurine, line 146: I don't think this is necessary.

breeswish:
> Is time range already query from the record that is restricted by max history size?

Nop, the time range is based on existing records. However some records may be not evicted even long after the expiration time. This will happen if the record is not recently modified. Notice that this doesn't mean accessing a very old time range will result in correct results, since the information of other statements in this time range may be evicted .

breeswish (commented):
I'm just too lazy :) Would you like to post how it looks like now?

shhdgit:
Ok, and “restrict oldest record according to max history size” will be updated after #888 .

shhdgit:
https://user-images.githubusercontent.com/11549583/114731718-8cdd1b80-9d74-11eb-9f56-0b58353560be.mov

breeswish:
Could you change the dot dot dot icon back? I'm afraid that dot dot dot is less obvious.. The rest looks very good to me! (And actually I think we need something like https://developer.microsoft.com/en-us/fluentui#/controls/web/overflowset

shhdgit:
Okay... Maybe antd design `more` icon is to slim. We need another `more` icon.

breeswish (approved):

ui/lib/apps/Statement/pages/List/index.tsx
```diff
@@ -105,9 +101,9 @@ export default function StatementsOverview() {
   )
 
   return (
-    <div style={{ height: '100vh', display: 'flex', flexDirection: 'column' }}>
+    <div className={styles.list_container}>
```
breeswish, line 104: BTW we prefer using `styles.listContainer`. We have two conventions in the code and we can change it. This is a previously given up task in https://github.com/pingcap/tidb-dashboard/pull/768

shhdgit: Oh, wow. I just follow the "code". I also prefer camelCase.

PR #886, opened 2021-04-11
validate the statement setting values

### LINKED ISSUE #883: Should limit the tidb_stmt_summary_history_size value not beyond 255
## Bug Report

**What did you do?**

Set the statement collection interval to 22 min.

Report update setting successfully.

**What did you see instead?**

Report error:

image

Reason: currently the `tidb_stmt_summary_history_size` can't beyond 255, when setting collection interval to 22 min and retain duration to 5 days, then we get `tidb_stmt_summary_history_size` 5*24*60/22 = 328.

master

baurine:
close #883 

what did:
1. show friendly error message in local before sending request when history_size is greater than 255
2. adjust the range of data retain duration from [1d, 30day] to [4h, 10d], because we can't select a duration that is longer than 10days even we choose the longest refresh interval (60min). and when we set the refresh interval to 1min, the longest data retain duration can be about 4h.

before:

image

after:
WeCom20210411-193128@2x


breeswish:
This calculation is too hard for users.... how about refining the UI? We can change to provide two options:

1. Time window counts: Scroll between 1 and 255
2. Time window size: Select between some options
3. Max history: (display calculated value about how long history current settings will retain)

in this way, user can know how long the history is, as well as limiting the history size.

baurine:
> This calculation is too hard for users.... how about refining the UI? We can change to provide two options:
> 
> 1. Time window counts: Scroll between 1 and 255
> 2. Time window size: Select between some options
> 3. Max history: (display calculated value about how long history current settings will retain)
> 
> in this way, user can know how long the history is, as well as limiting the history size.

Good idea! will update it.

baurine:
new implementation:

WeCom20210412-233808@2x

breeswish (approved):
Cool! Could you make the text more visible? Currently they are too hard to see I think.

baurine:
Update:
image

ui/lib/apps/Statement/pages/List/StatementSettingForm.tsx
```diff
@@ -186,9 +185,9 @@ function StatementSettingForm({ onClose, onConfigUpdated }: Props) {
                       </Input.Group>
                     </Form.Item>
                     <Form.Item label={t('statement.settings.keep_duration')}>
-                      <Typography.Text disabled>
+                      <span style={{ color: '#555' }}>
```
breeswish, line 188: I think it would be better to avoid this hardcoding.. Why not just using the default color?

baurine: yep, it seems not good, just want to differ the item label and content a bit.

PR #887, opened 2021-04-13
export: refactor & add test case

### LINKED ISSUE #867: Add test cases for exports
Considering https://github.com/pingcap/tidb-dashboard/issues/865, we need to add tests to avoid further problems.

unbyte:
close #867 

looking for a suitable method to test exporting

### DIFF
--- a/pkg/apiserver/utils/export.go
+++ b/pkg/apiserver/utils/export.go
@@ -5,21 +5,107 @@ import (
 	"encoding/csv"
 	"errors"
 	"fmt"
+	"github.com/Xeoncross/go-aesctr-with-hmac"
+	"github.com/gin-gonic/gin"
+	"github.com/gtank/cryptopasta"
+	"github.com/pingcap/log"
+	"go.uber.org/zap"
 	"io"
 	"io/ioutil"
 	"os"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


pkg/apiserver/utils/export.go
```diff
@@ -5,22 +5,108 @@ import (
...
+	"strings"
 
+	"github.com/oleiade/reflections"
 	"reflect"
 	"time"
 )
 
+const namespace = "export"
+
+func ExportCSV(data [][]string, filename string) (token string, err error) {
```
breeswish, line 25: So why not serving just from the memory?

unbyte: The current behavior is downloading immediately after exporting. The purpose of writing to the temp file should be to separate the behavior of exporting and downloading.
@baurine 
Should these two actions be combined (and save data in memory)? Or provide a new user action **share** that export and show user the link to download (seems to be meaningless for downloaded file can also be shared 😂)

unbyte: Tried to combine them and incorporate the `download` action into `query list`. It looks good but it's hard to save the data as a file due to the auto-generated request layer based on axios. Give up🤐

PR #888, opened 2021-04-14
refine(ui): sider layout & introducing di

shhdgit:


ui/lib/utils/DI.md
```diff
@@ -0,0 +1,52 @@
+# Dependency Injection (DI) Documentation
```
breeswish, line 1: You may like to move it to our Wiki :)  That's where our development docs locates.
https://github.com/pingcap/tidb-dashboard/wiki

shhdgit: Ok, it's just a demo-like implementation now. I'll reorganize it and refine docs after `debug the api`.

breeswish:
Which PR should I merge first?

shhdgit:
#885 will be better.

ui/lib/utils/DI.md
```diff
@@ -0,0 +1,52 @@
...
+// MyService.ts
+import { useToken } from '@lib/utils/di'
+
+// service impl
+export const useMyService = () => {
+  // ...
+}
+
+// service token
+export const MyService = useToken(useMyService)
```
breeswish, line 21: What is a "token"?

shhdgit: Yes, just a lookup token associated with a dependency provider.
And it provide better type experience in ts. I should redesign `useToken` so that users can easily use ts interface, factory or static value.

ui/dashboardApp/layout/main/Sider/SiderService.ts
```diff
@@ -0,0 +1,36 @@
+import { useState } from 'react'
```
breeswish, line 1: I'm not sure whether it is fine to implement this as a service, as Service is usually considered something stateful and long-running, while the implementation here is stateless and is a react-hook only thing.

shhdgit: This is a borrowed concept. https://angular.io/guide/architecture-services
This is kind of concept `Business Service`, rather than concept `Technical Service`? Which means kind of business logic that can be extracted, standalone, easy for test.

shhdgit: What's meaning about "stateless and is a react-hook only here". You mean no side effect?
Maybe stateless service also make sense?

shhdgit:
These codes are quite old and now is ready to review. PTAL~

/cc @breeswish @baurine

breeswish:
If we would like to introduce mobile layout, should we provide some menu that can be collapsed totally instead?

PR #892, opened 2021-04-19
fix(ttlcache): goroutine leak

shhdgit:
Related: [pd/pull/3604](https://github.com/tikv/pd/pull/3604)

Cleanup ttl cache at end of lifecycle.

### DIFF
--- a/pkg/utils/sys_schema.go
+++ b/pkg/utils/sys_schema.go
@@ -29,11 +29,9 @@ type SysSchema struct {
 	cache *ttlcache.Cache
 }
 
-func NewSysSchema() *SysSchema {
-	c := ttlcache.NewCache()
-	c.SkipTTLExtensionOnHit(true)
+func NewSysSchema(cache *ttlcache.Cache) *SysSchema {
 	return &SysSchema{
-		cache: c,
+		cache: cache,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


shhdgit:
I'm not sure this will solve the problem https://github.com/tikv/pd/pull/3604/checks?check_run_id=2375055816. Maybe we should also add goleak testing.
@breeswish @baurine  PTAL.

pkg/utils/sys_schema.go
```diff
@@ -29,11 +29,9 @@ type SysSchema struct {
 	cache *ttlcache.Cache
 }
 
-func NewSysSchema() *SysSchema {
-	c := ttlcache.NewCache()
-	c.SkipTTLExtensionOnHit(true)
+func NewSysSchema(cache *ttlcache.Cache) *SysSchema {
 	return &SysSchema{
-		cache: c,
+		cache: cache,
```
breeswish, line 34: This may be wrong since multiple NewSysSchema will share the same cache and cause cache entity conflicts?

shhdgit: Introduce lifecycle hooks is a good idea.

breeswish (approved):
Looks much nicer!

PR #893, opened 2021-04-19
tidb: forwarder only uses tidb whose status is Up

HunDunDM:
close [tikv/pd#3612](https://github.com/tikv/pd/issues/3612)

* tidb: forwarder only uses tidb whose status is Up
  * reduce warn log `fail to recv activity from remote, stay inactive and wait to next checking round`

### DIFF
--- a/pkg/utils/topology/tidb.go
+++ b/pkg/utils/topology/tidb.go
@@ -71,7 +71,7 @@ func FetchTiDBTopology(ctx context.Context, etcdClient *clientv3.Client) ([]TiDB
 			if err == nil {
 				nodesAlive[keyParts[0]] = struct{}{}
 				if !alive {
-					log.Debug("Alive of TiDB has expired, maybe time is incorrect",
+					log.Warn("Alive of TiDB has expired, maybe local time in different hosts are not synchronized",
 						zap.String("key", key),
 						zap.String("value", string(kv.Value)))
 				}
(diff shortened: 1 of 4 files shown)


lhy1024 (approved):

breeswish (commented):
I suggest to also drop the active detection for the TiDB in  https://github.com/pingcap/tidb-dashboard/blob/master/pkg/utils/topology/tidb.go#L70 since there is a known case that TiDB will be treated as down when machine local time diffs a lot. Otherwise we will face TiDB Dashboard cannot be logged in , instead of only a "Down" notice, in such case, which is a much bigger functionality break.

pkg/utils/topology/tidb.go
```diff
@@ -69,7 +69,12 @@ func FetchTiDBTopology(ctx context.Context, etcdClient *clientv3.Client) ([]TiDB
 		case "ttl":
 			alive, err := parseTiDBAliveness(kv.Value)
 			if err == nil {
-				nodesAlive[keyParts[0]] = alive
+				nodesAlive[keyParts[0]] = struct{}{}
+				if !alive {
+					log.Debug("Alive of TiDB has expired, maybe time is incorrect",
```
breeswish, line 74: How about making it warning?

breeswish (approved):

Issue #880, opened 2021-03-29
Drop Mixpanel


breeswish:
## Feature Request

**Is your feature request related to a problem? Please describe:**

The default quota of Mixpanel can be easily exceeded in our usage scenario. We need to investigate and find other solutions. 

**Describe the feature you'd like:**

Replace Mixpanel with other services. Maybe Google Analytics can be suitable.

**Describe alternatives you've considered:**

Pay for Mixpanel. Not a best option.

**Teachability, Documentation, Adoption, Migration Strategy:**

breeswish:
Dropped.

Issue #784, opened 2020-10-30
Auto report errors and exceptions


baurine:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently, if the dashboard crashes on the user side, we don't know what happened. We need to communicate with the user.

**Describe the feature you'd like:**

Import a tool that could automatically report the errors and exceptions for us, likes [sentry](https://docs.sentry.io/platforms/javascript/)

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

breeswish:
How about utilizing current Mixpanel tracking?

baurine:
> How about utilizing current Mixpanel tracking?

That's an option, we can try it.

breezewish:
Implemented by #895

PR #895, opened 2021-04-19
monitoring: setup sentry

### LINKED ISSUE #894: Sentry setup
## Feature Request

Use Sentry.io under pingcap org for better web performance/error monitoring.

Sentry captures uncaught errors and some basic web vitals, we can also manually collect datas like routing change performance and ajax usage to help us find the bottlenecks so we can improve the user experience.

this can be done with routing change APIs and interceptors that single-spa and axios provided to track and tag each operation.

zoubingwu:
close #894 

this sentry configuration collects:

1. basic browser tracings like pageload speed, web vitals.
2. ajax call usage, including time, status, path, error report.
3. routing change events, how long it takes from route A to route B.
4. dashboard ui release info and environment.

also, single-spa was upgraded to v5.9.2, no breaking change was introduced.


ui/src/sentry.ts
```diff
@@ -0,0 +1,26 @@
...
+const release = `${name}@${version}+${hash}`
+
+// this is on purpose
+console.log('current release: ', release)
+
+const SAMPLE_RATE_PROD = 0.6
+
+Sentry.init({
+  dsn:
+    'https://10930d3b0a8d427cad2147d6d845be56@o226447.ingest.sentry.io/5721090',
```
breeswish, line 18: Maybe it should be moved to env variable to be a constant :)

ui/src/sentry.ts
```diff
@@ -0,0 +1,26 @@
+import * as Sentry from '@sentry/react'
+import { Integrations } from '@sentry/tracing'
+
+import { initSentryRoutingInstrument } from '@lib/utils/sentryHelpers'
+
+const version = process.env.REACT_APP_VERSION
+const name = process.env.REACT_APP_NAME
+const hash = process.env.REACT_APP_COMMIT_HASH
```
breeswish, line 8: Maybe better to use the information in Dashboard Version page? I'm just not sure in what case these two information will be mismatched.

breeswish (commented):
/cc @baurine @shhdgit

ui/src/sentry.ts
shhdgit: Maybe there are some differences between On-premise and cloud?

ui/src/sentry.ts
zoubingwu: @breeswish 

It might be better to have this kind of data during build time.

There are some version infos can be retrieved from `/info/info` API, but then we'll have to delay the sentry init until API response arrives if we relies on that. Also the API/network might fail or takes longer time than we expected, so...

ui/gulpfile.esm.js
```diff
@@ -21,9 +21,19 @@ task('swagger:watch', () =>
   watch(['../cmd/**/*.go', '../pkg/**/*.go'], series('swagger:generate'))
 )
 
-task('webpack:dev', shell.task('yarn react-app-rewired start'))
+task(
+  'webpack:dev',
```
shhdgit, line 25: I was wondering why we need sentry in dev mode. It's very unstable when developing.

shhdgit: Maybe we need a switch to control whether Sentry is on or not?

zoubingwu: @shhdgit What do you mean by being unstable？ Could you elaborate on this? Sending requests to sentry shouldn't hurt performance in a way you can feel.

It just tells us if it works well during dev so we can test all the features sentry provided. We can turn it off in dev tho, but it will bring a little bit more code complexity.

ui/src/sentry.ts
zoubingwu: fixed.

> Maybe there are some differences between On-premise and cloud?

currently tidb-dashboard was embedded and distributed with pd, so I'd say we keep it in env variable for now and use differenct dsn when it really needs.

ui/gulpfile.esm.js
shhdgit: My bad. Just curious about the usage of dev.
I mean there are too many useless error reports during development. May cause information overload.

zoubingwu: > may cause information overload

since we'll pass the environment property to sentry when init, it can be easily filtered.

image

> maybe we can test all features when the switch on

I mean when setting up sentry (or changing configs afterwards) there are something I'd like to be certain that it works well before we publish and distribute the code to real users. Then it cannot be avoided to send requests to sentry during dev.

shhdgit: Yes, I agree that we should test full function of Sentry before we publish. I mean like `dev-sentry` env will be better when testing sentry. Everyone in the dev env now reports to Sentry.

zoubingwu: OK, I think I get your point.

How about adding another `REACT_APP_SENTRY_ENABLED` env variable? It will be false by default in development and we can also manually change it (without commit it to git) if we want to test it in dev mode.

shhdgit: Yes, I think it's great.

zoubingwu: fixed

ui/lib/utils/sentryHelpers.ts
```diff
@@ -0,0 +1,65 @@
+import * as Sentry from '@sentry/react'
+
+const firstMountTransaction = Sentry.startTransaction({ name: 'first mount' })
+const transactionHub = new Map<string, typeof firstMountTransaction>()
+
+export const sentryEnabled = process.env.REACT_APP_SENTRY_ENABLED === 'true'
```
baurine, line 6: Maybe we should control this switch from the backend? same as the current mixpanel?

ui/src/sentry.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  const SAMPLE_RATE_PROD = 0.6
+
+  Sentry.init({
+    dsn: process.env.REACT_APP_SENTRY_DSN,
+    integrations: [new Integrations.BrowserTracing()],
+    tracesSampleRate:
+      process.env.NODE_ENV === 'production' ? SAMPLE_RATE_PROD : 1.0,
+    release,
+    environment: process.env.NODE_ENV,
```
baurine, line 26: What does environment mean? should it keep the same value as process.env?

ui/lib/utils/sentryHelpers.ts
```diff
@@ -0,0 +1,65 @@
...
+export function initSentryRoutingInstrument() {
+  window.addEventListener('single-spa:first-mount', () => {
+    firstMountTransaction.finish()
+  })
+
+  window.addEventListener('single-spa:before-routing-event', (e: any) => {
+    const transaction = markStart('routing')
+    transaction.setData('routing.detail', e.detail)
+    transaction.setTag('routing.from', e.detail.oldUrl)
+    transaction.setTag('routing.to', e.detail.newUrl)
```
baurine, line 51: Same as the mixpanel, we can't collect the URL prefix and some sensitive information.

baurine:
related issue: #880 #784 

We can drop the mixpanel by sentry as well.

zoubingwu:
@breeswish discussed with baurine about dropping mixpanel, I think we can always capture pageload speed/web vitals and errors/exceptions, and reuse the old `enable_telemetry` switch to decide if other custom metrics should be reported. 

As for the proxy host (sending request to our own server then proxy it to the original mixpanel/sentry server), do you think it is necessary when using third party SaaS tools like sentry? Any thoughts about that?

ui/lib/utils/sentryHelpers.ts
zoubingwu: fixed, now ot only collect the pathname

ui/lib/utils/sentryHelpers.ts
zoubingwu: we'll be using that switch to decide if routing and ajax usage should be collected, and should init sentry to collect basic perfs by this env variable.

ui/.env.development
```diff
@@ -1,2 +1,4 @@
-REACT_APP_MIXPANEL_TOKEN=6b8d3fb93ffd04f2e76325540c838624
-PORT=3001
```
baurine, line 2: I suggest keeping this configuration. It is because the local TiDB cluster includes grafana whose default port is 3000 as well. Without this configuration, when we run `yarn start`, it will ask us whether to use another port instead.

```
Would you like to run the app on another port instead? (Y/n)
```

ui/lib/utils/sentryHelpers.ts
baurine: Just curious, `process.env.REACT_APP_SENTRY_ENABLED` shouldn't be boolean type already when we set as `REACT_APP_SENTRY_ENABLED=true` in the env?

ui/.env.development
```diff
@@ -1,2 +1,4 @@
...
+REACT_APP_VERSION=$npm_package_version
+REACT_APP_NAME=$npm_package_name
+REACT_APP_SENTRY_DSN=https://10930d3b0a8d427cad2147d6d845be56@o226447.ingest.sentry.io/5721090
+REACT_APP_SENTRY_ENABLED=false
```
baurine, line 4: Not big deal, but better to leave a more blank line.

baurine (approved):
lgtm

ui/lib/utils/sentryHelpers.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+export function initSentryRoutingInstrument() {
+  window.addEventListener('single-spa:first-mount', () => {
+    firstMountTransaction.finish()
+  })
+
+  window.addEventListener('single-spa:before-routing-event', (e: any) => {
+    const { pathname: newUrlPath } = url.parse(e.detail.newUrl)
+    const { pathname: oldUrlPath } = url.parse(e.detail.oldUrl)
```
baurine, line 51: Because we use the hash route, not sure whether this can parse correctly.

ui/.env.development
zoubingwu: fixed, vscode was auto removing that new line for no reason..

ui/lib/utils/sentryHelpers.ts
zoubingwu: [according to nodejs doc](https://nodejs.org/api/process.html#process_process_env): 

> Assigning a property on process.env will implicitly convert the value to a string. This behavior is deprecated. Future versions of Node.js may throw an error when the value is not a string, number, or boolean.

dotenv also follows that, there is an issue about this https://github.com/motdotla/dotenv/issues/51

ui/.env.development
zoubingwu: fixed

ui/lib/utils/sentryHelpers.ts
zoubingwu: fixed

baurine (approved):
lgtm

ui/lib/utils/sentryHelpers.ts
```diff
@@ -0,0 +1,107 @@
...
+      config.headers['x-sentry-trace'] = transaction.traceId
+    }
+    return config
+  })
+
+  instance.interceptors.response.use(
+    function (response) {
+      const id = response.config.headers['x-sentry-trace']
+      if (id) {
+        const { pathname } = url.parse(response.config.url!)
```
breeswish, line 92: will we leak sensitive data when reporting the URL?

breeswish (commented):
the rest LGTM

ui/lib/utils/sentryHelpers.ts
zoubingwu: this only reports url pathname to track each API calls, quries or hostname or payloads won't be reported.

image

breeswish (approved):
Cool! BTW will we have same fields as previous Mixpanel does? For example, in Mixpanel we can know primary user browsers, OS, screen sizes, etc.

zoubingwu:
Sentry is mostly a so-called "transaction-based" monitoring tool, consider each event like an ajax call, a route change is a transaction, and every transaction has a context, for example the browser name or versions or other things we manually added into the context, so yes, we can have this kind of information like you mentioned, there is a screenshot I took from the a test project.

image

You can find them at the bottom-left of the image.

What sentry is not providing is an overall analysis like Google Analytics or Mixpanel, it is more focusing on errors and performance issues with very limited data/behavior analysis features.

It also provides a feature called "Discover", you can write your own quries to read from all the transactions sentry was saving and export or visualize the results. But it requires a business plan, we may talk to them to see if we can get any sponsoringship or discount tho since we are a open source project.

breeswish:
So seems that in order to know overall reports, like what's the most favourite browser user is using, we'd better add Google Analytics, right? /cc @baurine

breeswish (approved):

PR #896, opened 2021-04-19
feat(ui): rocksdb fields in stmt/slowquery copr tab

shhdgit:
Show rocksdb fields in Slow Query and Statement detail -> copr tab.

Slow Query:
image

Statement:
image


ui/lib/utils/callableCache.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * cachedSomeFunction.call() // e.g. 12345
+ * cachedSomeFunction.call() // 12345
+ *
+ * cachedSomeFunction.clear()
+ *
+ * cachedSomeFunction.call() // e.g. 12346
+ */
+export const cache = <Fn extends Function>(fn: Fn): Cacheable<Fn> => {
+  const call: any = (...args: any[]) => {
```
breeswish, line 22: Maybe better to memorize this functions.

ui/lib/apps/SlowQuery/pages/Detail/DetailTabCopr.tsx
```diff
@@ -30,9 +21,22 @@ export default function TabCopr({ data }: ITabCoprProps) {
...
+    genShortValueTooltipValueItem(data, 'rocksdb_key_skipped_count'),
   ]
   const columns = valueColumns('slow_query.fields.')
   return (
     <CardTable cardNoMargin columns={columns} items={items} extendLastColumn />
   )
 }
+
+// TODO: refactor items code gen for all DetailsList
+function genShortValueTooltipValueItem(data: SlowqueryModel, key: string) {
```
breeswish, line 37: What if the field does not exist? It would be confusing to display the value as 0, instead of not display at all.

ui/lib/apps/SlowQuery/pages/Detail/DetailTabCopr.tsx
```diff
@@ -30,9 +21,22 @@ export default function TabCopr({ data }: ITabCoprProps) {
       key: 'cop_wait_addr',
       value: data.cop_wait_addr,
     },
+    genShortValueTooltipValueItem(data, 'rocksdb_block_cache_hit_count'),
```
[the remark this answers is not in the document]
shhdgit: Sorry, I was in too much of a hurry. Would this be better now?

ui/lib/apps/Statement/translations/zh.yaml
```diff
@@ -152,3 +152,13 @@ statement:
     rocksdb_block_read_count_tooltip: RocksDB 从文件系统中读数据的次数 (block_read_count)
     rocksdb_block_read_byte: RocksDB 文件系统读数据量
     rocksdb_block_read_byte_tooltip: RocksDB 从文件系统中读数据的数据量 (block_read_byte)
```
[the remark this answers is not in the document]
shhdgit: These are used in the selection columns of the statement. Maybe I can extract these into a single file.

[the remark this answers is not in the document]
breeswish: I got it. So seems that there is a mistake in the previous column selection implementation, for example, in statement we should display only one column for `rocksdb_delete_skipped_count`, named as "Mean RocksDB ..." and display a bar (order by avg) with a max line.

ui/lib/apps/Statement/utils/useSchemaColumns.ts
```diff
@@ -0,0 +1,28 @@
...
+  .getInstance()
+  .statementsTableColumnsGet.bind(client.getInstance())
+
+export const useSchemaColumns = () => {
+  const [isLoading, setLoading] = useState(true)
+  const [schemaColumns, setSchemaColumns] = useState<string[]>([])
+
+  useEffect(() => {
+    const fetchSchemaColumns = async () => {
+      const { data } = await cache(statementsTableColumnsGet).call()
```
breeswish, line 16: Does this mean that it will only send requests for the first time? Then there may be issues since we want to support the upgrading scenario. In this scenario, the columns may be changed.

I think we might be able to just remove this cache, or add TTL to it.

BTW I would prefer to add the cache layer to the API (inside utils), instead of inside the business part, if it is really necessary. The reason is simple that there may be multiple business parts calling this API endpoint and they should share the same cache.

ui/lib/apps/SlowQuery/pages/Detail/DetailTabCopr.tsx
```diff
@@ -30,9 +21,25 @@ export default function TabCopr({ data }: ITabCoprProps) {
...
   const columns = valueColumns('slow_query.fields.')
   return (
     <CardTable cardNoMargin columns={columns} items={items} extendLastColumn />
   )
 }
+
+// TODO: refactor items code gen for all DetailsList
+function genShortValueTooltipValueItem(
+  data: SlowqueryModel,
+  key: keyof SlowqueryModel
```
breeswish, line 39: 👍 knowledge++

ui/lib/apps/Statement/utils/useSchemaColumns.ts
shhdgit: Yes, I'm considering adding it to useClientRequest.

breeswish: I suspect the design of this cache may take some time to be nice. Let's separate it out in this PR :)

breeswish: And, if we are already talking about useClientRequest + cache, then.... 😏 it may be time to consider something like [SWR](https://swr.vercel.app/)!

shhdgit: That's right! SWR is a good option. We do need API cache switch in some scenario, now is global configuration.

shhdgit: And I think in-memory TTL is unnecessary in client side. Maybe upgrading scenario requires its own mechanism in client side? Like refresh all cache, etc. Otherwise the upgrade logic will be separated everywhere.

shhdgit:
In table list:
image
image

In detail list:
image

ui/lib/apps/Statement/utils/useSchemaColumns.ts
shhdgit: I found that [ahooks useRequest](https://ahooks.js.org/zh-CN/hooks/async) is good enough.

breeswish: A common problem for these libraries is cancellation I remember.

ui/lib/apps/Statement/utils/tableColumns.tsx
```diff
@@ -232,34 +232,46 @@ export function statementColumns(
...
+    avgMaxColumn(
+      tcf,
+      'avg_rocksdb_block_read_count',
+      'short',
+      rows
+    ).patchConfig({
       minWidth: 220,
       maxWidth: 250,
     }),
+    avgMaxColumn(tcf, 'avg_rocksdb_block_read_byte', 'short', rows).patchConfig(
```
breeswish, line 269: bytes as a unit may be better.

ui/lib/apps/Statement/pages/Detail/PlanDetailTabCopr.tsx
```diff
@@ -4,32 +4,45 @@ import { StatementModel } from '@lib/client'
...
-    },
-  ]
+    genShortValueTooltipValueItem(data, 'avg_processed_keys'),
+    genShortValueTooltipValueItem(data, 'max_processed_keys'),
+    genShortValueTooltipValueItem(data, 'avg_total_keys'),
+    genShortValueTooltipValueItem(data, 'max_total_keys'),
+    genShortValueTooltipValueItem(data, 'avg_rocksdb_block_cache_hit_count'),
+    genShortValueTooltipValueItem(data, 'max_rocksdb_block_cache_hit_count'),
+    genShortValueTooltipValueItem(data, 'avg_rocksdb_block_read_byte'),
+    genShortValueTooltipValueItem(data, 'max_rocksdb_block_read_byte'),
```
breeswish, line 25: Seems that these two should be formatted as xxxKiB or xxxMiB.

ui/lib/apps/SlowQuery/pages/Detail/DetailTabCopr.tsx
```diff
@@ -30,9 +21,25 @@ export default function TabCopr({ data }: ITabCoprProps) {
       key: 'cop_wait_addr',
       value: data.cop_wait_addr,
     },
+    genShortValueTooltipValueItem(data, 'rocksdb_block_cache_hit_count'),
+    genShortValueTooltipValueItem(data, 'rocksdb_block_read_byte'),
```
breeswish, line 25: Ditto. It's a byte unit data.

breeswish (commented):
The rest LGTM

ui/lib/apps/Statement/utils/tableColumns.tsx
shhdgit: fixed

ui/lib/apps/Statement/pages/Detail/PlanDetailTabCopr.tsx
```diff
@@ -46,3 +50,13 @@ function genShortValueTooltipValueItem(
     value: <ShortValueWithTooltip value={Number(data[key])} />,
   }
 }
+
+function genScaledBytesTooltipValueItem(
```
breeswish, line 54: I suspect whether `genShortValueTooltipValueItem` and `genScaledBytesTooltipValueItem` is necessary :) Seems that it introduces new concepts / abstractions without reducing code duplicates.

ui/lib/components/index.ts
```diff
@@ -44,8 +44,11 @@ export * from './InstanceSelect'
 export { default as InstanceSelect } from './InstanceSelect'
 export * from './MultiSelect'
 export { default as MultiSelect } from './MultiSelect'
-export * from './ShortValueWithTooltip'
-export { default as ShortValueWithTooltip } from './ShortValueWithTooltip'
+export * from './ValueWithTooltip'
+export {
+  default as ShortValueWithTooltip,
+  ScaledBytesWithTooltip,
```
breeswish, line 50: I would prefer to use `<ValueWithTooltip.Short />`, `<ValueWithTooltip.Bytes />`, etc.

Actually it can also be `<FormatValue model="plain|short|bytes|..." value="..." />`

breeswish (approved):

ui/lib/apps/Statement/pages/Detail/PlanDetailTabCopr.tsx
shhdgit: Yes, maybe I should resolve the todo I left here before. `// TODO: refactor items code gen for all DetailsList`
There's second time I meet the code.

ui/lib/apps/Statement/utils/useSchemaColumns.ts
shhdgit: About what?

https://github.com/alibaba/hooks/blob/master/packages/use-request/src/useAsync.ts#L210

ui/lib/apps/Statement/utils/useSchemaColumns.ts
```diff
@@ -1,28 +1,29 @@
...
 const statementsTableColumnsGet = client
   .getInstance()
   .statementsTableColumnsGet.bind(client.getInstance())
 
 export const useSchemaColumns = () => {
-  const [isLoading, setLoading] = useState(true)
   const [schemaColumns, setSchemaColumns] = useState<string[]>([])
+  const { data: resp, loading } = useRequest(statementsTableColumnsGet, {
+    cacheKey: 'stmt_schema',
+    staleTime: 300000,
```
breeswish, line 14: We have a cache at server side, do we really need it?

ui/lib/apps/Statement/utils/useSchemaColumns.ts
```diff
@@ -1,28 +1,29 @@
...
-import { cache } from '@lib/utils/callableCache'
 
 const statementsTableColumnsGet = client
   .getInstance()
   .statementsTableColumnsGet.bind(client.getInstance())
 
 export const useSchemaColumns = () => {
-  const [isLoading, setLoading] = useState(true)
   const [schemaColumns, setSchemaColumns] = useState<string[]>([])
+  const { data: resp, loading } = useRequest(statementsTableColumnsGet, {
```
breeswish, line 12: I believe `useRequest` and similar hooks (like useSWR) are not well adapted if they are introduced directly. For example, as you can see, we must create a binding manually `statementsTableColumnsGet` to use, which is unpleasant. This cache is also not shared cross modules. The adoption of such libraries needs some more refinements :)

breeswish (approved):

ui/lib/apps/Statement/utils/useSchemaColumns.ts
breeswish: I'm open to implementations that even use `Proxy` to transparently handle these things, if it is needed.

shhdgit: That's right! Let me think about it.

breeswish (approved):

shhdgit:
/cc @baurine 
PTAL!

PR #898, opened 2021-04-22
feat: debug api

shhdgit:
## Preview

https://user-images.githubusercontent.com/11549583/115909906-b8f84b00-a49e-11eb-88df-90a8731b5fa5.mov


shhdgit:
/cc @breeswish @baurine 
PTAL!

pkg/apiserver/debugapi/service.go
```diff
@@ -0,0 +1,77 @@
...
+var (
+	ErrNS      = errorx.NewNamespace("error.api.debugapi")
+	ErrFromDst = ErrNS.NewType("err_from_dst")
+)
+
+func registerRouter(r *gin.RouterGroup, auth *user.AuthService, s *Service) {
+	endpoint := r.Group("/debugapi")
+	endpoint.Use(auth.MWAuthRequired())
+
+	endpoint.GET("/proxy", s.Proxy)
```
shhdgit, line 35: I'm confused about whether the proxy service should be placed in the gin routes. Is this a good practice?

pkg/apiserver/debugapi/service.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// @Summary Proxy request to tidb/tikv/tiflash/pd http api
+// @Security JwtAuth
+// @Success 200 {object} string
+// @Failure 400 {object} utils.APIError "Bad request"
+// @Failure 401 {object} utils.APIError "Unauthorized failure"
+// @Failure 500 {object} utils.APIError
+// @Router /debugapi/proxy [get]
+func (s *Service) Proxy(c *gin.Context) {
```
breeswish, line 59: I would prefer it to be something like `/debugapi/request [post]` and accept a `RequestOptions` that specify request details, like the destination address and the structural parameters. In short, instead of creating a proxy that accepts anything and do some "dangerous" proxy verifying, create an endpoint that accepts structural and safe data and then assemble it and send requests. Constructing from a structural data is always better than filtering a constructed thing.

breeswish (commented):
Highlight: An architectural change would be to pass structural data and construct things at server side, instead of constructing the things at the client side and verify it at the server side, since verifying is always worse than no need to verify at all.

pkg/apiserver/debugapi/schema/endpoint_def.go
```diff
@@ -0,0 +1,175 @@
...
+
+import (
+	"net/http"
+
+	"github.com/pingcap/tidb-dashboard/pkg/apiserver/model"
+)
+
+var tidbIPParam EndpointAPIParam = EndpointAPIParam{
+	Name:   "tidb_ip",
+	Prefix: "http://",
```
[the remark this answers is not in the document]
shhdgit: Yes I should consider using the transformer.

pkg/apiserver/debugapi/service.go
```diff
@@ -0,0 +1,77 @@
...
+	proxyReq, err := s.proxy.Request(c.Request)
+	if err != nil {
+		_ = c.Error(err)
+		return
+	}
+	s.proxy.Server.ServeHTTP(c.Writer, proxyReq)
```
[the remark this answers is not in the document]
shhdgit: Nope, now it's using whitelist. `p.Server.OnRequest(goproxy.DstHostIs(endpoint.ID))` it register when init service.

pkg/apiserver/debugapi/proxy.go
```diff
@@ -0,0 +1,75 @@
...
+	ErrRequestParam   = ErrNS.NewType("invalid_request_param")
+)
+
+type proxy struct {
+	Server *goproxy.ProxyHttpServer
+}
+
+func newProxy() *proxy {
+	proxyServer := goproxy.NewProxyHttpServer()
+	proxyServer.KeepDestinationHeaders = true
```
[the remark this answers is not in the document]
shhdgit: This retains the header returned by the component http api. `KeepHeader` option is used to preserve the client's header.
I'd like add some endpoint hooks when we need adding the custom headers at the server side.

pkg/apiserver/debugapi/client.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+type tidbImplement struct {
+	fx.In
+	Client *tidb.Client
+}
+
+func (impl *tidbImplement) Get(request *http.Request) ([]byte, error) {
+	host := request.URL.Hostname()
+	port, err := strconv.Atoi(request.URL.Port())
```
breeswish, line 61: It seems wired to "build a request" first (as in `request *http.Request`), and then modify this request to extract the request information, and then use the extracted request information to build a new request (as in `SendGetRequest`). How about just pass in the structural request information directly?

pkg/apiserver/debugapi/endpoint.go
```diff
@@ -0,0 +1,178 @@
...
+	"net/http"
+	"net/url"
+	"regexp"
+
+	"github.com/pingcap/tidb-dashboard/pkg/apiserver/model"
+)
+
+var (
+	ErrPathParamNotMatched = ErrNS.NewType("path_param_not_matched")
+	ErrValueTransformed    = ErrNS.NewType("error_value_transformed")
```
breeswish, line 27: What is this error? The meaning is not easy to understand.

pkg/apiserver/debugapi/endpoint.go
```diff
@@ -0,0 +1,178 @@
...
+
+type EndpointAPIParam struct {
+	Name string `json:"name"`
+	// represents what param is
+	Model                EndpointAPIModel `json:"model"`
+	PreModelTransformer  ModelTransformer `json:"-"`
+	PostModelTransformer ModelTransformer `json:"-"`
+}
+
+func (p *EndpointAPIParam) Transform(value string) (string, error) {
```
breeswish, line 152: It takes me several minutes to understand what is transform, what does it do and why do we need it, even when I know the overall design. Could you add some comments?

pkg/apiserver/debugapi/endpoint.go
```diff
@@ -0,0 +1,178 @@
...
+
+type ValueGetter interface {
+	Get(key string) string
+}
+
+type EndpointAPI struct {
+	ID        string             `json:"id"`
+	Component model.NodeKind     `json:"component"`
+	Path      string             `json:"path"`
+	Method    string             `json:"method"`
```
breeswish, line 38: Should method be an enum instead?

pkg/apiserver/debugapi/endpoint.go
```diff
@@ -0,0 +1,178 @@
...
+	Host      EndpointAPIParam   `json:"host"`
```
breeswish, line 39: Do we really need this host field for a model to record something? I think its value is not decided until user feeds a real host (when model is no longer a model).

pkg/apiserver/debugapi/endpoint.go
```diff
@@ -0,0 +1,178 @@
...
+	Segment   []EndpointAPIParam `json:"segment"` // e.g. /stats/dump/{db}/{table} -> db, table
```
breeswish, line 40: Maybe better to call it PathParams

pkg/apiserver/debugapi/endpoint.go
```diff
@@ -0,0 +1,178 @@
...
+	Query     []EndpointAPIParam `json:"query"`   // e.g. /debug/pprof?seconds=1 -> seconds
```
breeswish, line 41: QueryParams

pkg/apiserver/debugapi/endpoint.go
```diff
@@ -0,0 +1,178 @@
...
+var (
+	ErrPathParamNotMatched = ErrNS.NewType("path_param_not_matched")
+	ErrValueTransformed    = ErrNS.NewType("error_value_transformed")
+)
+
+type ValueGetter interface {
+	Get(key string) string
+}
+
+type EndpointAPI struct {
```
breeswish, line 34: So this is a model right? Maybe better to call it EndpointAPIModel to explicitly mark it as a model.

pkg/apiserver/debugapi/endpoint_test.go
```diff
@@ -0,0 +1,98 @@
...
+	TestingT(t)
+}
+
+var _ = Suite(&testSchemaSuite{})
+
+type testSchemaSuite struct{}
+
+var testTiDBIPParam EndpointAPIParam = EndpointAPIParam{
+	Name:  "tidb_ip",
+	Model: EndpointAPIModelIPPort,
```
breeswish, line 39: I think the `IPPort` model can be removed, since every API "instance" needs it, and there is no need to mark each API model to have a IPPort to fill.

pkg/apiserver/debugapi/endpoint.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+		value = v
+	}
+
+	return value, nil
+}
+
+type ModelTransformer func(value string) (string, error)
+
+type EndpointAPIModel struct {
```
breeswish, line 175: How about calling it `EndpointAPIParamModel`

pkg/apiserver/debugapi/client.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return impl.Client.WithEnforced(tidb.EnforcedSettingStatusAPIAddress).WithStatusAPIAddress(host, port).SendGetRequest(request.URL.Path)
```
breeswish, line 66: In your use case you are setting an enforcement rule first, then update the address by respecting the enforce rule. How about merge two into one: `impl.Client.WithEnforcedStatusAPIAddress(host, port)`?

pkg/apiserver/debugapi/endpoint.go
```diff
@@ -0,0 +1,178 @@
...
+	replacedPath = ParamRegexp.ReplaceAllStringFunc(replacedPath, func(s string) string {
+		if returnErr != nil {
+			return s
+		}
+
+		key := ParamRegexp.ReplaceAllString(s, "${1}")
+		pv, ok := pvs.Find(key)
+		// means the param can be found in the endpoint path, but not in the endpoint param
+		if !ok {
+			returnErr = ErrPathParamNotMatched.New("path: %s, param: %s", e.Path, key)
```
breeswish, line 91: According to your comment in L89, this error is more like, missing required parameter?

pkg/apiserver/debugapi/endpoint.go
```diff
@@ -0,0 +1,178 @@
...
+	return pvs
+}
+
+func (v paramValues) Find(key string) (paramValue, bool) {
+	pv, ok := v[key]
+	return pv, ok
+}
+
+type paramValue struct {
+	Param    EndpointAPIParam
```
breeswish, line 136: Is it a Model?

pkg/apiserver/debugapi/endpoint.go
```diff
@@ -0,0 +1,178 @@
...
+	RawValue string
+}
+
+func (pv *paramValue) TransformedValue() (string, error) {
+	return pv.Param.Transform(pv.RawValue)
+}
```
breeswish, line 142: Is it possible to store the transformed value directly, instead of storing a model and a raw value?

pkg/apiserver/debugapi/endpoint.go
shhdgit: Of course.

pkg/apiserver/debugapi/endpoint.go
shhdgit: Oh yes, that's right.

pkg/apiserver/debugapi/endpoint.go
shhdgit: Yes.

pkg/apiserver/debugapi/client.go
shhdgit: Do we need a parallel interface for `sqlAPIAddress` in tidb/client.go? `EnforcedSetting` is a lazy design but not necessary.

pkg/apiserver/debugapi/client.go
shhdgit: That's right. Maybe we should create another pr for http client refactor and update the client implements in debug api?

pkg/apiserver/debugapi/client.go
shhdgit: You’re right, it should be merge into single interface. And there’s nothing related to EnforcedSetting.

pkg/apiserver/debugapi/client.go
breeswish: Each client currently accepts a path for get, seems to be enough for us!

pkg/apiserver/debugapi/endpoint.go
```diff
@@ -0,0 +1,168 @@
...
+func (e *EndpointAPIModel) PopulatePath(valMap map[string]string) (string, error) {
+	var returnErr error
+	replacedPath := e.Path
+	replacedPath = ParamRegexp.ReplaceAllStringFunc(replacedPath, func(s string) string {
+		if returnErr != nil {
+			return s
+		}
+
+		key := ParamRegexp.ReplaceAllString(s, "${1}")
+		val := valMap[key]
```
breeswish, line 93: `val, ok := valMap[key]` may be better in order to distinguish empty values.

pkg/apiserver/debugapi/endpoint.go
```diff
@@ -0,0 +1,168 @@
...
+	})
+	return replacedPath, returnErr
+}
+
+func (e *EndpointAPIModel) EncodeQuery(valMap map[string]string) (string, error) {
+	query := url.Values{}
+	for _, q := range e.QueryParams {
+		// cuz paramValues was generated by e.Query, paramValues can always find param by q.Name
+		val := valMap[q.Name]
+		if q.Required && val == "" {
```
breeswish, line 110: ditto. Using `val, ok := valMap[q.name]` may be better.

pkg/apiserver/debugapi/endpoint.go
```diff
@@ -0,0 +1,168 @@
...
+			return "", ErrMissingRequiredParam.New("query param: %s", q.Name)
```
breeswish, line 111: Please make the error readable.

pkg/apiserver/debugapi/service.go
```diff
@@ -0,0 +1,116 @@
...
+	endpointMap map[string]endpoint
+}
+
+func newService(clientMap *ClientMap) (*Service, error) {
+	s := &Service{endpointMap: map[string]endpoint{}}
+
+	for _, e := range endpointAPIList {
+		client, ok := clientMap.Get(e.Component)
+		if !ok {
+			return nil, ErrComponentClient.New("%s type client not found, id: %s", e.Component, e.ID)
```
breeswish, line 56: Should this be panic instead as it seems to be definitely a programming mistake?

pkg/tidb/client.go
```diff
@@ -81,40 +83,50 @@ func NewTiDBClient(lc fx.Lifecycle, config *config.Config, etcdClient *clientv3.
...
+	var err error
+
 	overrideEndpoint := os.Getenv(tidbOverrideSQLEndpointEnvVar)
-	if overrideEndpoint != "" && c.sqlAPIAddress != "" {
+	hasEnforcedSettingSQLAPIAddress := c.enforcedSettings.Has(EnforcedSettingSQLAPIAddress)
+	// the `tidbOverrideSQLEndpointEnvVar` and the `Client.sqlAPIAddress` have the same override priority, if both exist and have not enforced `Client.sqlAPIAddress` then an error is returned
+	if overrideEndpoint != "" && c.sqlAPIAddress != "" && !hasEnforcedSettingSQLAPIAddress {
 		log.Warn(fmt.Sprintf("Reject to establish a target specified TiDB SQL connection since `%s` is set", tidbOverrideSQLEndpointEnvVar))
 		return nil, ErrTiDBConnFailed.New("TiDB Dashboard is configured to only connect to specified TiDB host")
 	}
```
breeswish, line 116: How about moving the check logic into `WithSQLAPIAddress`? Then you don't need to store the enforcedSettings any more.

breeswish (commented):
Almost good!! Congrats

pkg/apiserver/debugapi/client.go
```diff
@@ -0,0 +1,115 @@
...
+	Client *tidb.Client
+}
+
+func (impl *tidbImplement) Get(req *Request) ([]byte, error) {
+	return impl.Client.WithEnforcedStatusAPIAddress(req.Host, req.Port).SendGetRequest(req.Path)
+}
+
+func (impl *tidbImplement) Send(req *Request) ([]byte, error) {
+	return defaultSendRequest(impl, req)
+}
```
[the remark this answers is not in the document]
shhdgit: Dispatch the request function by request method. So far only GET function.

pkg/tidb/client.go
shhdgit: Does this mean I need to store errors for chained api?

pkg/apiserver/debugapi/service.go
```diff
@@ -0,0 +1,116 @@
...
+	ErrComponentClient   = ErrNS.NewType("invalid_component_client")
+	ErrEndpointConfig    = ErrNS.NewType("invalid_endpoint_config")
+	ErrInvalidStatusPort = ErrNS.NewType("invalid_status_port")
+)
+
+func registerRouter(r *gin.RouterGroup, auth *user.AuthService, s *Service) {
+	endpoint := r.Group("/debugapi")
+	endpoint.Use(auth.MWAuthRequired())
+
+	endpoint.POST("/endpoint", s.RequestEndpoint)
```
breeswish, line 37: It might be good to introduce a verb to the POST interface, like `request_endpoint`.

breeswish (approved):

shhdgit:
Thanks for so many helpful reviews! I feel knowledge+++++++++. : D
And I am now better able to distinguish between design and implementation.

baurine (approved):
LGTM!

PR #900, opened 2021-04-25
tidb_client: improve behavior when no alive tidb instance

### LINKED ISSUE #899: Misleading log and incorrect connection when no TiDB is available
## Bug Report

**What did you do?**

Use TiDB Dashboard to log in when no TiDB is available (this can be simulated by either not opening a TiDB at all, or open a TiDB using `skipRegisterToDashboard`).

1. Display nice error messages like No TiDB is alive.
2. Make no connection at all.

**What did you see instead?**

The error message is "Failed to connect to TiDB" which is misleading.

The logs shown that it is connecting to the TiDB Dashboard proxy endpoint (and then closed by the proxy because no endpoint is alive).

master.

unbyte:
closes #899 

image

image


unbyte:
/cc @breeswish @baurine 
PTAL

pkg/tidb/forwarder.go
```diff
@@ -93,10 +92,8 @@ func (f *Forwarder) pollingForTiDB() {
 			return err
 		}, bo)
 		if err != nil {
-			if errorx.IsOfType(err, ErrNoAliveTiDB) {
```
breeswish, line 96: This may mean that there is connection error to the PD. In this case TiDB can be still alive and I think there is no need to clear TiDB remotes.

pkg/tidb/forwarder.go
```diff
@@ -118,6 +115,28 @@ func (f *Forwarder) pollingForTiDB() {
...
 
+func (f *Forwarder) resolveSqlAddr(override string) (string, error) {
+	if override != "" {
+		return override, nil
+	}
+	if f.sqlProxy.noAliveRemote.Load() {
+		log.Warn("Unable to resolve sql connection address since no alive TiDB instance")
+		return "", ErrNoAliveTiDB.NewWithNoMessage()
+	}
+	return fmt.Sprintf("127.0.0.1:%d", f.sqlPort), nil
```
breeswish, line 126: These two functions seems to be too verbose. Is it possible to reduce some duplicates? For example, I believe it can be as simple as:

```golang
func (f *Forwarder) getEndpointAddr(port int) (string, error) {
    if f.sqlProxy.noAliveRemote.Load() {
        return "", ErrNoAliveTiDB.NewWithNoMessage()
    }
    return fmt.Sprintf("127.0.0.1:%d", port), nil
}
```

Note: I exclude the "override" logic because this is NOT a behaviour that a forwarder / proxyer / load balancer should have.

pkg/tidb/forwarder.go
```diff
@@ -92,12 +91,7 @@ func (f *Forwarder) pollingForTiDB() {
 			allTiDB, err = topology.FetchTiDBTopology(bo.Context(), f.etcdClient)
 			return err
 		}, bo)
-		if err != nil {
-			if errorx.IsOfType(err, ErrNoAliveTiDB) {
-				f.sqlProxy.updateRemotes(nil)
-				f.statusProxy.updateRemotes(nil)
-			}
-		} else {
+		if err == nil {
```
unbyte, line 94: `errorx.IsOfType(err, ErrNoAliveTiDB)` is a legacy code and is always false, so I simply remove entire `err != nil` block

pkg/tidb/forwarder.go
```diff
@@ -118,6 +113,14 @@ func (f *Forwarder) pollingForTiDB() {
 	}
 }
 
+func (f *Forwarder) getEndpointAddr(port int) (string, error) {
+	if f.statusProxy.noAliveRemote.Load() {
+		log.Warn("Unable to resolve connection address since no alive TiDB instance")
```
breeswish, line 118: @HunDunDM PTAL about this line to see whether it will affect PD cluster without TiDB.

breeswish (commented):
The rest LGTM. /cc @shhdgit PTAL

shhdgit:
LGTM.

pkg/tidb/proxy.go
```diff
@@ -95,12 +98,32 @@ func (p *proxy) updateRemotes(remotes map[string]struct{}) {
 		}
 		return true
 	})
+	p.noAliveRemote.Store(false)
```
shhdgit, line 101: Sorry for the slow response, why would this state change here, we don't know if any of remotes are alive yet.

pkg/tidb/client.go
```diff
@@ -111,7 +111,10 @@ func (c *Client) OpenSQLConn(user string, pass string) (*gorm.DB, error) {
 		if overrideEndpoint != "" {
 			addr = overrideEndpoint
 		} else {
-			addr = fmt.Sprintf("127.0.0.1:%d", c.forwarder.sqlPort)
+			var err error
+			if addr, err = c.forwarder.getEndpointAddr(c.forwarder.sqlPort); err != nil {
```
HunDunDM, line 115: how about `getSQLEndpointAddr`?

pkg/tidb/client.go
```diff
@@ -158,10 +164,19 @@ func (c *Client) SendGetRequest(path string) ([]byte, error) {
 		if overrideEndpoint != "" {
 			addr = overrideEndpoint
 		} else {
-			addr = fmt.Sprintf("127.0.0.1:%d", c.forwarder.statusPort)
+			var err error
+			if addr, err = c.forwarder.getEndpointAddr(c.forwarder.statusPort); err != nil {
```
HunDunDM, line 168: how about `getStatusEndpointAddr`?

pkg/tidb/forwarder.go
```diff
@@ -118,6 +113,14 @@ func (f *Forwarder) pollingForTiDB() {
 	}
 }
 
+func (f *Forwarder) getEndpointAddr(port int) (string, error) {
```
HunDunDM, line 116: ditto

HunDunDM (approved):
rest LGTM

PR #904, opened 2021-04-29
Release v2021.04.29.1

shhdgit:
Tiny updates. PTAL!
/cc @breeswish


breeswish:
I think maybe it's good to carry https://github.com/pingcap/tidb-dashboard/pull/896 , otherwise the release doesn't seem to be very useful~

shhdgit:
Yes, that would be better.

Issue #910, opened 2021-05-07
Support adjusting Statement LRU size


breeswish:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Statement by default accepts only 200 statements. In many cases it needs to be adjusted, so that a settings for it may be good.

**Describe the feature you'd like:**

Add a setting in the UI to adjust the 200 statement limitation.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

breeswish:
Implemented by #914

PR #914, opened 2021-05-10
feat(stmt): support config maximum number of stmt kept in memory

shhdgit:
Related: #910 

image

### DIFF
--- a/pkg/apiserver/statement/config.go
+++ b/pkg/apiserver/statement/config.go
@@ -80,5 +80,5 @@ func buildConfigUpdateSQL(config interface{}, extract ...string) string {
 	}
 
 	// skip `SQL string formatting (gosec)` lint
-	return "SET " + strings.Join(stmts, ",") // nolints
+	return "SET " + strings.Join(stmts, ", ") // nolints
 }
--- a/pkg/apiserver/statement/config.go
+++ b/pkg/apiserver/statement/config.go
@@ -42,7 +42,7 @@ func buildConfigQuerySQL(config interface{}) string {
 			continue
 		}
 		column := utils.GetGormColumnName(gormTag)
-		stmts = append(stmts, fmt.Sprintf("@@GLOBAL.%s as %s", column, column))
+		stmts = append(stmts, fmt.Sprintf("@@GLOBAL.%s AS %s", column, column))
 	}
 
 	// skip `SQL string formatting (gosec)` lint
--- a/pkg/apiserver/statement/config.go
+++ b/pkg/apiserver/statement/config.go
@@ -46,7 +46,7 @@ func buildConfigQuerySQL(config interface{}) string {
 	}
 
 	// skip `SQL string formatting (gosec)` lint
-	return "SELECT " + strings.Join(stmts, ",") // nolints
+	return "SELECT " + strings.Join(stmts, ", ") // nolints
 }
 
 // sql will be built like this, gorm:"column:some_global_var" -> @@GLOBAL.some_global_var = some_global_var_value
--- a/pkg/apiserver/statement/config.go
+++ b/pkg/apiserver/statement/config.go
@@ -1,0 +1,84 @@
+// Copyright 2021 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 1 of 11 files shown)


ui/lib/apps/Statement/pages/List/StatementSettingForm.tsx
```diff
@@ -134,6 +135,12 @@ function StatementSettingForm({ onClose, onConfigUpdated }: Props) {
               return (
                 getFieldValue('enable') && (
                   <Form.Item noStyle>
+                    <Form.Item
+                      name="max_size"
+                      label={t('statement.settings.max_size')}
+                    >
+                      <InputNumber min={1} />
```
shhdgit, line 142: Do we need a limit on the maximum value?

shhdgit:
/cc @breeswish @baurine 
PTAL.

pkg/apiserver/statement/service.go
```diff
@@ -89,23 +97,30 @@ func (s *Service) configHandler(c *gin.Context) {
...
+	var config EditableConfig
+	if err = c.ShouldBindJSON(&config); err != nil {
 		utils.MakeInvalidRequestErrorFromError(c, err)
 		return
 	}
 	db := utils.GetTiDBConnection(c)
-	err := updateStmtConfig(db, &req)
+
+	if !config.Enable {
+		err = db.Exec(buildConfigUpdateSQL(&config, "Enable")).Error
```
breeswish, line 115: What does this Enable do?

pkg/apiserver/statement/config.go
```diff
@@ -0,0 +1,84 @@
...
+			continue
+		}
+
+		val := configValue.Field(i)
+		column := utils.GetGormColumnName(gormTag)
+		stmts = append(stmts, fmt.Sprintf("@@GLOBAL.%s = %v", column, val))
+	}
+
+	// skip `SQL string formatting (gosec)` lint
+	return "SET " + strings.Join(stmts, ",") // nolints
```
breeswish, line 83:
```suggestion
	return "SET " + strings.Join(stmts, ", ") // nolints
```

pkg/apiserver/statement/config.go
```diff
@@ -0,0 +1,84 @@
...
+			continue
+		}
+		gormTag, ok := f.Tag.Lookup("gorm")
+		if !ok {
+			continue
+		}
+
+		val := configValue.Field(i)
+		column := utils.GetGormColumnName(gormTag)
+		stmts = append(stmts, fmt.Sprintf("@@GLOBAL.%s = %v", column, val))
```
breeswish, line 79: Seems that `val` may lead to security risks as it can be anything!

pkg/apiserver/statement/config.go
```diff
@@ -0,0 +1,84 @@
...
+		gormTag, ok := f.Tag.Lookup("gorm")
+		if !ok {
+			continue
+		}
+		column := utils.GetGormColumnName(gormTag)
+		stmts = append(stmts, fmt.Sprintf("@@GLOBAL.%s as %s", column, column))
+	}
+
+	// skip `SQL string formatting (gosec)` lint
+	return "SELECT " + strings.Join(stmts, ",") // nolints
```
breeswish, line 49:
```suggestion
	return "SELECT " + strings.Join(stmts, ", ") // nolints
```

pkg/apiserver/statement/config.go
```diff
@@ -0,0 +1,84 @@
...
+	stmts := []string{}
+	fNum := configType.NumField()
+	for i := 0; i < fNum; i++ {
+		f := configType.Field(i)
+		gormTag, ok := f.Tag.Lookup("gorm")
+		if !ok {
+			continue
+		}
+		column := utils.GetGormColumnName(gormTag)
+		stmts = append(stmts, fmt.Sprintf("@@GLOBAL.%s as %s", column, column))
```
breeswish, line 45:
```suggestion
		stmts = append(stmts, fmt.Sprintf("@@GLOBAL.%s AS %s", column, column))
```

breeswish (commented):
Nice refinement!

pkg/apiserver/statement/service.go
shhdgit: Extract fields from struct, only the extract fields will be left in SQL.

Like this:
```
func (t *testConfigSuite) Test_buildConfigUpdateSQL_extract_fields(c *C) {
	testConfigStmt := "SET @@GLOBAL.tidb_enable_stmt_summary = true"
	c.Assert(buildConfigUpdateSQL(&testConfig{Enable: true, RefreshInterval: 1800}, "Enable"), Equals, testConfigStmt)
}
```

pkg/apiserver/statement/config.go
shhdgit: Right...Let me think about it.

breeswish (approved):

PR #916, opened 2021-05-11
update(deps): upgrade gorm to v2

shhdgit:
This issue blocks [#914-comment](https://github.com/pingcap/tidb-dashboard/pull/914#discussion_r629388548)


breeswish (commented):
Let's hold this PR after the release. I'm afraid it will introduce some conflicts and undiscovered bugs.

pkg/apiserver/queryeditor/service.go
```diff
@@ -141,7 +141,11 @@ func (s *Service) runHandler(c *gin.Context) {
 	defer cancel()
 
 	startTime := time.Now()
-	colNames, rows, err := executeStatements(ctx, utils.GetTiDBConnection(c).DB(), req.Statements)
+	sqlDB, err := utils.GetTiDBConnection(c).DB()
+	if err != nil {
```
breeswish, line 145: In what case will this error happen?

pkg/apiserver/statement/config.go
```diff
@@ -74,11 +73,9 @@ func buildConfigUpdateSQL(config interface{}, extract ...string) string {
 			continue
 		}
 
-		val := configValue.Field(i)
 		column := utils.GetGormColumnName(gormTag)
-		stmts = append(stmts, fmt.Sprintf("@@GLOBAL.%s = %v", column, val))
+		stmts = append(stmts, fmt.Sprintf("@@GLOBAL.%s = @%s", column, f.Name))
```
breeswish, line 77: Hmmm. Why is it `@%s`?

pkg/apiserver/utils/gorm.go
```diff
@@ -13,10 +13,11 @@
...
+import "gorm.io/gorm/schema"
 
 func GetGormColumnName(gormStr string) string {
-	// TODO: use go-gorm/gorm/schema ParseTagSetting. Prerequisite: Upgrade to the latest version
-	columnName := strings.Split(gormStr, ":")[1]
-	return columnName
+	gormStrMap := schema.ParseTagSetting(gormStr, ";")
+	// The key will be converted to uppercase in:
+	// https://github.com/go-gorm/gorm/blob/master/schema/utils.go#L33
+	return gormStrMap["COLUMN"]
```
breeswish, line 22: Let's add a test to it, in case of gorm change some semantics in future.

pkg/apiserver/statement/service.go
```diff
@@ -113,9 +113,9 @@ func (s *Service) modifyConfigHandler(c *gin.Context) {
 	db := utils.GetTiDBConnection(c)
 
 	if !config.Enable {
-		err = db.Exec(buildConfigUpdateSQL(&config, "Enable")).Error
+		err = db.Exec(buildConfigUpdateSQL(&config, "Enable"), &config).Error
 	} else {
-		err = db.Exec(buildConfigUpdateSQL(&config)).Error
+		err = db.Exec(buildConfigUpdateSQL(&config), &config).Error
```
breeswish, line 118: Maybe better to integrate `buildConfigUpdateSQL` and `db.Exec` together! I guess the implementation will be easier to understand.

pkg/apiserver/user/auth.go
```diff
@@ -224,7 +224,12 @@ func (s *AuthService) authSQLForm(f *authenticateForm) (*utils.SessionUser, erro
 		// tidb.ErrTiDBAuthFailed
 		return nil, err
 	}
-	defer db.Close() //nolint:errcheck
+	defer func() {
+		sqlDB, err := db.DB()
+		if err == nil {
+			sqlDB.Close()
+		}
+	}()
```
breeswish, line 232: Why do we need to change to this? (I didn't find this change in [gorm v2 release note](https://gorm.io/docs/v2_release_note.html))

pkg/dbstore/dbstore.go
```diff
@@ -42,18 +41,23 @@ func NewDBStore(lc fx.Lifecycle, config *config.Config) (*DB, error) {
 
 	p := path.Join(config.DataDir, "dashboard.sqlite.db")
 	log.Info("Dashboard initializing local storage file", zap.String("path", p))
-	gormDB, err := gorm.Open("sqlite3", p)
+	gormDB, err := gorm.Open(sqlite.Open(p), &gorm.Config{
+		Logger: zapgorm2.New(log.L()),
```
breeswish, line 45: nice catch! could you test whether it really works? (for example, you can simulate it by producing some error connections). In my current implementation the logger does not really work.

pkg/apiserver/diagnose/diagnose.go
```diff
@@ -134,7 +134,7 @@ func (s *Service) genReportHandler(c *gin.Context) {
 	db := utils.TakeTiDBConnection(c)
 
 	go func() {
-		defer db.Close()
+		defer utils.RemoveAndCloseTiDBConnFrom(c) //nolint:errcheck
```
breeswish, line 137: This might cause problems I guess, since the "db" has already been taken? (That's what "take" means)

pkg/apiserver/user/auth.go
shhdgit: Cuz there's no `Close` interface in v2 `*gorm.DB`. https://pkg.go.dev/gorm.io/gorm#pkg-index

And, is it really necessary to close the connection pool?
https://github.com/go-gorm/gorm/issues/3834
https://github.com/go-gorm/gorm/issues/3535
https://github.com/go-gorm/gorm/issues/4157

pkg/apiserver/queryeditor/service.go
shhdgit: According to https://gorm.io/docs/generic_interface.html. If the underlying database connection is not a *sql.DB, like in a transaction, it will returns error.

pkg/apiserver/diagnose/diagnose.go
shhdgit: My fault.

pkg/apiserver/utils/gorm.go
shhdgit: That's right!

pkg/apiserver/statement/service.go
shhdgit: Right, let me try.

pkg/dbstore/dbstore.go
shhdgit: I think it's works by check `cat /tmp/dashboard-data/dashboard.sqlite.db` output. And why the logger not work in current implementation?

breeswish: You can simulate a database query error, like introducing a syntax error. And then you will find unexpected log output in current implementation.

breeswish (approved):

PR #923, opened 2021-05-12
ui: set sentry env as option in gulp

unbyte:
also provide a fallback hash string.


unbyte:
besides, windows doesn't support `key=value command` 🌚

breeswish:
I think we can prefer https://github.com/pingcap/tidb-dashboard/pull/930 over this PR. Close for now! :)

PR #930, opened 2021-05-26
fix(ui): localstorage compatibility issue

### LINKED ISSUE #902: Error tips: unknown field
## Bug Report

**What did you do?**

Open slowquery page in dashboard after cluster upgrade.

See the slowquery data.

**What did you see instead?**

Error tips without correct data.
- error.api.other: unknown field sql
- error.api.other: unknown field Time

v4.0.9

shhdgit:
close #902 
related: [docs](https://github.com/pingcap/docs/pull/5522)

Local storage key will have release-version suffix by using `useCompatibilityLocalstorage` helper function.

image


unbyte:
LGTM, some suggestions:

- check and auto clear items of older versions to reduce storage

- the implicitly reset of certain setting items after update may cause trouble to users, which may need to be mentioned in the document

shhdgit:
> LGTM, some suggestions:
> 
> * check and auto clear items of older versions to reduce storage
> * the implicitly reset of certain setting items after update may cause trouble to users, which may need to be mentioned in the document

Thanks for the suggestion. Considering that cleaning up localstorage after a release version change is too frequent on the cloud. Maybe we should simply use the `v2` suffix. Do you have any better suggestions?

unbyte:
is 'on cloud' means users always see the latest version? If so, two entry for the old one and the new one should be enough. 

However I prefer to mark a version having breaking change of the localstorage schema and then migrate automatically rather than use a new entry every time update, which may lead to the risk that users lose some settings when there is no conflict between two versions.

Just a suggestion.

shhdgit:
I use the package.json version field as the compatibility key. And use the release version as the sentry version.

ui/.env.development
```diff
@@ -3,3 +3,4 @@ REACT_APP_VERSION=$npm_package_version
 REACT_APP_NAME=$npm_package_name
 REACT_APP_SENTRY_DSN=https://10930d3b0a8d427cad2147d6d845be56@o226447.ingest.sentry.io/5721090
 REACT_APP_SENTRY_ENABLED=false
+RELEASE_VERSION=$set_by_config_overrides
```
breeswish, line 6: Maybe no need to set this one?

ui/config-overrides.js
```diff
@@ -106,6 +106,25 @@ const supportDynamicPublicPathPrefix = () => (config) => {
...
+  plugin.definitions['process.env'] = {
+    ...processEnv,
+    ...value,
+  }
+
+  return config
+}
+
+const getInternalVersion = () => {
+  const out = execSync(`grep -v '^\#' ../release-version`)
```
breeswish, line 124: Maybe just reading the file information and process it by code is better!

ui/lib/apps/SlowQuery/pages/List/index.tsx
```diff
@@ -27,6 +27,7 @@ import {
   MultiSelect,
 } from '@lib/components'
 import { CacheContext } from '@lib/utils/useCache'
+import { useCompatibilityLocalstorage } from '@lib/utils/useCompatibilityLocalstorage'
```
breeswish, line 30:
```suggestion
import { useVersionedLocalStorage } from '@lib/utils/useVersionedLocalStorage'
```

breeswish (commented):
Mostly LGTM! Let's prioritize this PR so that it can be carried in 4.0.14.

ui/.env.development
shhdgit: I would like to have a single source for environment variables. So this is something like a comment.
Maybe `RELEASE_VERSION=DO_NOT_EDIT_it_set_in_config-overrides.js` will be better? Or just delete this one.

ui/config-overrides.js
shhdgit: Can you explain the reasoning about this? Then I can also make better decisions next time. : D

breeswish: My suggestion is based on that it is more efficient while also code is few.

ui/.env.development
breeswish: How about `REACT_APP_RELEASE_VERSION=unknown` so that when something gets wrong we have at least a "unknown" value?

breeswish: It would be good if we can write some comments here. But it depends on how dotenv works.

ui/config-overrides.js
unbyte: #WIndows_Users_Lives_Matter

ui/dashboardApp/layout/main/index.tsx
```diff
@@ -1,9 +1,10 @@
 import React, { useState, useCallback, useEffect } from 'react'
 import { Root } from '@lib/components'
-import { useLocalStorageState } from 'ahooks'
 import { HashRouter as Router } from 'react-router-dom'
 import { useSpring, animated } from 'react-spring'
 
+import { useLocalStorageState } from '@lib/utils/useLocalstorage'
```
breeswish, line 6:
```suggestion
import { useLocalStorageState } from '@lib/utils/useLocalStorage'
```

ui/.env.development
```diff
@@ -3,3 +3,5 @@ REACT_APP_VERSION=$npm_package_version
 REACT_APP_NAME=$npm_package_name
 REACT_APP_SENTRY_DSN=https://10930d3b0a8d427cad2147d6d845be56@o226447.ingest.sentry.io/5721090
 REACT_APP_SENTRY_ENABLED=false
+# RELEASE_VERSION is set in config-overrides.js
```
breeswish, line 6: Cool!

ui/config-overrides.js
```diff
@@ -106,6 +107,43 @@ const supportDynamicPublicPathPrefix = () => (config) => {
...
+    ...value,
+  }
+
+  return config
+}
+
+const getInternalVersion = () => {
+  // react-app-rewired does not support async override config method right now,
+  // subscribe: https://github.com/timarney/react-app-rewired/pull/543
+  const data = fs.readFileSync('../release-version').toString().split(os.EOL)
```
breeswish, line 127: How about:

```js
const version = fs.readFileSync('../release-version', 'utf8')
  .split('\n')
  .map(l => l.trim())
  .filter(l => !l.startsWith('#') && l !== '')[0]

if (!version) {
  ....
}
```

ui/.env.development
```diff
@@ -3,3 +3,5 @@ REACT_APP_VERSION=$npm_package_version
...
+RELEASE_VERSION=unknown
```
breeswish, line 7: Prefer to be REACT_APP_XXX since environment variables not starting with `REACT_APP` are considered to be ignored in CRA.

breeswish (approved):

ui/src/sentry.ts
```diff
@@ -4,7 +4,7 @@ import { Integrations } from '@sentry/tracing'
 import { sentryEnabled } from '@lib/utils/sentryHelpers'
 
 if (sentryEnabled) {
-  const version = process.env.REACT_APP_VERSION
+  const version = process.env.RELEASE_VERSION
   const name = process.env.REACT_APP_NAME
   const hash = process.env.REACT_APP_COMMIT_HASH
   const release = `${name}@${version}+${hash}`
```
breeswish, line 10: I think you can substitute the `release` with the version totally.

shhdgit: Now `RELEASE_VERSION ` is the actual version of the dashboard, and `REACT_APP_VERSION ` is the suffix of LocalStorage. Is this good?

breeswish: Yes, I think it's fine, since in sentry we just want a version as well.

breeswish (approved):

PR #939, opened 2021-06-22
fix(stmt & slowquery): ignored virtual fields

shhdgit:
Append ignored virtual fields to table_columns api. Then plan_count came back.

image

### DIFF
--- a/pkg/apiserver/slowquery/model.go
+++ b/pkg/apiserver/slowquery/model.go
@@ -14,6 +14,8 @@
 package slowquery
 
 import (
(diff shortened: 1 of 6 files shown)


breeswish (approved):
Great! Could you also test whether sorting works for it?

shhdgit:
Yes, it works. Because it's sorted on the client.

breeswish (approved):

breeswish:
I think we need some rest and clean up some tech debt! For example, adding some tests.

PR #951, opened 2021-07-07
Create multi branch PRs to pd

shhdgit:
Same as the title. PTAL.
/cc @breeswish

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -8,6 +8,8 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


breeswish (commented):
Maybe we should also add the source branch to the fork branch in order to avoid conflict!

shhdgit:
Will there be conflict? We checkout from the upstream source branch.

breeswish:
> Will there be conflict? We checkout from the upstream source branch.

I mean the branch name that we make commits

shhdgit:
You're right!

breeswish (approved):

PR #963, opened 2021-07-21
refine: thread safety write intro data

shhdgit:


baurine (commented):
lgtm

pkg/utils/distro/distro.go
```diff
@@ -14,12 +14,26 @@
...
+var data atomic.Value
+
+func Replace(distro introData) {
+	data.Store(distro)
 }
 
 func Data(k string) string {
-	return data[k].(string)
+	d := data.Load().(introData)
+	if d[k] == nil {
```
breeswish, line 34: When will this happen?

pkg/utils/distro/distro.go
```diff
@@ -14,12 +14,26 @@
...
+		log.Info("Distro information doesn't exist", zap.String("key", k))
```
breeswish, line 35: Maybe we can remove this log! Since we can just know the missing key from the output.

shhdgit: Yeah, let's go!

breeswish (approved):

PR #964, opened 2021-07-22
fix: debug api pprof timeout & refine api model

shhdgit:
Fix pprof timeout exceeded caused by network overheads or something else.

image


pkg/apiserver/debugapi/endpoint/endpoint.go
```diff
@@ -29,15 +30,23 @@ var (
 	ErrInvalidParam         = ErrNS.NewType("invalid_parameter")
 )
 
+type APIModelPreHooks func(req *Request, data map[string]string, m *APIModel) error
```
breeswish, line 33:
```suggestion
type APIModelPreHook func(req *Request, data map[string]string, m *APIModel) error
```

pkg/apiserver/debugapi/endpoint/endpoint.go
```diff
@@ -29,15 +30,23 @@ var (
...
 	Method      Method         `json:"method"`
 	PathParams  []APIParam     `json:"path_params"`  // e.g. /stats/dump/{db}/{table} -> db, table
 	QueryParams []APIParam     `json:"query_params"` // e.g. /debug/pprof?seconds=1 -> seconds
+
+	PreHooks  []APIModelPreHooks  `json:"-"`
+	PostHooks []APIModelPostHooks `json:"-"`
 }
 
+// Transformers execution order:
+// APIParam.PreModelTransformer -> APIParamModel.PreTransformer -> required check -> APIParamModel.Transformer -> APIParam.PostModelTransfomer
```
breeswish, line 49: Why do we have so many transformers :/ Can we make them simple?

pkg/apiserver/debugapi/endpoint/param_model_def.go
```diff
@@ -0,0 +1,89 @@
...
+	"net/url"
+	"strconv"
+	"strings"
+
+	"github.com/thoas/go-funk"
+)
+
+var APIParamModelText = NewAPIParamModel("text")
+
+var APIParamModelBool = NewAPIParamModel("bool")
```
breeswish, line 28: Why we don't have transformers for bool? (like making 1, 2, 3, 4, 5 to true)

breeswish (commented):
The hooks and the transformers have been much complicated. Prefer to simplify these concepts.

pkg/apiserver/debugapi/endpoint/param_model_def.go
shhdgit: Nice catch. There's a bug.

pkg/apiserver/debugapi/endpoint/endpoint.go
shhdgit: Yeah, it's too complicated. Maybe I can design it as middleware. Let me try.

shhdgit:
Let me try it quickly.

pkg/apiserver/debugapi/endpoint/param.go
```diff
@@ -0,0 +1,85 @@
...
+// limitations under the License.
+
+package endpoint
+
+import "net/url"
+
+type APIParamModel interface {
+	AddMiddleware(handler ...ModelMiddlewareHandlerFunc) APIParamModel
+	GetMiddlewares(param *APIParam, isPathParam bool) []MiddlewareHandler
+}
```
breeswish, line 22: This looks very similar to `Middlewarer` but not exactly the same. Is it a mistake?

pkg/apiserver/debugapi/endpoint_def.go
```diff
@@ -0,0 +1,521 @@
...
+		ID:        "pd_region_start_key",
+		Component: model.NodeKindPD,
+		Path:      "/regions/key",
+		Method:    endpoint.MethodGet,
+		QueryParams: []*endpoint.APIParam{
+			{Model: endpoint.APIParamModelText, Name: "key", Required: true},
+			{Model: endpoint.APIParamModelInt, Name: "limit"},
+		},
+	}, endpoint.MiddlewareHandlerFunc(func(req *endpoint.Request) error {
+		req.PathValues.Set("key", url.QueryEscape(req.PathValues.Get("key")))
```
breeswish, line 334: How about making `endpoint.MiddlewareHandlerFunc` to be `OnReceiveRequest`? Then `AddEndpoint` may be not necessary any more!

pkg/apiserver/debugapi/endpoint_def.go
```diff
@@ -0,0 +1,521 @@
...
+	c.AddEndpoint(&endpoint.APIModel{
+		ID:        "pd_region_key",
+		Component: model.NodeKindPD,
+		Path:      "/region/key/{regionKey}",
+		Method:    endpoint.MethodGet,
+		PathParams: []*endpoint.APIParam{
+			{Model: endpoint.APIParamModelText, Name: "regionKey", Required: true},
+		},
+	}, endpoint.MiddlewareHandlerFunc(func(req *endpoint.Request) error {
+		req.PathValues.Set("regionKey", url.QueryEscape(req.PathValues.Get("regionKey")))
```
breeswish, line 299: Is it possible for them to be a special ParamModel (which inherits Text model while having a middleware)?

breeswish (commented):
Mostly fine!

pkg/apiserver/debugapi/endpoint_def.go
```diff
@@ -0,0 +1,521 @@
...
+	Model: endpoint.APIParamModelEnum([]endpoint.EnumItem{
+		{Name: "allocs"},
+		{Name: "block"},
+		{Name: "cmdline"},
+		{Name: "goroutine"},
+		{Name: "heap"},
+		{Name: "mutex"},
+		{Name: "profile"},
+		{Name: "threadcreate"},
+		{Name: "trace"},
```
breeswish, line 39: Maybe better to call them "Value" (and let the missing name to be filled with value).

pkg/apiserver/debugapi/endpoint_def.go
```diff
@@ -0,0 +1,521 @@
...
+		PathParams: []*endpoint.APIParam{
+			pprofKindsParam,
+		},
+		QueryParams: []*endpoint.APIParam{
+			{Model: endpoint.APIParamModelConstant("1"), Name: "debug"},
+			pprofSecondsParam,
+		},
+	}, endpoint.MiddlewareHandlerFunc(func(req *endpoint.Request) error {
+		return timeoutMiddleware(req, req.QueryValues.Get("seconds"))
+	}))
```
breeswish, line 193: As a middleware, this would be better:

```
c.AddEndpoint(&Model{
  ...
  Middlewares: []{ timeoutMiddleware("seconds") },
})      
```

pkg/apiserver/debugapi/endpoint/param_model_def.go
```diff
@@ -0,0 +1,109 @@
...
+		return item
+	}).([]EnumItem)
+	enumModel := &enumAPIParamModel{*apiParamModelEnum, items}
+	enumModel.AddMiddleware(func(p *ModelParam) error {
+		isValid := funk.Contains(items, func(item EnumItem) bool {
+			v := p.Value()
+			return item.Value == v
+		})
+		if !isValid {
+			return fmt.Errorf("param is not a valid enum type, value: %s", p.Value())
```
breeswish, line 81:
```suggestion
			return fmt.Errorf("%s is not a valid value", p.Value())
```

pkg/apiserver/debugapi/endpoint/param_model_def.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+		return item
+	}).([]EnumItem)
+	enumModel := &enumAPIParamModel{*apiParamModelEnum, items}
+	enumModel.AddMiddleware(func(p *ModelParam) error {
+		isValid := funk.Contains(items, func(item EnumItem) bool {
+			v := p.Value()
+			return item.Value == v
+		})
+		if !isValid {
```
breeswish, line 80: Could you add tests about this invalid logic?

pkg/apiserver/debugapi/endpoint/param_model_def.go
```diff
@@ -0,0 +1,109 @@
...
+	p.SetValues(funk.Map(vals, func(str string) string {
+		v, _ := url.QueryUnescape(str)
+		return v
+	}).([]string))
+	return nil
+})
+
+var APIParamModelInt = NewAPIParamModel("int").AddMiddleware(func(p *ModelParam) error {
+	if _, err := strconv.Atoi(p.Value()); err != nil {
+		return fmt.Errorf("param not a number")
```
breeswish, line 50: Maybe better to include the param name in the error message (or wrap with param name at outside?)

pkg/apiserver/debugapi/endpoint/param.go
shhdgit: I want to make it easier to define the model middleware and restrict data access to the model middleware, so the interface is different. A little strange, any better suggestions?

pkg/apiserver/debugapi/endpoint_def.go
shhdgit: That makes sense. I went the opposite way.

pkg/apiserver/debugapi/endpoint/param_model_def.go
shhdgit: https://github.com/pingcap/tidb-dashboard/pull/964/files#diff-2baed377c0061c67cf03809fcd8bade87ac17e0620be46ec26122aa4c4fa5e72R149

pkg/apiserver/debugapi/client.go
```diff
@@ -66,36 +98,69 @@ func buildRelativeURI(path string, query string) string {
...
 func (impl *tiflashImplement) Get(req *endpoint.Request) (*httpc.Response, error) {
 	return impl.Client.
-		WithTimeout(defaultTimeout).
-		Get(req.Host, req.Port, buildRelativeURI(req.Path, req.Query))
+		WithTimeout(req.Timeout).
+		Get(req.Host, req.Port, buildRelativeURI(req.Path(), req.Query()))
+}
+
+func (impl *tiflashImplement) CheckDest(host string, port int) (bool, error) {
+	_, info, err := topology.FetchStoreTopology(impl.PDClient)
```
breeswish, line 157: I believe this needs more careful treatment. For example, logically, the host restriction should be performed at the client side. In this way multiple services are benefited. However considering the scope I think you can separate them to multiple PRs.

pkg/apiserver/debugapi/client.go
```diff
@@ -66,36 +98,69 @@ func buildRelativeURI(path string, query string) string {
...
 func (impl *tikvImplement) Get(req *endpoint.Request) (*httpc.Response, error) {
 	return impl.Client.
-		WithTimeout(defaultTimeout).
-		Get(req.Host, req.Port, buildRelativeURI(req.Path, req.Query))
+		WithTimeout(req.Timeout).
+		Get(req.Host, req.Port, buildRelativeURI(req.Path(), req.Query()))
+}
+
+func (impl *tikvImplement) CheckDest(host string, port int) (bool, error) {
+	info, _, err := topology.FetchStoreTopology(impl.PDClient)
```
breeswish, line 135: Fetch topology deserves a small cache (e.g. 10s).

pkg/apiserver/debugapi/client.go
```diff
@@ -26,34 +28,64 @@ import (
...
+	"github.com/pingcap/tidb-dashboard/pkg/utils/topology"
+	"github.com/thoas/go-funk"
 )
 
 const (
-	defaultTimeout = time.Second * 45 // Default profiling can be as long as 30s.
+	defaultTimeout = time.Second * 35 // Default profiling can be as long as 30s. Add 5 seconds for other overheads.
 )
 
+type Dispatcher struct {
```
breeswish, line 39: What is a Dispatcher? BTW I think it is not a good idea to abuse the Fx. For example, in this case seems that Dispatcher is totally internal. Then it should not be exposed to the outside.

pkg/apiserver/debugapi/endpoint/param_model_def.go
```diff
@@ -0,0 +1,115 @@
...
+}
+type EnumItem struct {
+	Name  string `json:"name"`
+	Value string `json:"value"`
+}
+
+func APIParamModelEnum(items []EnumItem) APIParamModel {
+	items = funk.Map(items, func(item EnumItem) EnumItem {
+		if item.Value == "" {
+			panic("enum item requires a valid value")
```
breeswish, line 74: I guess empty value is valid case in future :)

pkg/apiserver/debugapi/endpoint_def.go
```diff
@@ -0,0 +1,517 @@
...
+		Component: model.NodeKindTiKV,
+		Path:      "/config",
+		Method:    endpoint.MethodGet,
+	}).
+		AddEndpoint(&endpoint.APIModel{
+			ID:        "tikv_profile",
+			Component: model.NodeKindTiKV,
+			Path:      "/debug/pprof/profile",
+			Method:    endpoint.MethodGet,
+			Ext:       ".svg",
```
breeswish, line 488: Is it possible to use mimeTypes to handle it? From the architecture point of view it is wired for an endpoint to have a n "Ext" field. This is very coupled with the business.

pkg/apiserver/debugapi/endpoint_def.go
```diff
@@ -0,0 +1,517 @@
...
+	Name: "seconds",
+}
+
+func timeoutMiddleware(req *endpoint.Request, sec string) error {
+	i, err := strconv.ParseInt(sec, 10, 64)
+	if err != nil {
+		return err
+	}
+	duration := time.Duration(i) * time.Second
+	req.Timeout = duration + duration/2
```
breeswish, line 59: I think it is not a good idea to store Timeout in the request. This makes request no longer a request.

pkg/apiserver/debugapi/endpoint/request.go
```diff
@@ -0,0 +1,69 @@
...
+		key := pathReplaceRegexp.ReplaceAllString(s, "${1}")
+		val := r.PathValues.Get(key)
+		return val
+	})
+	return path
+}
+
+func (r *Request) Query() string {
+	return r.QueryValues.Encode()
+}
```
breeswish, line 69: Where does this Request come from? Seems that I incorrectly assumed that it comes from the frontend?

pkg/apiserver/debugapi/endpoint/middleware.go
```diff
@@ -0,0 +1,31 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package endpoint
+
+type Middlewarer interface {
+	AddMiddleware(handler ...MiddlewareHandler)
+	GetMiddlewares() []MiddlewareHandler
+}
```
breeswish, line 20: Do we really have implementor for it?

breeswish (commented):
I suggest to give a long comment, maybe with some ASCII graphs, to tell the reader about the architecuture: the browser, the server, and the API endpoint, what data structure is exchanged by what kind of functions among these three parties.

pkg/apiserver/debugapi/endpoint_def.go
shhdgit: Maybe use context with timeout?

pkg/apiserver/debugapi/endpoint/request.go
shhdgit: Request is manipulated and assembled on the server side. From the frontend, only the value of param.

pkg/apiserver/debugapi/endpoint_def.go
shhdgit: The endpoint doesn't return the correct mime type for now.
I have an idea to implement the "true middleware" with `next`. Then we got a AOP struct. We can manipulate response mime type in custom middleware after the "send request middleware" exec `next`. That would be better.

pkg/apiserver/debugapi/client.go
shhdgit: Maybe call it fetcher or add some notes to it? Yes, this should be injected via the parameter instead of Fx.

pkg/apiserver/debugapi/endpoint_def.go
```diff
@@ -0,0 +1,537 @@
...
+			QueryParams: []*endpoint.APIParam{
+				pprofSecondsParam,
+			},
+			Middleware: func(ctx *endpoint.Context) {
+				if err := timeoutMiddleware(ctx.Request, ctx.Request.QueryValues.Get("seconds")); err != nil {
+					ctx.Abort(err)
+					return
+				}
+				ctx.Next(func() error {
+					ctx.Response.Header.Set("Content-Type", "image/svg+xml")
```
breeswish, line 505: This is very wired~~ Why isn't it set in the response but we need to manually overwrite?

pkg/apiserver/debugapi/param_model_def.go
```diff
@@ -0,0 +1,116 @@
...
+		v, _ := url.QueryUnescape(str)
+		return v
+	}).([]string))
+	ctx.Next()
+})
+
+var APIParamModelInt = endpoint.NewAPIParamModel("int").Use(func(p *endpoint.ModelParam, ctx *endpoint.Context) {
+	if _, err := strconv.Atoi(p.Value()); err != nil {
+		ctx.Abort(fmt.Errorf("[%s] %s is not a number", p.Name(), p.Value()))
+		return
```
breeswish, line 58: just returning error is enough. No need to be a complex `Abort` / `Next`.

pkg/apiserver/debugapi/param_model_def.go
```diff
@@ -0,0 +1,116 @@
...
+var APIParamModelBool = endpoint.NewAPIParamModel("bool").Use(func(p *endpoint.ModelParam, ctx *endpoint.Context) {
+	if funk.Contains(falselyValues, p.Value()) {
+		p.SetValue("false")
+	} else {
+		p.SetValue("true")
+	}
+	ctx.Next()
+})
+
+var APIParamModelTags = endpoint.NewAPIParamModel("tags").Use(func(p *endpoint.ModelParam, ctx *endpoint.Context) {
```
breeswish, line 46: Seems that this is more like a MultiValue? (and maybe better to allow customizing the separator).

pkg/apiserver/debugapi/param_model_def_test.go
```diff
@@ -0,0 +1,167 @@
...
+		Path:      "/test",
+		Method:    http.MethodGet,
+		QueryParams: []*endpoint.APIParam{
+			{
+				Name:  "param1",
+				Model: APIParamModelEnum([]EnumItem{{Value: value1}}),
+			},
+		},
+	}
+	client.RegisterEndpoint([]*endpoint.APIModel{testEndpoint})
```
breeswish, line 153: Seems that endpoint and API should be unified to one, reducing number of concepts.

pkg/apiserver/debugapi/endpoint_def.go
```diff
@@ -0,0 +1,537 @@
...
+		return err
+	}
+	duration := time.Duration(i) * time.Second
+	req.Timeout = duration + duration/2
+	return nil
+}
+
+func registerEndpoint(c *endpoint.Client) {
+	// tidb
+	c.RegisterEndpoint([]*endpoint.APIModel{
```
breeswish, line 65:
```suggestion
	c.RegisterEndpoints([]*endpoint.APIModel{
```

pkg/apiserver/debugapi/endpoint_def.go
```diff
@@ -0,0 +1,537 @@
...
+	i, err := strconv.ParseInt(sec, 10, 64)
+	if err != nil {
+		return err
+	}
+	duration := time.Duration(i) * time.Second
+	req.Timeout = duration + duration/2
+	return nil
+}
+
+func registerEndpoint(c *endpoint.Client) {
```
breeswish, line 63: Seems that this function can be eliminated and can be moved to `newClient()`, since it is not necessary for us to register multiple different endpoints (as endpoints are static):

```go
func newClient(..., endpoints: []endpoint.APIModel) {
    // deal with the endpoints, build hashmap for it.
}
```

pkg/apiserver/debugapi/service.go
```diff
@@ -101,18 +89,7 @@ func (s *Service) RequestEndpoint(c *gin.Context) {
...
-		return
-	}
-	endpointReq, err := ep.NewRequest(req.Host, req.Port, req.Params)
-	if err != nil {
-		_ = c.Error(err)
-		return
-	}
-
-	res, err := SendRequest(ep.Client, endpointReq)
+	res, err := s.Client.Send(req.ID, req.Host, req.Port, req.Params)
```
breeswish, line 92: How about rearranging the code as follows:

debugapi/endpoint/:

```go
type EndpointRequestPayload struct {
  EndpointID     string            `json:"endpoint_id"`
  Host   string            `json:"host"`
  Port   int               `json:"port"`
  Params map[string]string `json:"params"`
}

type EndpointRequestPayloadResolved struct {
  originalPayload EndpointRequestPayload
  timeout time.Duration
  params map[string]interface{}
}

func (api *APIModel) ResolvePayload(payload EndpointRequestPayload) (*EndpointRequestPayloadResolved, error) {
  resolved := &resolvedEndpointRequestPayload{
    originalPayload: payload,
    params: new(...),
  }
  // 1. execute `api.OnResolve(resolved)` to update the request the timeout.
  // 2. execute each parameter's `OnResolve(resolved)` to verify payload and update the `resolved.params`
  return resolved, nil
}

func (xx) SendToEndpoint(payload EndpointRequestPayload) .. {
  // 1. resolve the endpoint via the endpoint_id in payload
  // 2. 
  resolvedPayload := endpoint.Resolve(payload)
  // 3. Send to corresponding client with the `resolvedPayload`.
}
```

I believe this makes your code shorter, since there are just less concepts involved.

pkg/apiserver/debugapi/service.go
```diff
@@ -41,27 +41,15 @@ func registerRouter(r *gin.RouterGroup, auth *user.AuthService, s *Service) {
...
-	for _, e := range endpoint.APIListDef {
-		client, ok := (*clientMap)[e.Component]
-		if !ok {
-			panic(fmt.Sprintf("%s type client not found, id: %s", e.Component, e.ID))
-		}
-		s.endpointMap[e.ID] = endpointModel{APIModel: e, Client: client}
-	}
-
-	return s, nil
+func newService(fp fetcherParam) *Service {
```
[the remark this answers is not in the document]
shhdgit: Yes, `dig` won't check the visibility in this case. It's like a provider search token.

pkg/apiserver/debugapi/client.go
```diff
@@ -19,41 +19,64 @@ import (
...
+}
+
+type ClientMap map[model.NodeKind]Client
+
 type Client interface {
 	Get(request *endpoint.Request) (*httpc.Response, error)
 }
 
-type ClientMap map[model.NodeKind]Client
+type fetcherParam struct {
```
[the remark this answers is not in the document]
shhdgit: It's used to inject internal parameters at `func newService(fp fetcherParam) *Service`.

shhdgit:
/cc @breeswish
 
Some updates, PTAL, thanks~

pkg/apiserver/debugapi/param_model_def.go
```diff
@@ -0,0 +1,136 @@
...
+			Type: "enum",
+			OnResolve: func(v *endpoint.ResolvedValues) error {
+				isValid := funk.Contains(items, func(item EnumItem) bool {
+					v := v.GetValue()
+					return item.Value == v
+				})
+				if !isValid {
+					return fmt.Errorf("[%s] %s is not a valid value", v.Name(), v.GetValue())
+				}
+				return nil
```
breeswish, line 103: Sounds like if you just use a `for` to iterate the array the lines of code is fewer, and clearer:

```go
for _, item := range items {
    if item.Value == v.GetValue() {
        return nil
    }
}
return fmt.Errorf(...)
```

pkg/apiserver/debugapi/param_model_def.go
```diff
@@ -0,0 +1,136 @@
...
+			v.SetValue("false")
+		} else {
+			v.SetValue("true")
+		}
+		return nil
+	},
+}
+
+var APIParamModelMultiValue = &endpoint.BaseAPIParamModel{
+	Type: "multi_value",
```
breeswish, line 53: This is only valid for values in the query. It cannot resolve a multi value pattern for path. Please add a `query` in the type.

pkg/apiserver/debugapi/param_model_def.go
```diff
@@ -0,0 +1,136 @@
...
+	OnResolve: func(v *endpoint.ResolvedValues) error {
+		vals := strings.Split(v.GetValue(), ",")
+		v.SetValues(funk.Map(vals, func(str string) string {
+			v, _ := url.QueryUnescape(str)
```
breeswish, line 57: Seems that there is an unescape behaviour. What kind of text will user input? Does user need to input an escaped value instead?

pkg/apiserver/statement/queries.go
```diff
@@ -165,7 +165,7 @@ func (s *Service) queryPlanDetail(
 		return
 	}
 
-	query := db.
+	query := db.Debug().
```
breeswish, line 168: Please remove it.

pkg/apiserver/debugapi/endpoint_def.go
```diff
@@ -0,0 +1,539 @@
...
+		QueryParams: []*endpoint.APIParam{
+			pprofSecondsParam,
+			pprofDebugParam,
+		},
+		OnResolve: func(resolvedPayload *endpoint.ResolvedRequestPayload) error {
+			if err := timeoutMiddleware(resolvedPayload, resolvedPayload.QueryParams.Get("seconds")); err != nil {
+				return err
+			}
+			return nil
+		},
```
breeswish, line 204: this can be `OnResolve: resolveTimeoutFromQuery("seconds")`

pkg/apiserver/debugapi/endpoint_def.go
```diff
@@ -0,0 +1,539 @@
...
+		Path:      "/stores",
+		Method:    endpoint.MethodGet,
+		QueryParams: []*endpoint.APIParam{
+			{Model: APIParamModelMultiValue, Name: "state"},
+		},
+		OnResolve: func(resolvedPayload *endpoint.ResolvedRequestPayload) error {
+			vals := resolvedPayload.QueryParams["state"]
+			resolvedPayload.QueryParams.Del("state")
+			if len(vals) != 0 {
+				for _, state := range vals {
```
breeswish, line 450: 1 less indention level: 
```go
			if len(vals) != 0 {
				return nil
			}
```

pkg/apiserver/debugapi/endpoint_def.go
```diff
@@ -0,0 +1,539 @@
...
+		},
+	},
+	// tikv
+	{
+		ID:        "tikv_config",
+		Component: model.NodeKindTiKV,
+		Path:      "/config",
+		Method:    endpoint.MethodGet,
+	},
+	// TODO: open after http client refactor
```
breeswish, line 496: Why?

pkg/apiserver/debugapi/endpoint/param.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (v *ResolvedValues) GetValues() []string {
+	return v.Values[v.param.Name]
+}
+
+func (v *ResolvedValues) SetValues(val []string) {
+	v.Values[v.param.Name] = val
+}
+
+type ParamResolveFn func(v *ResolvedValues) error
```
breeswish, line 44: How about making it to be `func (v string) ([]string, error)`? Since for `ParamResolveFn`, seems that it always deal with a single param. So that there is no necessary to carry other information in this function. Then `ResolvedValues` is not necessary any more.

pkg/apiserver/debugapi/http_client.go
```diff
@@ -0,0 +1,130 @@
...
+		return nil, fmt.Errorf("invalid request method `%s`, host: %s, path: %s", payload.Method, payload.Host, payload.Path())
+	}
+}
+
+func buildRelativeURI(path string, query string) string {
+	if len(query) == 0 {
+		return path
+	}
+	return fmt.Sprintf("%s?%s", path, query)
+}
```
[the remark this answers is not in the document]
shhdgit: Make sense.

pkg/apiserver/debugapi/endpoint/client.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	return c.httpClient.Fetch(resolvedPayload)
+}
+
+func (c *Client) GetAPIModel(id string) *APIModel {
+	return c.apiMap[id]
+}
+
+func (c *Client) GetAllAPIModels() []APIModel {
```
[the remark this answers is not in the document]
shhdgit: That's right! It's strange that I got the bug of using pointer and swag together before, but now it's fine.

pkg/apiserver/debugapi/endpoint/client.go
```diff
@@ -0,0 +1,161 @@
...
+
+func (c *Client) resolve(payload *RequestPayload) (*ResolvedRequestPayload, error) {
+	api, ok := c.apiMap[payload.EndpointID]
+	if !ok {
+		return nil, fmt.Errorf("invalid endpoint id: %s", payload.EndpointID)
+	}
+
+	resolvedPayload := &ResolvedRequestPayload{
+		originalPayload: payload,
+		pathSchema:      api.Path,
```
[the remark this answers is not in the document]
shhdgit: Maybe I should remove both the `originalPayload` and `pathSchema`, then add `APIModel` into `ResolvedRequestPayload`?

pkg/apiserver/debugapi/endpoint/client.go
```diff
@@ -0,0 +1,161 @@
...
+
+		resolvedValues, err := param.Model.Resolve(param, payload.Params[param.Name])
+		if err != nil {
+			return err
+		}
+
+		if isPathParam {
+			resolvedPayload.PathParams[param.Name] = resolvedValues.Get(param.Name)
+		} else {
+			resolvedPayload.QueryParams[param.Name] = resolvedValues.Values[param.Name]
```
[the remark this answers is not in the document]
shhdgit: It looks like `url.Values` has no function to get the completed `[]string` data.

pkg/apiserver/debugapi/http_client.go
```diff
@@ -0,0 +1,130 @@
...
+			model.NodeKindPD:      &p.PDImpl,
+		},
+	}
+}
+
+func (d *HTTPClient) Fetch(payload *endpoint.ResolvedRequestPayload) (*httpc.Response, error) {
+	if payload.Timeout <= 0 {
+		payload.Timeout = defaultTimeout
+	}
+	c := d.clients[payload.Component]
```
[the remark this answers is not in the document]
shhdgit: `payload.Component` is `model.NodeKind` type. Currently, all `NodeKind` have a corresponding client implementation.

pkg/apiserver/statement/queries.go
shhdgit: Nice catch!

pkg/apiserver/debugapi/endpoint_def.go
shhdgit: Master!

pkg/apiserver/debugapi/endpoint_def.go
shhdgit: Here we need http client support to add the header to fetch the pprof data:
`resolvedPayload.Header.Add("Content-Type", "application/protobuf")`

or we need to support override response header to fetch the svg data.:
`xxx.Response.Header.Set("Content-Type", "image/svg+xml")`

pkg/apiserver/debugapi/endpoint/param.go
shhdgit: Yeah, that's right! That's what I forgot in my mind.

pkg/apiserver/debugapi/endpoint/apimodel.go
```diff
@@ -0,0 +1,73 @@
...
+func (m *APIModel) Resolve(resolvedPayload *ResolvedRequestPayload) error {
+	if m.OnResolve == nil {
+		return nil
+	}
+	return m.OnResolve(resolvedPayload)
+}
+
+// EachParams simplifies the process of iterating over path & query params
+// and ends the iteration when return error
+func (m *APIModel) ForEachParam(fn func(p *APIParam, isPathParam bool) error) error {
```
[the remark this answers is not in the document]
shhdgit: Me neither. Let me figure out a way.

shhdgit:
Looks much easier, Thanks!

/cc @breeswish

pkg/apiserver/debugapi/endpoint_def.go
breeswish: What will happen if we do nothing?

shhdgit: we will get svg data without the ext?

breeswish (approved):

pkg/apiserver/debugapi/endpoint_def.go
breeswish: interesting.. Does TiKV not respond a correct content-type?

shhdgit: Yeah, maybe you can help me take a look at this pr~ https://github.com/tikv/tikv/pull/10951

PR #970, opened 2021-07-27
tweak: slow query & stmt db select placeholder

shhdgit:
The placeholder description of the db selection in the slow query & stmt toolbar is ambiguous. Now unified with detail page.

Update:
`所有数据库` -> `执行数据库`
`All Databases` -> `Execution Databases`

### DIFF
--- a/ui/lib/apps/SlowQuery/pages/List/index.tsx
+++ b/ui/lib/apps/SlowQuery/pages/List/index.tsx
@@ -74,10 +74,7 @@ function List() {
   } = controller
 
   function exportCSV() {
-    const hide = message.loading(
-      t('statement.pages.overview.toolbar.exporting') + '...',
-      0
-    )
+    const hide = message.loading(t('slow_query.toolbar.exporting') + '...', 0)
     downloadCSV().finally(hide)
   }
 
@@ -93,8 +90,8 @@ function List() {
     <Menu onClick={menuItemClick}>
       <Menu.Item key="export" disabled={downloading} icon={<ExportOutlined />}>
         {downloading
-          ? t('statement.pages.overview.toolbar.exporting')
-          : t('statement.pages.overview.toolbar.export')}
+          ? t('slow_query.toolbar.exporting')
+          : t('slow_query.toolbar.export')}
       </Menu.Item>
     </Menu>
   )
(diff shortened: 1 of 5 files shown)


shhdgit:
/cc @breeswish @baurine 
PTAL, thanks!

baurine (approved):
lgtm

breeswish:
The problem is that, now:

- Select foo database: Displayed as "1 Databases"
- Select foo and bar databases: Displayed as "2 Databases"
- Select all databases: Displayed as "All Databases"

After this change, the UX is no longer consistent:

- Select foo database: Displayed as "1 Databases"
- Select foo and bar databases: Displayed as "2 Databases"
- Select all databases: Displayed as "Execution Databases"

This will lead to confusions.

shhdgit:
Actually, previous "All Databases" is used as a placeholder:
- Not selected: Displayed as "All Databases"
- Select foo database: Displayed as "1 Databases"
- Select foo and bar databases: Displayed as "2 Databases"
- Select all databases(5 databases): Displayed as "5 Databases"

After this change:
- Not selected: Displayed as "Execution Databases"
- Select foo database: Displayed as "1 Databases"
- Select foo and bar databases: Displayed as "2 Databases"
- Select all databases(5 databases): Displayed as "5 Databases"

breeswish:
@shhdgit I guess you have some misunderstandings. The UI now looks like:

image

As you can see, all components are displaying the current state:
- "Recent 30 mins" indicate that the (time range is to display) recent 30 min.
- "All database" indicates that the (database filter is to display) all databases.
- "All kinds" indicates that the (kind filter is to display) all kinds.
- (Empty keywords) indicates that the keyword filter is to not filter by any keywords.

If you change "All databases" to "Execution databases" then they are not consistent any more.

How about changing to this, which also gives user hint about what's the definition of "Database":

image

shhdgit:
I got you. Give the hint will be better.
But there's another question about "All database". When you don't select any databases, it means all the sqls that contains no execution via `use databases` is displayed?
Maybe we should use placeholder as hint instead of state. No value in filter field would have implied that all data is displayed. Like "all kinds" or "all sql with/without use databases".

breeswish:
@shhdgit I think what you are talking about is to distinguish `null` and `""`? There could be `db` valued as `""`. When I want to filter by it, it indicates that I would like to see statements where db is `""`. This scenario is currently unsupported in both frontend and backend. It can be implemented, but how it is displayed needs some careful consideration. For example, maybe displaying as *(null)* instead of just empty strings?

shhdgit:
I think it's clear enough to describe the column title as execution database~

image

breeswish (approved):

Issue #973, opened 2021-08-05
Add timezone information along with all timestamp


SunRunAway:
## Feature Request

**Is your feature request related to a problem? Please describe:**

For example: 
1. cluster info
![image](https://user-images.githubusercontent.com/867381/128309294-c6b0ff7d-83fb-4f3e-812b-f7688475388c.png)

2.  SQL statement

![image](https://user-images.githubusercontent.com/867381/128309408-e8f96034-4643-46aa-ba24-68877b168de3.png)

**Describe the feature you'd like:**

Jul 23, 2021 2:24 PM -> Jul 23, 2021 2:24 PM (+08:00)

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

SunRunAway:
@breeswish

breeswish:
This is a very good suggestion! @shhdgit Would you like to take a look?

shhdgit:
Yeah, sure. Display timezone is useful when communicating and troubleshooting.

PR #974, opened 2021-08-05
feat(ui): add timezone information

shhdgit:
Related: #973 . Add timezone information.

covered cases:
- table field
- tooltip

not covered:
- form value (unnecessary)
- chart value (should be shorter and clearer)

image

image

image


shhdgit:
/cc @baurine @breeswish 
PTAL, thanks~

breeswish (approved):

breeswish:
Good job!

PR #977, opened 2021-08-10
Fix endpoint scheme detection for TLS origination to PD

LINKIWI:
`NormalizePDEndPoint` is responsible for setting the URL scheme of the PD endpoint ([ref](https://github.com/pingcap/tidb-dashboard/blob/f6c50593b5de31cd2f2eeeaf9dbf33cf396cb4c2/pkg/config/config.go#L55)). However, it's currently called *before* `cfg.CoreConfig.ClusterTLSConfig` is set, which causes Dashboard to make plaintext HTTP requests to PD all the time, regardless of whether `--cluster-[ca|cert|key]` flags are set.

This change fixes the TLS origination behavior by running this logic *after* `cfg.CoreConfig.ClusterTLSConfig` is (optionally) populated by `buildTLSConfig`. This allows Dashboard to properly use HTTPS to PD when cluster TLS is enabled.

I didn't find any existing unit tests to cover this change, but if there's a test I can add or update to cover this, I'm happy to include it in this PR.

### DIFF
--- a/cmd/tidb-dashboard/main.go
+++ b/cmd/tidb-dashboard/main.go
@@ -98,9 +98,6 @@ func NewCLIConfig() *DashboardCLIConfig {
 	}
 
 	cfg.CoreConfig.NormalizePublicPathPrefix()
-	if err := cfg.CoreConfig.NormalizePDEndPoint(); err != nil {
-		log.Fatal("Invalid PD Endpoint", zap.Error(err))
-	}
 
 	// setup TLS config for TiDB components
 	if len(*clusterCaPath) != 0 && len(*clusterCertPath) != 0 && len(*clusterKeyPath) != 0 {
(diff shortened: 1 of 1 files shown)


bb7133:
Thanks for the contribution!

bb7133:
The change looks reasonable. I've tried TLS configuration manually and it did fix the issue.

bb7133 (approved):

bb7133:
LGTM

bb7133:
PTAL @xhebox and @breeswish

tisonkun (approved):
Thanks for your contribution @LINKIWI ! Changes LGTM.

For testing, it seems we must test the config logic as a whole since the order is significant. May with suggestion of @shhdgit & @breeswish we can push a follow up to add such test. I'd prefer avoid expanding a lot this PR since there is no test yet.

xhebox (approved):

shhdgit:
LGTM!

Sorry for the lack of tests. Additional tests on config initialization are welcome and may require some code restructuring. Improving test coverage is already in our roadmap.
Thanks again~

breeswish (approved):

PR #982, opened 2021-08-13
fix(stmt): min value of max stmt count size & blank detail page

shhdgit:
- Min of max stmt count size cannot be zero.
- Fix blank detail page bug caused by incorrect time range.


shhdgit:
/cc @breeswish @baurine 
PTAL, Thanks!

unbyte:
with more and more complex the cache keys seem really dirty.
just wonder could we find a more concise method for caching.

shhdgit:
I have tried to separate the cache. The other way around in this God controller would be much dirtier and we need to consider refactoring it.

The purpose of the current fix is to sync the data cache with the real time range. It needs to be fixed.

shhdgit:
Thanks for the advice~~ We do need to refactor the large controller for stmt and slow query, which makes it difficult to understand the intent of the code changes. T T

ui/lib/apps/Statement/utils/useStatementTableController.ts
```diff
@@ -141,9 +141,11 @@ export default function useStatementTableController(
     const { schemas, stmtTypes, searchText, timeRange } = queryOptions
     const cacheKey = `${schemas.join(',')}_${stmtTypes.join(
       ','
-    )}_${searchText}_${stringifyTimeRange(timeRange)}_${selectedFields}`
+    )}_${searchText}_${stringifyTimeRange(timeRange)}_${
```
breeswish, line 144: This just makes the cache ineffective: when returning back, the data may be changed. I think what you should do is to set the time picker to the actual memorized time when returning back.

What do you think @baurine

shhdgit: The logic of the time range calculation takes the nearest whole point or half point. For example, for recent 30 mins, 8:12 would be calculated as 8:00 - 8:30 time range, so the cache is still effective.

And the behavior is more in accordance with the "recent" - When we use the time range of the recent x minutes, the data should be the recent x minutes.

breeswish: I think how time range calculates doesn't matter. From a user perspective, user is supposed to always see a UI with corresponding data. With cache, we provides a feature that after user visits a record and goes back to the list view the user should see the list data exactly as the previous one. This two should be the primary principles that guides all behaviours.

breeswish: Under these two principles, after visiting a detail page and goes back, user should still seeing the previous list, in all cases, no matter when and how. And what we need to ensure is that all UI controls are correctly reflecting this list. For example, if this list is the 10:00 ~ 10:30 time range then the UI should indicate this range (if the last selected "recent XXX" no longer matches).

breeswish: As an alternative, it would also be great if we give user hints (while keeping the UI controls), like "This list is outdated. To see latest data, click "Refresh").

shhdgit: I got you. Refreshing data needs to be done manually.

And each response of the data from the server needs to be cached with a time range for detail plans queries. I need to recover this part of the code...

shhdgit: This would be great.

baurine: Agree with @breeswish , this change would make the cache ineffective, and it's better to show the memoried time range in the time picker when backing from the detail page.

And for the caching strategy, currently we implement it manually, not sure whether react-query or swr would help us simplify the logic.

shhdgit:
When the list data outdated it looks like this:

image

ui/lib/apps/Statement/pages/List/index.tsx
```diff
@@ -70,11 +70,29 @@ export default function StatementsOverview() {
...
+  const [showTimeRangeOutdatedTip, setShowTimeRangeOutdatedTip] = useState(
+    false
+  )
+  useEffect(() => {
+    let timer
+    if (isTimeRangeOutdated) {
+      setShowTimeRangeOutdatedTip(true)
+      timer = setTimeout(() => {
+        setShowTimeRangeOutdatedTip(false)
+      }, 3000)
```
breeswish, line 88: Maybe it would be simpler to just keep it, and remove it when refresh is clicked. Otherwise user will no longer see the tooltip any more?

ui/lib/apps/Statement/pages/List/index.tsx
```diff
@@ -181,11 +199,18 @@ export default function StatementsOverview() {
             )}
             {enable && (
               <Tooltip title={t('statement.pages.overview.toolbar.refresh')}>
-                {loadingStatements ? (
-                  <LoadingOutlined />
-                ) : (
-                  <ReloadOutlined onClick={refresh} />
-                )}
+                <Tooltip
```
breeswish, line 202: Will user just see 2 tooltips when hovering?

breeswish (commented):
BTW I wonder whether I'm the only person not understanding well about the logic of statements, statementControllers, and cache managers. :/

ui/lib/apps/Statement/pages/List/index.tsx
shhdgit: That's right, maybe I should hide the refresh one when the outdated tooltip is popped up.

shhdgit:
That's true. I can open another pr to refactor the code.

baurine (approved):
lgtm

ui/lib/apps/Statement/pages/List/index.tsx
```diff
@@ -234,3 +235,24 @@ export default function StatementsOverview() {
     </div>
   )
 }
+
+function RefreshTooltip({ isOutedated, children }) {
+  const { t } = useTranslation()
+  return isOutedated ? (
```
breeswish, line 241:
```suggestion
  return isOutdated ? (
```

breeswish (approved):

PR #983, opened 2021-08-17
feat(stmt): support query list limit

shhdgit:
image


breeswish (commented):
Will this limit affect sorting?

shhdgit:
Do we need to use server-side sorting?

breeswish:
If there is a limit, then it must be performed at server side. Otherwise you will get wrong results.

PR #984, opened 2021-08-23
Bump golangci to get the latest lints & fix lints

shhdgit:


.golangci.yml
```diff
@@ -10,6 +10,14 @@ linters-settings:
   golint:
     min-confidence: 0
 
+issues:
+  include:
```
breeswish, line 14: Is it possible to always include all?

pkg/apiserver/diagnose/report.go
```diff
@@ -2150,11 +2150,7 @@ func GetClusterHardwareInfoTable(startTime, endTime string, db *gorm.DB) (TableD
 		if !ok {
 			m[s] = &hardWare{s, map[string]int{row[1]: 1}, make(map[string]int), 0, make(map[string]float64), ""}
 		}
-		if _, ok := m[s].Type[row[1]]; ok {
```
breeswish, line 2153: Smart linter...

pkg/apiserver/logsearch/task.go
```diff
@@ -74,7 +74,8 @@ func (tg *TaskGroup) SyncRun() {
 
 	// Create log directory
 	dir := path.Join(tg.service.logStoreDirectory, strconv.Itoa(int(tg.model.ID)))
-	if err := os.MkdirAll(dir, 0777); err == nil {
+	err := os.MkdirAll(dir, 0777) // #nosec
```
breeswish, line 77: Why do we need a nosec for this checked error?

pkg/apiserver/logsearch/task.go
```diff
@@ -260,7 +261,7 @@ func (t *Task) searchLog(client diagnosticspb.DiagnosticsClient, targetType diag
 		}
 		for _, msg := range res.Messages {
 			line := logMessageToString(msg)
-			_, err := bufWriter.Write(*(*[]byte)(unsafe.Pointer(&line)))
+			_, err := bufWriter.Write(*(*[]byte)(unsafe.Pointer(&line))) // #nosec
```
breeswish, line 264: ditto, this error is checked and why do we need `nosec`?

breeswish (commented):
Generally great!

pkg/apiserver/logsearch/task.go
shhdgit: There's a least privilege issue: `Expect directory permissions to be 0750 or less`. I'm not sure about the perm in this case.

pkg/apiserver/logsearch/task.go
shhdgit: Every `unsafe` api invoke should use `nosec` tag... I'm wondering whether we need to turn on all gosec rules. Obviously, we know what we are doing when we use a specific api.

pkg/apiserver/diagnose/report.go
unbyte: すげー😲

.golangci.yml
shhdgit: I think it still makes sense for some low risk issues to be excluded. Is it possible to get the security team to audit low risk issues instead of add `gosec` tag?

PR #985, opened 2021-08-24
check required privileges for login sql user

baurine:
What did:
 
- Support non-root user login but it must have required privileges

Screenshots:

Lack required privileges:

image

Has required privileges:

screenshot-20210825-085044

TODO:
- remove root user limit for sso (will submit in a new PR after this PR is merged)


pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -9,7 +13,36 @@ import (
...
+//   ...
+//   "enable-sem": true/false,
+//   ...
+// },
+type tidbSecurityConfig struct {
+	Security tidbSEMConfig `json:"security"`
+}
+
+type tidbSEMConfig struct {
+	EnableSem bool `json:"enable-sem"`
```
breeswish, line 44: SEM enable status should be read from TiDB

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -50,6 +79,64 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+	grants := parseGrants(grantRows)
+	// 3. Check
+	// Following base privileges are required
+	// - ALL PRIVILEGES
+	// - or
+	// - PROCESS
+	// - SHOW DATABASES
+	// - SYSTEM_VARIABLES_ADMIN or SUPER (SUPER includes SYSTEM_VARIABLES_ADMIN)
+	// - CONFIG
+	// - DASHBOARD_CLIENT
```
breeswish, line 108: DASHBOARD_CLIENT is a sub privilege of SUPER.
I'm not sure whether CONFIG is a sub privilege of SUPER.

@crazycs520 Please provide necessary information.

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -9,7 +13,36 @@ import (
...
+	privAll             string = "ALL PRIVILEGES"
+	privProcess         string = "PROCESS"
+	privShowDB          string = "SHOW DATABASES"
+	privSystemVarAdmin  string = "SYSTEM_VARIABLES_ADMIN"
+	privSuper           string = "SUPER"
+	privConfig          string = "CONFIG"
+	privDashboardClient string = "DASHBOARD_CLIENT"
+	// Extra privileges when TiDB SEM is enabled
+	privRestrictedTablesAdmin string = "RESTRICTED_TABLES_ADMIN"
+	privRestrictedStatusAdmin string = "RESTRICTED_STATUS_ADMIN"
```
breeswish, line 29: Seems that these constants cannot be reused in other places. May be better to be not a constant.

ui/lib/client/translations/zh.yaml
```diff
@@ -7,6 +7,7 @@ error:
     user:
       signin:
         invalid_code: 授权码无效或已过期
+      miss_required_privileges: 此用户缺少访问 Dashboard 所需的权限
```
breeswish, line 10:
```suggestion
      miss_required_privileges: 此用户缺少访问 TiDB Dashboard 所需的权限
```

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -60,3 +147,22 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+	for _, row := range grantRows {
+		m := grantRegex.FindStringSubmatch(row)
+		if len(m) == 2 {
+			curRowGrants := strings.Split(m[1], ",")
+			grants = append(grants, curRowGrants...)
+		}
+	}
+
+	return grants
+}
```
breeswish, line 168: Please provide unit tests for this parser.

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -60,3 +147,22 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
 		IsWriteable:  true,
 	}, nil
 }
+
+// grantRows examples:
+// - GRANT PROCESS,SHOW DATABASES,CONFIG ON *.* TO 'dashboardAdmin'@'%'
+// - GRANT SYSTEM_VARIABLES_ADMIN,RESTRICTED_VARIABLES_ADMIN,RESTRICTED_STATUS_ADMIN,RESTRICTED_TABLES_ADMIN ON *.* TO 'dashboardAdmin'@'%'
+// - GRANT ALL PRIVILEGES ON *.* TO 'dashboardAdmin'@'%'
+func parseGrants(grantRows []string) []string {
```
breeswish, line 155: This parser can only parse global grants but not table grants. Prefer to indicate this feature in the name.

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -60,3 +147,22 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
 		IsWriteable:  true,
 	}, nil
 }
+
+// grantRows examples:
+// - GRANT PROCESS,SHOW DATABASES,CONFIG ON *.* TO 'dashboardAdmin'@'%'
+// - GRANT SYSTEM_VARIABLES_ADMIN,RESTRICTED_VARIABLES_ADMIN,RESTRICTED_STATUS_ADMIN,RESTRICTED_TABLES_ADMIN ON *.* TO 'dashboardAdmin'@'%'
```
breeswish, line 153: Please also check out what will happen when there are table grants, and role grants.

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -60,3 +147,22 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+// - GRANT ALL PRIVILEGES ON *.* TO 'dashboardAdmin'@'%'
+func parseGrants(grantRows []string) []string {
+	grants := make([]string, 0)
+	grantRegex := regexp.MustCompile(`GRANT (.+) ON`)
+
+	for _, row := range grantRows {
+		m := grantRegex.FindStringSubmatch(row)
+		if len(m) == 2 {
+			curRowGrants := strings.Split(m[1], ",")
```
breeswish, line 162: I feel highly unsafe about this process logic. We need to confirm how TiDB may possibly outputs these grants.

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -50,6 +79,64 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+			basePrivileges = basePrivileges | (1 << 1)
+		case privSystemVarAdmin:
+		case privSuper:
+			basePrivileges = basePrivileges | (1 << 2)
+		case privConfig:
+			basePrivileges = basePrivileges | (1 << 3)
+		case privDashboardClient:
+			basePrivileges = basePrivileges | (1 << 4)
+		case privAll:
+			basePrivileges = 0b11111
```
breeswish, line 128: How about just using:

```
if hasPriv(ALL PRIVILEGES) {
  pass
} else {
  if !hasPriv(PROCESS) {
    not pass
  }
  if ... {
    ...
  }
}
```

The golden rule to check whether your code is general enough or provides enough flexibility is to try to add something new to it, or removing something from it. For example, suppose we want a new privilege FOO to be checked in addition to current privileges, how many code do we need to change? how is the likelihood that something is forget to change? will we easily discover that something is forgotten?

By using this rule, it can be easily seen that:
1. During adding a new privilege, `privAll` is something easily to forget, and it must be changed accordingly, leaving a mistake possibility.
2. During removing a privilege, similar issues occurs. And what's worse, the bit shift order 1, 2, 3, 4, 5 needs to be changed.

A more flexible implementation should provide less and more integrated change for these future change scenarios. And as you can see, the simple `if else` works just better.

pkg/apiserver/user/sqlauth/sqlauth.go
baurine: OK

pkg/apiserver/user/sqlauth/sqlauth.go
baurine: @crazycs520 told me we can get it from TiDB `/config` HTTP API.

pkg/apiserver/user/sqlauth/sqlauth.go
baurine: OK, @crazycs520 please help check, currently, the logic is following the design doc.

pkg/apiserver/user/sqlauth/sqlauth.go
baurine: Ok

pkg/apiserver/user/sqlauth/sqlauth.go
baurine: Searched some documents, it seems there are 2 kinds of format to show grants for the current user:

- `GRANT [grants] ON [db.table] TO [user]`
- `GRANT [roles] TO [user]`

@crazycs520 please help confirm.

pkg/apiserver/user/sqlauth/sqlauth.go
crazycs520: Yes, we should get it from TiDB `config` HTTP API.

pkg/apiserver/user/sqlauth/sqlauth.go
crazycs520: Just as @baurine said, there are only 2 kinds of grant syntax.

pkg/apiserver/user/sqlauth/sqlauth.go
crazycs520: `CONFIG` is not a sub privilege of `SUPER`.

baurine: > `CONFIG` is not a sub privilege of `SUPER`.

What about the `DASHBOARD_CLIENT`?

pkg/apiserver/user/sqlauth/sqlauth.go
crazycs520: It's ok here. TiDB is using `,` to separate the privilege.

https://github.com/pingcap/tidb/blob/master/privilege/privileges/cache.go#L1322

pkg/apiserver/user/sqlauth/sqlauth.go
crazycs520: `DASHBOARD_CLIENT` also is not a sub privilege of `SUPER`.

breeswish: @crazycs520 DASHBOARD_CLIENT is a dynamic privilege, why is it not covered by SUPER?

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -50,6 +65,74 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+		} else if !hasPriv("SYSTEM_VARIABLES_ADMIN") && !hasPriv("SUPER") {
+			fullfillPrivileges = false
+		} else if !hasPriv("CONFIG") {
+			fullfillPrivileges = false
+		} else if !hasPriv("DASHBOARD_CLIENT") {
+			fullfillPrivileges = false
+		} else {
+			fullfillPrivileges = true
+		}
+	}
```
baurine, line 124: Also tried:

```
	fullfillPrivileges :=
		hasPriv("ALL PRIVILEGES") ||
			(hasPriv("PROCESS") &&
				hasPriv("SHOW DATABASES") &&
				(hasPriv("SYSTEM_VARIABLES_ADMIN") || hasPriv("SUPER")) &&
				hasPriv("CONFIG") &&
				hasPriv("DASHBOARD_CLIENT"))
```

pkg/apiserver/user/sqlauth/sqlauth.go
crazycs520: After confirm with @Morgan, `DASHBOARD_CLIENT` should be covered by SUPER

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -50,6 +65,80 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+	if config.Security.EnableSem {
+		fullfillPrivileges =
+			fullfillPrivileges &&
+				hasPriv("RESTRICTED_VARIABLES_ADMIN") &&
+				hasPriv("RESTRICTED_TABLES_ADMIN") &&
+				hasPriv("RESTRICTED_STATUS_ADMIN")
+	}
+	if !fullfillPrivileges {
+		return nil, user.ErrMissPrivileges.NewWithNoMessage()
+	}
```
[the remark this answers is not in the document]
baurine: Ok, let me try it.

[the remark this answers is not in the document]
baurine: Done!

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -60,3 +98,81 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+// - SHOW DATABASES
+// - CONFIG
+// - SYSTEM_VARIABLES_ADMIN or SUPER (SUPER includes SYSTEM_VARIABLES_ADMIN)
+// - DASHBOARD_CLIENT or SUPER (SUPER includes DASHBOARD_CLIENT)
+// When TiDB SEM is enabled, following extra privileges are required
+// - RESTRICTED_VARIABLES_ADMIN
+// - RESTRICTED_TABLES_ADMIN
+// - RESTRICTED_STATUS_ADMIN
+func checkDashboardPrivileges(grants []string, enableSEM bool) bool {
+	hasPriv := func(priv string) bool {
```
breeswish, line 138: How about using a Map to turn `O(N)` into `O(1)`? You can use [`funk.Map`](https://github.com/thoas/go-funk#funkmap) to map slice to map.

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -60,3 +98,81 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+		if !hasPriv("SUPER") {
+			if !hasPriv("SYSTEM_VARIABLES_ADMIN") {
+				return false
+			}
+			if !hasPriv("DASHBOARD_CLIENT") {
+				return false
+			}
+		}
+	}
+	if enableSEM {
```
breeswish, line 166: Does `ALL PRIVILEGES` contain the following `RESTRICTED_` privileges? @crazycs520

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -60,3 +98,81 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+		if !hasPriv("SHOW DATABASES") {
+			return false
+		}
+		if !hasPriv("CONFIG") {
+			return false
+		}
+		if !hasPriv("SUPER") {
+			if !hasPriv("SYSTEM_VARIABLES_ADMIN") {
+				return false
+			}
```
breeswish, line 160: `SYSTEM_VARIABLES_ADMIN` is no longer a required privilege (however when lacking such privileges it would be better to set `IsWriteable` to `false`.

ui/dashboardApp/layout/signin/index.tsx
```diff
@@ -226,7 +226,7 @@ function TiDBSignInForm({ successRoute, onClickAlternative }) {
             label={t('signin.form.username')}
             rules={[{ required: true }]}
           >
-            <Input onInput={clearErrorMsg} prefix={<UserOutlined />} disabled />
+            <Input onInput={clearErrorMsg} prefix={<UserOutlined />} />
```
breeswish, line 229: Before TiDB 5.4 or something maybe better to keep disabling this field to avoid confusing user, since in these versions user is just not possible to use other accounts, due to lacking the `DASHBOARD_` privilege.

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -50,6 +65,29 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
 	}
 	defer utils.CloseTiDBConnection(db) //nolint:errcheck
 
+	// Check dashboard privileges
+	// 1. Check whether TiDB SEM is enabled
+	resData, err := a.tidbClient.SendGetRequest("/config")
+	if err != nil {
+		return nil, user.ErrSignInOther.WrapWithNoMessage(err)
+	}
```
breeswish, line 73: Will this lead to confusing error messages when TiDB is down?

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -50,6 +65,29 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+	var config tidbSecurityConfig
+	err = json.Unmarshal(resData, &config)
```
breeswish, line 75: Will this cause issues when running against a TiDB that does not have SEM feature (like 4.0.x)?

breeswish:
@shhdgit PTAL, thanks!

pkg/apiserver/user/sqlauth/sqlauth.go
baurine: Good idea, let me try it.

shhdgit:
Generally LGTM, looking forward to new commits.

ui/dashboardApp/layout/signin/index.tsx
shhdgit: It looks like we have a long-term need for version comparison util.

breeswish: It can also be implemented by moving this logic to PD, e.g. Pass "allowNonRootLogin: false" in PD 4.0.x, 5.0.x, ...

pkg/apiserver/user/sqlauth/sqlauth.go
crazycs520: No, `ALL PRIVILEGES` doesn't contain the `RESTRICTED_xxx` privileges.

pkg/apiserver/user/sqlauth/sqlauth.go
baurine: Set `IsWriteable` to `false` will make the logged in SQL user can't do the following things except modify config:

- Authorize another SQL user to login by SSO: `endpoint.POST("/impersonation", auth.MWRequireWritePriv(), s.createImpersonationHandler)`
- Run any SQL in queryeditor: `endpoint.POST("/run", auth.MWRequireWritePriv(), s.runHandler)`

Is it expected? @breeswish

pkg/apiserver/user/sqlauth/sqlauth.go
baurine: I have tested it by the following code:

```
	var config tidbSecurityConfig
	_ = json.Unmarshal([]byte(""), &config)
	fmt.Println("sem:", config.Security.EnableSem)
```

It will output:

```
sem: false
```

pkg/apiserver/user/sqlauth/sqlauth.go
baurine: Let me check.

baurine: Add `time.Sleep(10*time.Second)` before requesting `/config`, and kill the TiDB after clicking the `Sign In` button, it reports `Sign in failed: authenticate failed, cause: error.api.user.signin.other: error.tidb.no_alive_tidb` error message, I think it is clear enough.

image

baurine: Btw, in the RP #992 , this part logic is refactored (extracted for reusing), the `error.tidb.no_alive_tidb` error is returned directly, finally it only show `error.tidb.no_alive_tidb` error here.

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -50,13 +66,129 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+	return grants
+}
+
+// Following base privileges are required
+// - ALL PRIVILEGES
+// - or
+// - PROCESS
+// - SHOW DATABASES
+// - CONFIG
+// - SYSTEM_VARIABLES_ADMIN or SUPER (SUPER includes SYSTEM_VARIABLES_ADMIN)
```
crazycs520, line 132:
```suggestion
```

ui/dashboardApp/layout/signin/index.tsx
baurine: How about implementing it in an individual PR.

My thought:

1. Add `allow-non-root-login` option likes `experimental`  for dashboard 
2. If PD check allow non root login, it pass allow-non-root-login as true to start dashboard, else false. So we don't need to care about how PD implements the check logic.

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -176,3 +174,21 @@ func checkDashboardPrivileges(grants []string, enableSEM bool) bool {
...
 }
+
+func checkWriteablePriv(grants []string) bool {
+	privsMap := funk.Map(grants, func(priv string) (string, bool) {
+		return priv, true
+	}).(map[string]bool)
+	hasPriv := func(priv string) bool {
+		_, ok := privsMap[priv]
+		return ok
+	}
```
breeswish, line 184: How about returning grants as a map from the beginning? As you can see you have multiple places transforming from array to map.

baurine: I have thought that, although changing grants to map in the method parameter reduces some duplicated logic, but the map type is not that reasonable for the 2 methods, because just the keys of the map are meaningful for them, we can feel that when we write unit tests for them.

But I only insist on this when they only are duplicated twice.

breeswish: If you want a set, maybe you can use `map[string]struct{}`. This is a common pattern to make a Set instead of a Map in golang.

baurine: Done!

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -50,13 +66,125 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+		m := grantRegex.FindStringSubmatch(row)
+		if len(m) == 2 {
+			curRowGrants := strings.Split(m[1], ",")
+			grants = append(grants, curRowGrants...)
+		}
+	}
+
+	privsSet := funk.Map(grants, func(priv string) (string, struct{}) {
+		return priv, struct{}{}
+	}).(map[string]struct{})
```
baurine, line 125: wait, I don't need this method, I shoud make `grants` as map[string]struct{}`.

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -50,13 +67,123 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+	if enableSEM {
+		if !hasPriv("RESTRICTED_VARIABLES_ADMIN") {
+			return false
+		}
+		if !hasPriv("RESTRICTED_TABLES_ADMIN") {
+			return false
+		}
+		if !hasPriv("RESTRICTED_TABLES_ADMIN") {
+			return false
+		}
```
breeswish, line 171: How about reorg this part to the first? Then the overall indention can be reduced:

```go
if enableSEM {
	// Note: When SEM is enabled, these additional privileges need to be checked even if "ALL PRIVILEGES" is granted.
	if !hasPriv("RESTRICTED_VARIABLES_ADMIN") {
		return false
	}
	if !hasPriv("RESTRICTED_TABLES_ADMIN") {
		return false
	}
	if !hasPriv("RESTRICTED_TABLES_ADMIN") {
		return false
	}
}

if hasPriv("ALL PRIVILEGES") {
	// ALL PRIVILEGES contains privileges below. If it is set, privilege requirement is met.
	return true
}

if !hasPriv("PROCESS") {
	return false
}
if !hasPriv("SHOW DATABASES") {
	return false
}
if !hasPriv("CONFIG") {
	return false
}

if hasPriv("SUPER") {
	// SUPER contains privileges below. If it is set, privilege requirement is met.
	return true
}

if !hasPriv("DASHBOARD_CLIENT") {
	return false
}
```

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -11,6 +15,23 @@ import (
...
+// },
+type tidbSecurityConfig struct {
+	Security tidbSEMConfig `json:"security"`
+}
+
+type tidbSEMConfig struct {
+	EnableSem bool `json:"enable-sem"`
+}
+
+type set map[string]struct{}
```
breeswish, line 33: Maybe this doesn't need a type alias, as it is a common pattern seen everywhere. Adding a new type simply makes people harder to understand what it is.

pkg/apiserver/user/sqlauth/sqlauth.go
breeswish: According to Golang's naming convention it should `EnableSEM`

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -50,13 +67,123 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+
+// Currently, There are 2 kinds of grant output format in TiDB:
+// - GRANT [grants] ON [db.table] TO [user]
+// - GRANT [roles] TO [user]
+// Examples:
+// - GRANT PROCESS,SHOW DATABASES,CONFIG ON *.* TO 'dashboardAdmin'@'%'
+// - GRANT SYSTEM_VARIABLES_ADMIN,RESTRICTED_VARIABLES_ADMIN,RESTRICTED_STATUS_ADMIN,RESTRICTED_TABLES_ADMIN ON *.* TO 'dashboardAdmin'@'%'
+// - GRANT ALL PRIVILEGES ON *.* TO 'dashboardAdmin'@'%'
+// - GRANT `app_read`@`%` TO `test`@`%`
+func parseCurUserGrants(grantRows []string) set {
```
breeswish, line 112: Maybe it can be called `parseUserGrants`, as it can parse grants for not only current user, but for any users, as long as it is a user grant.

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -50,13 +67,123 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+	grants := set{}
+	grantRegex := regexp.MustCompile(`GRANT (.+) ON`)
```
breeswish, line 114: It is recommended to move this `MustCompile` to the outside, as a variable, so that:

1. It is not repeatedly compiled.
2. An regex mistake can be discovered early, i.e. resulting in a panic at TiDB Dashboard starting phase, instead of panic when `parseCurUserGrants` is called.

pkg/apiserver/user/sqlauth/sqlauth.go
```diff
@@ -50,13 +67,123 @@ func (a *Authenticator) Authenticate(f user.AuthenticateForm) (*utils.SessionUse
...
+// - DASHBOARD_CLIENT or SUPER (SUPER includes DASHBOARD_CLIENT)
+// When TiDB SEM is enabled, following extra privileges are required
+// - RESTRICTED_VARIABLES_ADMIN
+// - RESTRICTED_TABLES_ADMIN
+// - RESTRICTED_STATUS_ADMIN
+func checkDashboardPriv(privs set, enableSEM bool) bool {
+	hasPriv := func(priv string) bool {
+		_, ok := privs[priv]
+		return ok
+	}
```
breeswish, line 144: How about moving this to a new function called `func hasPriv(privs ...) bool`, as you are using it in multiple places?

ui/dashboardApp/layout/signin/index.tsx
breeswish: How would this `If PD check allow non root login` being implemented? I think there should always be a place hard coding what kind of versions are allowed to do non root login, right? As it is a product feature, not a user provided choice.

ui/lib/client/translations/zh.yaml
```diff
@@ -7,6 +7,7 @@ error:
     user:
       signin:
         invalid_code: 授权码无效或已过期
+      lack_required_privileges: 此用户缺少访问 TiDB Dashboard 所需的权限
```
breeswish, line 10: I think we should provide some better guidelines for what kind of privileges user need. Maybe better to leave a TODO and add a link to the documentation later.

ui/lib/client/translations/en.yaml
```diff
@@ -7,6 +7,7 @@ error:
     user:
       signin:
         invalid_code: Authorization Code is invalid or expired
+      lack_required_privileges: The User lacks required privileges to access dashboard
```
breeswish, line 10:
```suggestion
      insufficient_privileges: The user does not have sufficient privileges to access TiDB Dashboard
```

pkg/apiserver/user/auth.go
```diff
@@ -35,6 +35,7 @@ import (
 var (
 	ErrNS                  = errorx.NewNamespace("error.api.user")
 	ErrUnsupportedAuthType = ErrNS.NewType("unsupported_auth_type")
+	ErrLackPrivileges      = ErrNS.NewType("lack_required_privileges")
```
breeswish, line 38: How about calling it insufficient privileges? As our privilege rule is complicated and some privileges are required only in some cases.

```suggestion
	ErrInsufficientPriv = ErrNS.NewType("insufficient_privileges")
```

ui/lib/client/translations/zh.yaml
```diff
@@ -7,6 +7,7 @@ error:
...
     slow_query:
       export_no_data: 没有可导出的慢查询日志
     statement:
```
breeswish, line 13: Who added this previously! 🤣 It is a wrong place for them.. LOL

breeswish (commented):
The rest LGTM. @crazycs520 @shhdgit PTAL

ui/lib/client/translations/zh.yaml
baurine: Let me fix it.

ui/lib/client/translations/zh.yaml
baurine: Yep, @crazycs520 has discussed it with me, we should explain what privileges it needs in the doc.

pkg/apiserver/user/sqlauth/sqlauth.go
baurine: Done.

pkg/apiserver/user/sqlauth/sqlauth.go
baurine: Done.

pkg/apiserver/user/sqlauth/sqlauth.go
baurine: Done.

pkg/apiserver/user/sqlauth/sqlauth.go
baurine: Done.

pkg/apiserver/user/sqlauth/sqlauth.go
baurine: Done.

ui/lib/client/translations/zh.yaml
baurine: Done.

baurine:
All comments are addressed, PTAL, thanks!

cmd/tidb-dashboard/main.go
```diff
@@ -74,6 +74,7 @@ func NewCLIConfig() *DashboardCLIConfig {
 	flag.StringVar(&cfg.CoreConfig.PDEndPoint, "pd", cfg.CoreConfig.PDEndPoint, "PD endpoint address that Dashboard Server connects to")
 	flag.BoolVar(&cfg.CoreConfig.EnableTelemetry, "telemetry", cfg.CoreConfig.EnableTelemetry, "allow telemetry")
 	flag.BoolVar(&cfg.CoreConfig.EnableExperimental, "experimental", cfg.CoreConfig.EnableExperimental, "allow experimental features")
+	flag.BoolVar(&cfg.CoreConfig.EnableNonRootLogin, "non-root-login", cfg.CoreConfig.EnableNonRootLogin, "allow non root sql user login")
```
shhdgit, line 77: Why does `non-root-login` become a cmd option? It seems to depend on the tidb version.

ui/dashboardApp/layout/signin/index.tsx
baurine: @shhdgit see this discussion about the `non-root-login` option

ui/lib/client/translations/zh.yaml
```diff
@@ -7,6 +7,8 @@ error:
     user:
       signin:
         invalid_code: 授权码无效或已过期
+      # TODO: add doc link that explains what privileges are needed
+      insufficient_privileges: 此用户缺少访问 {{distro.tidb}} Dashboard 所需的权限
```
breeswish, line 11: Nice catch! We do need a distro here.

breeswish (approved):
LGTM

PR #992, opened 2021-08-29
Support non root for sso login

baurine:
What did:
- Support modify authorization at any time
- Support authorize to non-root user to login by sso

Need to wait for PR #985 to be merged.

Demo:

image

image

image

image


baurine:
Close it because of PR #1003

PR #994, opened 2021-09-01
Tweak: component invalid address check

shhdgit:
We need to check if the component address is valid to avoid potential security issues.


shhdgit:
/cc @baurine @breeswish 

PTAL, thanks~

baurine (approved):
lgtm

pkg/tiflash/client.go
```diff
@@ -79,6 +94,40 @@ func (c *Client) SendGetRequest(host string, statusPort int, relativeURI string)
...
+	err := c.checkValidAddress(fmt.Sprintf("%s:%d", host, statusPort))
+	if err != nil {
+		return nil, err
+	}
+
 	uri := fmt.Sprintf("%s://%s:%d%s", c.httpScheme, host, statusPort, relativeURI)
 	return c.httpClient.WithTimeout(c.timeout).SendRequest(c.lifecycleCtx, uri, http.MethodPost, body, ErrFlashClientRequestFailed, distro.Data("tiflash"))
 }
+
+func (c *Client) getMemebers() ([]topology.StoreInfo, error) {
```
breeswish, line 106:
```suggestion
func (c *Client) getMembers() ([]topology.StoreInfo, error) {
```

pkg/tiflash/client.go
```diff
@@ -79,6 +94,40 @@ func (c *Client) SendGetRequest(host string, statusPort int, relativeURI string)
...
+	cached, _ := c.cache.Get("tiflash_members")
```
breeswish, line 107: Seems that each get resets the TTL according to the [doc](https://github.com/ReneKroon/ttlcache/). Maybe this is not a desired behaviour.

pkg/tikv/client.go
```diff
@@ -79,6 +94,40 @@ func (c *Client) SendGetRequest(host string, statusPort int, relativeURI string)
...
 	return c.httpClient.WithTimeout(c.timeout).SendRequest(c.lifecycleCtx, uri, http.MethodPost, body, ErrTiKVClientRequestFailed, distro.Data("tikv"))
 }
+
+func (c *Client) getMemebers() ([]topology.StoreInfo, error) {
+	cached, _ := c.cache.Get("tikv_members")
+	if cached != nil {
+		return cached.([]topology.StoreInfo), nil
+	}
+
+	mem, _, err := topology.FetchStoreTopology(c.pdClient)
```
breeswish, line 112: Only TiKV stores should be allowed.

pkg/tiflash/client.go
```diff
@@ -79,6 +94,40 @@ func (c *Client) SendGetRequest(host string, statusPort int, relativeURI string)
...
+	if cached != nil {
+		return cached.([]topology.StoreInfo), nil
+	}
+
+	_, mem, err := topology.FetchStoreTopology(c.pdClient)
```
breeswish, line 112: Only TiFlash stores should be allowed.

pkg/pd/client.go
```diff
@@ -38,18 +44,22 @@ const (
...
 	timeout      time.Duration
+	cache        *ttlcache.Cache
 }
 
 func NewPDClient(lc fx.Lifecycle, httpClient *httpc.Client, config *config.Config) *Client {
 	client := &Client{
 		httpClient:   httpClient,
 		httpScheme:   config.GetClusterHTTPScheme(),
 		baseURL:      config.PDEndPoint,
+		specificURL:  "",
```
breeswish, line 59: What is this hack?

pkg/tiflash/client.go
shhdgit: That's right. I should set `SkipTTLExtensionOnHit` to true.

pkg/pd/client.go
shhdgit: Updated. This seems like an unnecessary hack.

pkg/tiflash/client.go
shhdgit: What does this mean? And I updated the variable name. PTAL~

pkg/tidb/client.go
```diff
@@ -66,22 +67,28 @@ func NewTiDBClient(lc fx.Lifecycle, config *config.Config, etcdClient *clientv3.
...
 		sqlAPITLSKey:             sqlAPITLSKey,
 		sqlAPIAddress:            "",
-		cache:                    ttlcache.NewCache(),
+		cache:                    cache,
 	}
 
 	lc.Append(fx.Hook{
 		OnStart: func(ctx context.Context) error {
 			client.lifecycleCtx = ctx
+			statusProxyIP, statusProxyPort, _ := host.ParseHostAndPortFromAddressURL("http://" + client.forwarder.statusProxy.listener.Addr().String())
```
breeswish, line 89: Will this cause problems under TLS?

breeswish (approved):

pkg/tidb/client.go
shhdgit: Nope, but I found it unnecessary to parse the `net.Addr`.

PR #1003, opened 2021-09-06
support authorize non root user to sso login

baurine:
What did:
- Support modify authorization at any time
- Support authorize to non-root user to login by sso

Demo:

image

image

image

image

### DIFF
--- a/pkg/apiserver/user/sso/service.go
+++ b/pkg/apiserver/user/sso/service.go
@@ -189,7 +189,7 @@ func (s *Service) newSessionFromImpersonation(userInfo *oAuthUserInfo, idToken s
 				_ = s.updateImpersonationStatus(userName, ImpersonateStatusAuthFail)
 				return nil, ErrInvalidImpersonateCredential.Wrap(err, "Invalid SQL credential")
 			}
-			if errorx.IsOfType(err, user.ErrInsufficientPriv) {
+			if errorx.IsOfType(err, user.ErrInsufficientPrivs) {
 				_ = s.updateImpersonationStatus(userName, ImpersonateStatusInsufficientPrivs)
 				return nil, ErrInvalidImpersonateCredential.Wrap(err, "Insufficient privileges")
 			}
@@ -217,6 +217,9 @@ func (s *Service) createImpersonation(userName string, password string) (*SSOImp
 		if err != nil {
 			if errorx.IsOfType(err, tidb.ErrTiDBAuthFailed) {
 				return nil, ErrInvalidImpersonateCredential.Wrap(err, "Invalid SQL credential")
+			}
+			if errorx.IsOfType(err, user.ErrInsufficientPrivs) {
+				return nil, ErrInvalidImpersonateCredential.Wrap(err, "Insufficient privileges")
 			}
 			return nil, err
 		}
--- a/pkg/apiserver/user/sso/service.go
+++ b/pkg/apiserver/user/sso/service.go
@@ -181,23 +181,22 @@ func (s *Service) newSessionFromImpersonation(userInfo *oAuthUserInfo, idToken s
 	if err != nil {
 		return nil, err
 	}
-	{
-		// Check whether this user can access dashboard
-		err := user.VerifySQLUser(s.params.TiDBClient, userName, password)
-		if err != nil {
-			if errorx.IsOfType(err, tidb.ErrTiDBAuthFailed) {
-				_ = s.updateImpersonationStatus(userName, ImpersonateStatusAuthFail)
-				return nil, ErrInvalidImpersonateCredential.Wrap(err, "Invalid SQL credential")
-			}
-			if errorx.IsOfType(err, user.ErrInsufficientPrivs) {
-				_ = s.updateImpersonationStatus(userName, ImpersonateStatusInsufficientPrivs)
-				return nil, ErrInvalidImpersonateCredential.Wrap(err, "Insufficient privileges")
-			}
-			return nil, err
-		}
 
-		_ = s.updateImpersonationStatus(userName, ImpersonateStatusSuccess)
+	// Check whether this user can access dashboard
+	writeable, err := user.VerifySQLUser(s.params.TiDBClient, userName, password)
+	if err != nil {
+		if errorx.IsOfType(err, tidb.ErrTiDBAuthFailed) {
+			_ = s.updateImpersonationStatus(userName, ImpersonateStatusAuthFail)
+			return nil, ErrInvalidImpersonateCredential.Wrap(err, "Invalid SQL credential")
+		}
+		if errorx.IsOfType(err, user.ErrInsufficientPrivs) {
+			_ = s.updateImpersonationStatus(userName, ImpersonateStatusInsufficientPrivs)
+			return nil, ErrInvalidImpersonateCredential.Wrap(err, "Insufficient privileges")
+		}
+		return nil, err
 	}
+	_ = s.updateImpersonationStatus(userName, ImpersonateStatusSuccess)
+
 	return &utils.SessionUser{
 		Version:      utils.SessionVersion,
 		HasTiDBAuth:  true,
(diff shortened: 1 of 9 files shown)


pkg/apiserver/user/sso/service.go
```diff
@@ -279,9 +283,6 @@ func (s *Service) discoverOIDC(issuer string) (*oidcWellKnownConfig, error) {
 		return nil, ErrDiscoverFailed.Wrap(err, "Failed to discover OIDC endpoints")
 	}
 	wellKnownConfig := resp.Result().(*oidcWellKnownConfig)
-	if wellKnownConfig.Issuer != issuer {
-		return nil, ErrDiscoverFailed.New("Issuer did not match in the OIDC provider, expect %s, got %s", issuer, wellKnownConfig.Issuer)
-	}
```
baurine, line 284: // TODO: add an explanation about why remove it.

pkg/apiserver/user/sso/service.go
```diff
@@ -235,11 +234,16 @@ func (s *Service) createImpersonation(user string, password string) (*SSOImperso
...
 		EncryptedPass:         encryptedInHex,
 		LastImpersonateStatus: nil,
 	}
-	err = s.params.LocalStore.Clauses(clause.Insert{Modifier: "OR REPLACE"}).Create(&record).Error
+	// currently, we only support to authorize one sql user
+	err = s.revokeAllImpersonations()
+	if err != nil {
+		return nil, err
+	}
+	err = s.params.LocalStore.Create(&record).Error
```
breeswish, line 246: This is not concurrent safe.

breeswish (commented):
The rest LGTM

pkg/apiserver/user/sso/service.go
breeswish: Yeah, why do we remove it?

baurine: Reasons:

1. For auth0, the `issuer` filed has `/` suffix in the `http://[org].us.auth0.com/.well-known/openid-configuration` response:

   ```json
   {
     issuer: "https://[org].us.auth0.com/",
     authorization_endpoint: "https://[org].us.auth0.com/authorize",
     token_endpoint: "https://[org].us.auth0.com/oauth/token",
     ...
   }
   ```

   For okta, the `issuer` filed has no `/` suffix in the `http://[org].okta.com/.well-known/openid-configuration` response:

   ```json
   {
     issuer: "https://dev-9991837.okta.com",
     authorization_endpoint: "https://dev-9991837.okta.com/oauth2/v1/authorize",
     token_endpoint: "https://dev-9991837.okta.com/oauth2/v1/token",
     ...
   }
   ```

   With this check, the user needs to use different rules to fill the **OIDC Discovery URL**, it confuses the user.

2. Remove this check, it still works (I have tested), even the user only fill the **OIDC Discovery URL** with `[org].us.auth0.com` or `[org].okta.com`, no `http://` or `https://` prefix and `/` suffix.

    Btw, I think this check is a little contradicting, we use an issuer value, get the correct `openid-configuration` response, but the original value is not exactly the same as the value in the response, then we tell the user input the wrong value. If the original issuer value is wrong, it should not get the correct `openid-configuration` response. So I think the check is redundant, is it right?

pkg/apiserver/user/sso/service.go
baurine: OK, let me have a look.

pkg/apiserver/user/sso/service.go
breeswish: I think we should keep this check (and add more checks) according to OIDC specification:

https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig

OpenID Provider Configuration Validation

If any of the validation procedures defined in this specification fail, any operations requiring the information that failed to correctly validate MUST be aborted and the information that failed to validate MUST NOT be used.

The issuer value returned MUST be identical to the Issuer URL that was directly used to retrieve the configuration information. This MUST also be identical to the iss Claim value in ID Tokens issued from this Issuer.

pkg/apiserver/user/sso/service.go
baurine: Add mutex.

pkg/apiserver/user/sso/service.go
baurine: Ok, let me have a look.

baurine: Revert and add more checks.

pkg/apiserver/user/sso/service.go
```diff
@@ -254,16 +264,21 @@ func (s *Service) revokeAllImpersonations() error {
...
-	EndSessionURL string `json:"end_session_endpoint"`
+	Issuer                           string   `json:"issuer"`
+	AuthURL                          string   `json:"authorization_endpoint"`
+	TokenURL                         string   `json:"token_endpoint"`
+	UserInfoURL                      string   `json:"userinfo_endpoint"`
+	EndSessionURL                    string   `json:"end_session_endpoint"`
+	JWKSURI                          string   `json:"jwks_uri"`
+	ResponseTypesSupported           []string `json:"response_types_supported"`
+	SubjectTypesSupported            []string `json:"subject_types_supported"`
+	IDTokenSigningAlgValuesSupported []string `json:"id_token_signing_alg_values_supported"`
```
baurine, line 275: These fields are required according to the OIDC specification. https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata

breeswish (commented):
Generally lgtm. @shhdgit PTAL, thanks

pkg/apiserver/user/sso/service.go
breeswish: Could you help verify that newly added checks works for Okta? We have production users using Okta and when we adding more checks we need to make sure not breaking them.

pkg/apiserver/user/sso/service.go
```diff
@@ -273,23 +288,35 @@ func (s *Service) discoverOIDC(issuer string) (*oidcWellKnownConfig, error) {
...
 		return nil, ErrDiscoverFailed.Wrap(err, "Failed to discover OIDC endpoints")
 	}
 	wellKnownConfig := resp.Result().(*oidcWellKnownConfig)
-	if wellKnownConfig.Issuer != issuer {
+	if strings.TrimSuffix(wellKnownConfig.Issuer, "/") != issuer {
 		return nil, ErrDiscoverFailed.New("Issuer did not match in the OIDC provider, expect %s, got %s", issuer, wellKnownConfig.Issuer)
 	}
 	if len(wellKnownConfig.TokenURL) == 0 {
-		return nil, ErrDiscoverFailed.New("TokenURL is not provided in the OIDC provider")
+		return nil, ErrDiscoverFailed.New("token_endpoint is not provided in the OIDC provider")
```
shhdgit, line 301: That's we validate many fields that must be implemented in the specification, `OpenID Provider Configuration Validation` only mentions `issuer url` and `iss Claim`?

And what is the `iss Claim value in ID Tokens issued`

pkg/apiserver/user/sso/service.go
baurine: Yep, both checked.

pkg/apiserver/user/sso/service.go
baurine: https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata mentions the required metadata.

breeswish (approved):

pkg/apiserver/user/sso/service.go
shhdgit: https://github.com/coreos/go-oidc
https://connect2id.com/blog/how-to-validate-an-openid-connect-id-token

It looks like we should verify the issuer and issuer claims from id tokens in the exchange phase, not in the discovery phase.

shhdgit: Maybe we need to do the verification here:

https://github.com/pingcap/tidb-dashboard/blob/master/pkg/apiserver/user/sso/service.go?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L362

PR #1006, opened 2021-09-15
fix(stmt): evict & unbound schema stmt query

shhdgit:
- fix evict & unbound schema stmt query to avoid blank detail page
- add evict stmt tips

Another issue that can cause blank detail page is #982. PTAL, thanks~


shhdgit:
/cc @baurine @breeswish

PTAL, Thanks~

pkg/apiserver/statement/queries.go
```diff
@@ -137,15 +137,24 @@ func (s *Service) queryPlans(
...
-		Where("digest = ?", digest).
-		Group("plan_digest").
-		Find(&result).
-		Error
+		Group("plan_digest")
+
+	if schemaName != "" {
+		query.Where("schema_name = ?", schemaName)
+	}
+	if digest == "" {
```
breeswish, line 149: So frontend will pass `digest=""` to the backend, even if the digest from the backend is null?

unbyte: Go: 🤧

shhdgit: Yeah, there are two things we need to improve:
- Define a new structure. Embed `sql.NullString` to scan NULL value from db and customize the MashallJSON/UnmashallJSON interface to serde null value
- The get request is not easy to carry null, we can use undefined to not pass this parameter

But I don't think that's going to help here. I reverted the code, LOL.

shhdgit: Because `is null` and `= ''` are always two patterns.

pkg/apiserver/statement/queries.go
```diff
@@ -168,12 +177,21 @@ func (s *Service) queryPlanDetail(
...
-		Where("schema_name = ?", schemaName).
-		Where("digest = ?", digest)
+		Where("summary_begin_time >= FROM_UNIXTIME(?) AND summary_end_time <= FROM_UNIXTIME(?)", beginTime, endTime)
+
 	if len(plans) > 0 {
 		query = query.Where("plan_digest in (?)", plans)
 	}
+	if schemaName != "" {
+		query.Where("schema_name = ?", schemaName)
+	}
```
breeswish, line 187: If `digest == ""` (if the record is "other record"), then there is no need to check plan digest or schema name I think. This can be a special branch.

ui/lib/apps/Statement/utils/useStatementTableController.ts
```diff
@@ -244,8 +244,20 @@ export default function useStatementTableController(
...
+        data.forEach((d) => {
+          if (d.digest !== '') {
+            return
+          }
+          d.digest = evictTips
+          d.digest_text = evictTips
+        })
+
+        setStatements(data)
+        cacheMgr?.set(cacheKey, data)
```
breeswish, line 260: Could you provide a screen shot about how it looks like for "Others" in the list page and in the detail page? Also maybe it is better to display this row as:

*(Others)* (gray)

tooltip: All of other dropped SQL statements

shhdgit: Yeah, I need to refine it, hold on.

shhdgit:
Updated, PTAL~ @breeswish 

Screen Shot 2021-09-26 at 22 15 44

ui/lib/apps/Statement/utils/useStatementTableController.ts
shhdgit: Should we show the detail page when user click evicted data?

breeswish (approved):

PR #1009, opened 2021-09-23
Refine non root login

baurine:
What did:

- Make non-root-login start argument default value false
- Add tooltip to ask for upgrade TiDB to support login with non root users

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,4 +46,4 @@ endif
 	go build -o bin/tidb-dashboard -ldflags '$(LDFLAGS)' -tags "${BUILD_TAGS}" cmd/tidb-dashboard/main.go
 
 run:
-	bin/tidb-dashboard --debug --experimental
+	bin/tidb-dashboard --debug --experimental --non-root-login
--- a/ui/dashboardApp/layout/translations/en.yaml
+++ b/ui/dashboardApp/layout/translations/en.yaml
@@ -4,6 +4,7 @@ signin:
     success: Sign in successfully
   form:
     username: Username
+    username_tooltip: You can login with any users by upgrading to TiDB 5.3 and higher versions
     password: Password
     button: Sign In
     tidb_auth:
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -46,4 +46,4 @@ endif
 	go build -o bin/tidb-dashboard -ldflags '$(LDFLAGS)' -tags "${BUILD_TAGS}" cmd/tidb-dashboard/main.go
 
 run:
-	bin/tidb-dashboard --debug --experimental
+	bin/tidb-dashboard --debug --experimental --non-root-login
```
unbyte, line 49: How about supporting using env var to enable/disable such features.

ui/dashboardApp/layout/translations/zh.yaml
```diff
@@ -4,6 +4,7 @@ signin:
     success: 登录成功
   form:
     username: 用户名
+    username_tooltip: 升级到 TiDB 5.3 及更高版本可使用任意用户登录
```
breeswish, line 7:
```suggestion
    username_tooltip: 升级到 TiDB 5.3 及更高版本后可自定义登录用户
```

ui/dashboardApp/layout/translations/en.yaml
```diff
@@ -4,6 +4,7 @@ signin:
     success: Sign in successfully
   form:
     username: Username
+    username_tooltip: You can login with any users by upgrading to TiDB 5.3 and higher versions
```
breeswish, line 7:
```suggestion
    username_tooltip: Sign in user can be customized in TiDB 5.3 or later versions
```

breeswish (approved):

shhdgit:
LGTM.

shhdgit (approved):

Makefile
shhdgit: cmd options is convenient as the switch. But like https://github.com/pingcap/tidb-dashboard/pull/985#discussion_r702553685 says here `non-root-login` is not required to be provided by the user, we're using it as a feature switch, and it's `tidb-dashboard` *library* switch.

baurine: Yep, it could be better if we can support that.

PR #1017, opened 2021-10-06
make code match comments

gregwebs:
This proposed change is not tested


baurine:
hi @gregwebs , thanks for your fix, because you don't sign the CLA, so I can't merge your PR, I will submit another PR to fix this silly error and fix the test as well soon.

PR #1020, opened 2021-10-09
fix bug of context is nil

crazycs520:


shhdgit:
Will these clients' be instantiated and invoked before the `OnStart` lifecycle of the `fx`?

breeswish:
Close for #1021

PR #1023, opened 2021-10-18
tweak(stmt): add more statement fields

shhdgit:
Add Statement fields:
 - max_latency
 - min_latency
 - max_result_rows
 - min_result_rows
 - avg_result_rows

Need to verify in tidb master branch:
image

### DIFF
--- a/pkg/apiserver/statement/models.go
+++ b/pkg/apiserver/statement/models.go
@@ -40,6 +40,9 @@ type Model struct {
 	AggMaxLatency            int    `json:"max_latency" agg:"MAX(max_latency)"`
 	AggMinLatency            int    `json:"min_latency" agg:"MIN(min_latency)"`
 	AggAvgLatency            int    `json:"avg_latency" agg:"CAST(SUM(exec_count * avg_latency) / SUM(exec_count) AS SIGNED)"`
+	AggMaxResultRows         int    `json:"max_result_rows" agg:"MAX(max_result_rows)"`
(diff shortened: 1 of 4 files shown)


lilyjazz:
From the request, users need to rank the min/max latency as they wish.
I think we have no question about this request.

So, from the perspective of display and ease of use, we can make some optimizations to the existing display.
image

lilyjazz:
> Min/Max/Mean should be displayed using a single column

@breeswish Are there some showcases?
I personally think this is not a good idea, which may increase the cost of interface display and user learning.

shhdgit:
About padding, I got two ways to optimize:
1. add border, which is ugly I think
image

2. change the direction of the data and the chart
image

breeswish:
> > Min/Max/Mean should be displayed using a single column
> 
> @breeswish Are there some showcases? I personally think this is not a good idea, which may increase the cost of interface display and user learning.

@lilyjazz Oh you have a GitHub account! :)

I'm thinking that what kind of user needs to distinguish min/max/mean. Currently the column is named as "Mean Xxx", and the progress bar is displaying the mean, and the sorting is performed over mean as well. The min and max are displayed as extra information along with the column. (and if user doesn't noticed about the yellow bar then there is no studying cost for them). We could supply more columns. But it might lead to harder choices. You know, the more options you provide, the harder choice it is for users. Considering that most of the columns have MIN/MAX/MEAN, this could be horrible. I would like to avoid this kind of learning cost as much as possible.

breeswish:

PR #1026, opened 2021-10-19
upgrade gin-jwt version to 2.7.0

### LINKED ISSUE #1024: upgrade gin-jwt version  to  2.7.0 to  get greater security
##  don't keep use dgrijalva/jwt-go

There are many  security issue from dgrijalva/jwt-go, which is not maintained, we should switch to community maintained golang-jwt/jwt
the gin-jwt v2.6.3  dependent  on dgrijalva/jwt-go  v3.2.0   but   gin-jwt v2.7.0 has been upgraded to golang-jwt/jwt/v4
image

so,we should upgrade gin-jwt version  to  2.7.0  to  get greater security

zhijian-pro:
for close #1024 
amazing ,this is the 1024th issue !


breeswish:
@zhijian-pro Hi, the build process is failed. Could you take a look and fix it?

PR #1028, opened 2021-10-28
refine(slow_query): use virtual view to build select stmt

shhdgit:
- Introducing virtual view util. (Statement should also be refined via virtual view after virtual view supporting sql like `CAST(.. AS SIGNED) ans COLUMN`)
- Fix the issue comes from AskTUG: https://asktug.com/t/topic/183588


breeswish:

PR #1031, opened 2021-10-31
fix legacy group all finish state for manually instance profiling

baurine:


pkg/apiserver/profiling/router.go
```diff
@@ -144,6 +144,22 @@ func (s *Service) getGroupDetail(c *gin.Context) {
...
+				finishedTasks++
+			}
+		}
+		if finishedTasks == len(tasks) {
+			taskGroup.State = TaskAllFinish
+		} else {
+			taskGroup.State = TaskPartialFinish
+		}
+		s.params.LocalStore.Save(taskGroup)
+	}
```
breeswish, line 161: Is it possible to do modification in other places? It is a bad practice to modify something in a GET, as you will face these problems:

1. Is this GET API still isomorphic?
2. What about concurrent GETs? Will there be concurrent issues?
3. It violates the "GET" restful semantic.

pkg/apiserver/profiling/model.go
```diff
@@ -27,10 +27,11 @@ type TaskState int
...
 const (
-	TaskStateError TaskState = iota
-	TaskStateRunning
-	TaskStateFinish
-	TaskPartialFinish
+	TaskStateError    TaskState = iota // 0
+	TaskStateRunning                   // 1
+	TaskStateFinish                    // 2, legacy state for group all success
+	TaskPartialFinish                  // 3
+	TaskAllFinish                      // 4
```
breeswish, line 34: Unified name:

```suggestion
	TaskStatePartialFinish                  // 3
	TaskStateAllFinish                      // 4
```

ui/lib/apps/InstanceProfiling/translations/en.yaml
```diff
@@ -23,6 +23,7 @@ instance_profiling:
         finished: Finished
         failed: Failed
         partial_finished: Partial Finished
+        legacy_all_finished: Partial Or All Finished
```
breeswish, line 26: Actually this description might not be correct.

Consider a case that all sub tasks are failed. Previously this will fall into "Finished". But it is not partial finish at all. None is finished.

What we can only for sure is that, in such case, there is at least one failure.

ui/lib/apps/InstanceProfiling/pages/List.tsx
```diff
@@ -153,6 +155,14 @@ export default function Page() {
                 )}
               />
             )
+          } else {
+            // new all success
+            return (
+              <Badge
+                status="success"
+                text={t('instance_profiling.list.table.status.finished')}
```
breeswish, line 163: As a success, then name it and display it as success.

pkg/apiserver/profiling/router.go
baurine: Yep, you're right, I have considered writing a script to modify the legacy state, let me think more.

ui/lib/apps/InstanceProfiling/translations/en.yaml
```diff
@@ -23,6 +23,7 @@ instance_profiling:
         finished: Finished
         failed: Failed
         partial_finished: Partial Finished
```
breeswish, line 25: Maybe better to call it "Partially Failed"

pkg/apiserver/profiling/router.go
breeswish: If migrating old data is complex then maybe we can just leave it. We can look forward and know that the improvement is still good for new profilings, which is good enough.

breeswish:
It's already very late at night! Take your time :)

I think it's fine to improve it later, as this is pure UX improvements.

pkg/apiserver/profiling/router.go
baurine: Yep, it is a little complex, I think it is fine to keep the old data. Let me close this PR and move some refinements in this PR to a new PR with other refinements for conprof.

baurine:
submit the new PR for this issue: https://github.com/pingcap/tidb-dashboard/pull/1039

PR #1039, opened 2021-11-02
migrate manually profiling task state

baurine:
What did: 

- Refine the group task state by `all failed`, `partially failed`, `all success` instead only one state `finished`
- Migrate the legacy state 
- Fix the wrong `running` state if the dashboard crashes


baurine:
Convert to draft, let me try to migrate the old data.

pkg/apiserver/profiling/service.go
```diff
@@ -87,11 +87,15 @@ var newService = fx.Provide(func(lc fx.Lifecycle, p ServiceParams, fts *fetchers
...
-			s.wg.Add(1)
+			s.wg.Add(2)
 			go func() {
 				defer s.wg.Done()
 				s.serviceLoop(ctx)
 			}()
+			go func() {
+				defer s.wg.Done()
+				s.migrateLegacyState(ctx)
+			}()
```
baurine, line 98: I think we can do the migration work at this moment. How do you think? @breeswish

baurine:
@breeswish PTAL, thanks!

pkg/apiserver/profiling/service.go
```diff
@@ -145,6 +149,43 @@ func (s *Service) serviceLoop(ctx context.Context) {
...
+		// Step 1: find out group tasks whose state is TaskStateFinish
+		err := s.params.LocalStore.Where("state = ?", TaskStateFinish).Order("id DESC").Limit(100).Find(&groupTasks).Error
+		if err != nil || len(groupTasks) == 0 {
+			break
+		}
+
+		for _, groupTask := range groupTasks {
+			// Step 2: find out all child tasks of the group task
+			err = s.params.LocalStore.Where("task_group_id = ?", groupTask.ID).Find(&tasks).Error
+			if err != nil || len(tasks) == 0 {
```
baurine, line 166: This condition should never happen, but if it happens, may cause the outer loop to become a dead loop, still consider how to avoid it.

baurine: Add way: Add `.Where("id < ?", lastMinID")` condition when querying group tasks.

baurine:
Haven't considered the compatibility for degrading TiDB Dashboard, will revert the migration logic.

pkg/apiserver/profiling/service.go
```diff
@@ -145,6 +149,54 @@ func (s *Service) serviceLoop(ctx context.Context) {
...
+				}
+			}
+			if successTasks == 0 {
+				groupTask.State = TaskStateFailed
+			} else if successTasks < len(tasks) {
+				groupTask.State = TaskStatePartialFailed
+			} else {
+				groupTask.State = TaskStateAllSuccess
+			}
+			s.params.LocalStore.Save(groupTask)
```
baurine, line 195: This way migration can't handle the situation when down-grade TiDB version, for example from TiDB 5.3.0 back to TiDB 5.1.2.

I have a new idea.

pkg/apiserver/profiling/model.go
```diff
@@ -25,12 +25,26 @@ import (
...
 type TaskState int
 
+type GroupTaskState int
+
 // Built-in task state
+// State change flow:
+// For single task: TaskStateRunning --> TaskStateFailed (fail) | TaskStateFinish (success)
+// For group task:
+// - Old: TaskStateRunning --> TaskStateFinish
+// - New: GroupTaskStateRunning --> GroupTaskStateAllFailed | GropuTaskStatePartialFailed | GroupTaskStateAllSuccess (all success)
```
baurine, line 35: Todo: remove `(all success)`

baurine:
Remain work: fix running state tasks.

baurine:
The PR is ready to review, PTAL, @breeswish @shhdgit , thanks!

Ref design doc (internal): https://docs.google.com/document/d/1O4j8H04GJ8Iyg3jVPjQliRZhUfQ4CWHJHvrJhdFKh6Y/edit#

pkg/apiserver/profiling/model.go
```diff
@@ -25,12 +25,26 @@ import (
...
 // Built-in task state
+// State change flow:
+// For single task: TaskStateRunning --> TaskStateFailed (fail) | TaskStateFinish (success)
+// For group task:
+// - Old: TaskStateRunning      --> TaskStateFinish
+// - New: GroupTaskStateRunning --> GroupTaskStateAllFailed | GropuTaskStatePartialFailed | GroupTaskStateAllSuccess
+const (
+	TaskStateFailed  TaskState = iota // 0
+	TaskStateRunning                  // 1
+	TaskStateFinish                   // 2, for single task, it means success; for group task, it only means all child tasks are done, no matter they are success or failed, it will be refined by GroupTaskState
```
breeswish, line 39:
```suggestion
	TaskStateFinished                   // 2
```

pkg/apiserver/profiling/model.go
```diff
@@ -25,12 +25,26 @@ import (
...
+)
+
 const (
-	TaskStateError TaskState = iota
-	TaskStateRunning
-	TaskStateFinish
-	TaskPartialFinish
+	GroupTaskStateUnknown       GroupTaskState = iota // 0
+	GroupTaskStateRunning                             // 1
```
breeswish, line 44: How about introducing `GroupTaskStateFinished = 2 // deprecated`? In this way, you can safely change GroupTask's state to the type `GroupTask` while maintaining compatibility with the old data.

pkg/apiserver/profiling/model.go
```diff
@@ -49,7 +63,8 @@ func (TaskModel) TableName() string {
 
 type TaskGroupModel struct {
 	ID                  uint                          `json:"id" gorm:"primary_key"`
-	State               TaskState                     `json:"state" gorm:"index"`
+	State               TaskState                     `json:"state" gorm:"index"` // legacy state, works before version 5.3.0
```
breeswish, line 66: This can be safely changed to:

```suggestion
	State               GroupTaskState                     `json:"state" gorm:"index"` // legacy state, works before version 5.3.0
```

as long as you introduce `GroupTaskStateFinished = 2`.

BTW there is no "version 5.3.0" concept in TiDB Dashboard. Please use TiDB Dashboard's release version.

pkg/apiserver/profiling/service.go
breeswish: How about move it to L78? As you can see in L78 there is already a "autoMigrate" process, which does similar things.

pkg/apiserver/profiling/model.go
```diff
@@ -49,7 +65,8 @@ func (TaskModel) TableName() string {
 
 type TaskGroupModel struct {
 	ID                  uint                          `json:"id" gorm:"primary_key"`
-	State               TaskState                     `json:"state" gorm:"index"`
+	State               GroupTaskState                `json:"state" gorm:"index"` // legacy state, works before version v2021.11.01.1
+	NewState            GroupTaskState                `json:"new_state" gorm:"index"`
```
shhdgit, line 69: Do we really need `NewState`? As [design docs comments](https://docs.google.com/document/d/1O4j8H04GJ8Iyg3jVPjQliRZhUfQ4CWHJHvrJhdFKh6Y/edit?disco=AAAARp92vmQ) said, I'm worried that having two such similar fields, `State` and `NewState`, will confuse developers.

pkg/apiserver/profiling/service.go
baurine: OK, I thought making it run as a goroutine can reduce the start time, but it is not necessary.

pkg/apiserver/profiling/model.go
baurine: Yep, I agree with you, but it seems only this way can handle the situation when down-grading.

baurine:
btw, I feel using numbers as the state may not be a good idea. If we insert a new state in the middle, all the following numbers meaning are changed, how about using string?

breeswish:
I am working on an architectural refinement for the profiling. This will be resolved by the way.

PR #1043, opened 2021-11-02
skip grant check when skip-grant-table enable

baurine:
Currently, if TiDB enables the `security.skip-grant-table`, running `show grants` and some other SQL statements will get errors, so we temporarily skip the grants check if `skip-grant-table` is enabled.

### DIFF
--- a/pkg/apiserver/user/verify_sql_user.go
+++ b/pkg/apiserver/user/verify_sql_user.go
@@ -38,7 +38,8 @@ type tidbSecurityConfig struct {
 }
 
 type tidbSEMConfig struct {
-	EnableSEM bool `json:"enable-sem"`
+	EnableSEM      bool `json:"enable-sem"`
+	SkipGrantTable bool `json:"skip-grant-table"`
 }
 
 func VerifySQLUser(tidbClient *tidb.Client, userName, password string) (writeable bool, err error) {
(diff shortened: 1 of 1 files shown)


pkg/apiserver/user/verify_sql_user.go
```diff
@@ -59,14 +60,20 @@ func VerifySQLUser(tidbClient *tidb.Client, userName, password string) (writeabl
 	if err != nil {
 		return false, err
 	}
-	// 2. Get grants
+	// 2. Check SkipGrantTable
+	// Note: Currently, if TiDB enable the skip-grant-table, running `show grants` will get error.
+	// So this is a workaround before the above bug is fixed.
+	if config.Security.SkipGrantTable {
+		return true, nil
```
baurine, line 67: In this situation, should the user has the write privilege?

crazycs520: I think the user can have the write privilege.

crazycs520 (commented):
LGTM

crazycs520 (approved):

crazycs520:
@breeswish or @shhdgit PTAL

pkg/apiserver/user/verify_sql_user.go
```diff
@@ -59,14 +60,20 @@ func VerifySQLUser(tidbClient *tidb.Client, userName, password string) (writeabl
 	if err != nil {
 		return false, err
 	}
-	// 2. Get grants
+	// 2. Check SkipGrantTable
+	// Note: Currently, if TiDB enable the skip-grant-table, running `show grants` will get error.
+	// So this is a workaround before the above bug is fixed.
```
shhdgit, line 65: Does this mean that the non-root user has more privileges in this case? What is the impact on the tidb-dashboard?

crazycs520: yes, if someone is connected to the tidb-server which `skip-grant-table=true`, he can do anything

shhdgit (approved):

PR #1044, opened 2021-11-02
more refines

baurine:
What did:

- Make the left side menu active for `instance_profiling` and `continuous_profiling` after refreshing
- Fix the profile retention duration select options
- Remember the last login username

Screenshots:

image

### DIFF
--- a/ui/dashboardApp/layout/main/Sider/index.tsx
+++ b/ui/dashboardApp/layout/main/Sider/index.tsx
@@ -140,15 +140,7 @@ function Sider({
     if (defaultCollapsed) {
       return []
     } else {
-      let openKeys = ['debug', 'experimental']
-      const urlHash = window.location.hash.substring(2)
-      if (
-        urlHash.startsWith('instance_profiling') ||
-        urlHash.startsWith('continuous_profiling')
-      ) {
-        openKeys.push('profiling')
-      }
-      return openKeys
+      return ['debug', 'experimental', 'profiling']
     }
   }, [defaultCollapsed])
 
--- a/ui/dashboardApp/layout/main/Sider/index.tsx
+++ b/ui/dashboardApp/layout/main/Sider/index.tsx
@@ -140,7 +140,15 @@ function Sider({
     if (defaultCollapsed) {
       return []
     } else {
-      return ['debug', 'experimental']
+      let openKeys = ['debug', 'experimental']
+      const urlHash = window.location.hash.substring(2)
+      if (
+        urlHash.startsWith('instance_profiling') ||
+        urlHash.startsWith('continuous_profiling')
+      ) {
+        openKeys.push('profiling')
+      }
+      return openKeys
     }
   }, [defaultCollapsed])
 
--- a/ui/lib/apps/ContinuousProfiling/pages/ConProfSettingForm.tsx
+++ b/ui/lib/apps/ContinuousProfiling/pages/ConProfSettingForm.tsx
@@ -11,6 +11,7 @@ import {
 } from 'antd'
 import { ExclamationCircleOutlined } from '@ant-design/icons'
 import { useTranslation } from 'react-i18next'
+import { TFunction } from 'i18next'
 import { getValueFormat } from '@baurine/grafana-value-formats'
 
 import client, {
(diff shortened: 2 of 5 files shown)


ui/dashboardApp/layout/main/Sider/index.tsx
```diff
@@ -140,7 +140,15 @@ function Sider({
...
-      return ['debug', 'experimental']
+      let openKeys = ['debug', 'experimental']
+      const urlHash = window.location.hash.substring(2)
+      if (
+        urlHash.startsWith('instance_profiling') ||
+        urlHash.startsWith('continuous_profiling')
+      ) {
+        openKeys.push('profiling')
+      }
+      return openKeys
```
breeswish, line 151: Is it possible to just open everything? In this way we can reduce the amount of these hard codings.

ui/lib/apps/ContinuousProfiling/pages/ConProfSettingForm.tsx
```diff
@@ -21,12 +22,24 @@ import { useClientRequest } from '@lib/utils/useClientRequest'
...
+  // in our case, the seconds value must be the multiple of one day seconds
+  if (seconds % ONE_DAY_SECONDS !== 0) {
+    console.warn(`${seconds} is not the mulitple of one day seconds`)
+  }
+  const day = seconds / ONE_DAY_SECONDS
+  return t('continuous_profiling.settings.profile_retention_duration_option', {
+    d: day,
+  })
+}
+
```
breeswish, line 42: How did we format them in the time selector? In time selector seems that we don't need to explicitly write xxx days as a translation key.

breeswish (commented):
Mostly looks good

ui/dashboardApp/layout/main/Sider/index.tsx
baurine: Yep, sure, it's much simpler, just `return ['debug', 'experimental', 'profiling']`, I have tested. I was just afraid it makes the left side menu too long.

ui/lib/apps/ContinuousProfiling/pages/ConProfSettingForm.tsx
baurine: Yep, because we happen to show 7/14/28 days in the time selector, if we choose to show 10 days, it will display `1 week, 3 day` as here.

image

```
const RECENT_SECONDS = [
  // ...
  2 * 24 * 60 * 60,
  3 * 24 * 60 * 60,

  7 * 24 * 60 * 60,
  14 * 24 * 60 * 60,
  28 * 24 * 60 * 60,
]
```

ui/dashboardApp/layout/main/Sider/index.tsx
baurine: Oh, you mean we can remove `debug`, `experimental` these hard codes as well?

ui/lib/apps/ContinuousProfiling/pages/ConProfSettingForm.tsx
```diff
@@ -21,12 +22,24 @@ import { useClientRequest } from '@lib/utils/useClientRequest'
...
 
+const ONE_DAY_SECONDS = 24 * 60 * 60
 const RETENTION_SECONDS = [
-  3 * 24 * 60 * 60,
-  5 * 24 * 60 * 60,
-  10 * 24 * 60 * 60,
+  3 * ONE_DAY_SECONDS,
+  5 * ONE_DAY_SECONDS,
+  10 * ONE_DAY_SECONDS,
 ]
```
breeswish, line 30: Considering that our translation key only support "day", how about we write `RETENTION_DAYS` here directly? Then you don't need to worry about the case that `second % DAYS != 0`.

breeswish (approved):

ui/lib/apps/ContinuousProfiling/pages/ConProfSettingForm.tsx
baurine: Because in the API, the `data_retention_seconds` filed value uses the `second` as the unit.

breeswish: How about:

```jsx
const RETENTION_DAYS = [3, 5, 10];

// ....

{RETENTION_DAYS.map((day) => (
  <Select.Option key={day} value={val * ONE_DAY_SECONDS}>
    {t('xxxxx', { d: day })}
  </Select.Option>
))}
```

baurine: Looks good, let me update it in the next PR.

PR #1047, opened 2021-11-04
Compitable with different TiDB versions for conprof and non-root-login features

baurine:
What did:

- Move conprof codes out of profiling
- Check whether conprof and nonrootlogin features are supported in the current version
- Make frontend adjustment for frontend when feature is not enabled

Screenshots:

- Hide "Continuous Profiling" menu if it is not enabled
  image


baurine:
The CI says build frontend failed and need to run `yarn fmt`, I did it and it changed much code, a little strange why this happen, will check it later.

ui/lib/client/translations/zh.yaml
```diff
@@ -19,6 +19,7 @@ error:
       ng_monitoring_not_start: |
         本功能存储组件离线，请尝试使用 TiUP Cluster 对 monitor 节点进行 reload 操作。
         若未能解决问题，请联系技术支持。
+    feature_not_enabled: 此功能未在当前版本开启。
```
breeswish, line 22: Maybe better to call it available, e.g. 当前版本的集群不支持或无法使用该功能，请联系技术支持了解详细情况。

pkg/apiserver/info/info.go
```diff
@@ -73,6 +77,8 @@ func (s *Service) infoHandler(c *gin.Context) {
 		Version:            version.GetInfo(),
 		EnableTelemetry:    s.params.Config.EnableTelemetry,
 		EnableExperimental: s.params.Config.EnableExperimental,
+		EnableConprof:      conprof.IsFeatureEnable(s.params.Config),
+		EnableNonRootLogin: nonrootlogin.IsFeatureEnable(s.params.Config),
```
breeswish, line 81: How about call them support instead of enable? "Enable" sounds like they can be enabled. However in fact they cannot.

For example:

```js
SupportConprof: ...
SupportNonRootLogin: ...
```

Also, considering that we may have more things to add in the support matrix, can we achieve this?

```js
SupportedFeatures: ["conprof", "nonRootLogin", ...]
```

breeswish (commented):
@shhdgit PTAL

breeswish:
> The CI says build frontend failed and need to run `yarn fmt`, I did it and it changed much code, a little strange why this happen, will check it later.

prettier version has been updated in https://github.com/pingcap/tidb-dashboard/commit/c33af34565070c6219f3df72e2980ce1f31760bd  you can run `yarn` to update your local prettier. Then you should get the same format results as in CI.

pkg/apiserver/info/info.go
baurine: OK

pkg/config/config.go
```diff
@@ -38,7 +38,7 @@ type Config struct {
 
 	EnableTelemetry    bool
 	EnableExperimental bool
-	EnableNonRootLogin bool
```
baurine, line 41: After removing this config item, the code in the PD side (PR: https://github.com/tikv/pd/pull/4141/files) doesn't work anymore. So when we submit a new release to PD, we need to remove that line code in PD. Just a reminder.

pkg/apiserver/apiserver.go
```diff
@@ -151,7 +153,7 @@ func (s *Service) Start(ctx context.Context) error {
 			info.RegisterRouter,
 			clusterinfo.RegisterRouter,
 			profiling.RegisterRouter,
-			profiling.RegisterConprofRouter,
+			conprof.RegisterRouter,
```
shhdgit, line 156: Maybe RegisterRouter can also be declared in the `conprof.Module`, like this:
```
var Module = fx.Options(
	fx.Provide(newService),
	fx.Invoke(registerRouter),
)
```

pkg/apiserver/conprof/feature_enable.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/pingcap/tidb-dashboard/pkg/apiserver/utils"
+	"github.com/pingcap/tidb-dashboard/pkg/config"
+)
+
+var (
+	supportedTiDBVersions = []string{">= 5.3.0"}
+	featureEnabled        *bool
+)
+
+func IsFeatureEnable(config *config.Config) (enabled bool) {
```
shhdgit, line 26: Considering that there are many features that require a feature flag, how about abstracting this into a util. e.g.
```
type FeatureFlag struct {
	supportedTiDBVersions []string
	featureEnabled                *bool
}

func (ff *FeatureFlag) IsFeatureEnable(targetVersion string) bool
func (ff *FeatureFlag) Middleware(targetVersion string) gin.HandlerFunc
```

And we can declare feature flags in module.go/mod.go(in the future)
```
var FeatureFlag = util.NewFeatureFlag([]string{">= 5.3.0"})
var Module = fx...
```

pkg/apiserver/apiserver.go
baurine: Let me try.

pkg/apiserver/conprof/feature_enable.go
shhdgit: LGTM, and I will do this refinement in https://github.com/pingcap/tidb-dashboard/pull/1035 .

baurine:
All comments are addressed, PTAL, thanks! @breeswish @shhdgit

shhdgit (approved):

PR #1050, opened 2021-11-08
fix(*): http client ssrf

shhdgit:
We need to check if the component address is valid to avoid potential ssrf issues.

ref: https://github.com/pingcap/tidb-dashboard/pull/1038


shhdgit:
/cc @baurine @breeswish 
PTAL, thanks~

baurine (commented):
lgtm

pkg/httpc/cache.go
```diff
@@ -0,0 +1,77 @@
...
+			log.Warn("Http client func cache failed", zap.Error(err))
+		}
+
+		return returns
+	}
+
+	// Make function of the input function type
+	newFn := reflect.MakeFunc(fn.Type(), cacheWrap)
+	return newFn.Interface()
+}
```
breeswish, line 77: This function is a typical example that it is too complex to be correct - no reviewers can actually ensure that it is ok in all cases by looking at its code. the guard of the correctness is the unit test, while as we all know that unit tests cannot cover all cases.

pkg/httpc/client.go
```diff
@@ -113,22 +109,27 @@ func (c *Client) Send(
...
-	if !(resp.StatusCode >= 200 && resp.StatusCode < 300) {
+	var data []byte
+	if !c.isRawBody {
+		data, err = ioutil.ReadAll(resp.Body)
 		defer resp.Body.Close()
-		data, _ := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			e := errType.Wrap(err, "Failed to read %s API response", errOriginComponent)
+			log.Warn("SendRequest failed", zap.String("uri", uri), zap.Error(err))
+			return nil, e
```
breeswish, line 119: When status code is not ok, it is possible that you cannot read the body I think.

pkg/apiserver/debugapi/http_client.go
```diff
@@ -81,43 +81,63 @@ type tidbImplement struct {
 	Client *tidb.Client
 }
 
-func (impl *tidbImplement) Get(payload *endpoint.ResolvedRequestPayload) (*httpc.Response, error) {
-	return impl.Client.
+func (impl *tidbImplement) Get(payload *endpoint.ResolvedRequestPayload) (*http.Response, error) {
```
breeswish, line 84: Why do we have `http.Response` sometimes and `httpc.Response` sometimes? Which one should I use in different cases?

pkg/httpc/client.go
```diff
@@ -113,22 +109,27 @@ func (c *Client) Send(
...
 
 type Response struct {
-	*http.Response
-}
-
-func (r *Response) Body() ([]byte, error) {
-	defer r.Response.Body.Close()
-	return ioutil.ReadAll(r.Response.Body)
+	RawResponse *http.Response
+	Body        []byte
```
breeswish, line 134: This API design of worse compared to the previous one. As an example,

suppose you see a piece of code (actually everywhere in the code):

```
...(resp.Body)
```

Can you tell out what kind of `resp.Body` it is? And how can we identify its correctness? You simply can't. Since there is ambiguity between `http.Response.Body` and `httpc.Response.Body`. Not to mention that `http.Response` and `httpc.Response` is passed everywhere (e.g. `resp, err := xxxx()`), which causes a new ambiguity.

pkg/httpc/client.go
shhdgit:
```
	// The http Client and Transport guarantee that Body is always
	// non-nil, even on responses without a body or responses with
	// a zero-length body. It is the caller's responsibility to
	// close Body. The default HTTP client's Transport may not
	// reuse HTTP/1.x "keep-alive" TCP connections if the Body is
	// not read to completion and closed.
```

I found some comments about Response.Body, and we've read the body when the status code is not ok before.

pkg/httpc/client.go
shhdgit: Yes! I should unify the response type.

pkg/httpc/client.go
```diff
@@ -102,22 +98,44 @@ func (c *Client) Send(
...
 		defer resp.Body.Close()
-		data, _ := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			e := errType.Wrap(err, "Failed to read %s API response", errOriginComponent)
+			log.Warn("SendRequest failed", zap.String("uri", uri), zap.Error(err))
+			return nil, e
+		}
+	}
+
+	if !(resp.StatusCode >= 200 && resp.StatusCode < 300) {
```
breeswish, line 112: When isRawBody is set, `string(data)` no longer works.

pkg/httpc/client.go
```diff
@@ -102,22 +98,44 @@ func (c *Client) Send(
...
+func (r *Response) Body() []byte {
+	if r.RawResponse == nil {
+		return []byte{}
+	}
+	return r.body
+}
+
+// RawBody method exposes the HTTP raw response body. Use this method in-conjunction with `WithRawBody()`
+// function otherwise you get an error as `read err: http: read on closed response body`.
+func (r *Response) RawBody() io.ReadCloser {
```
breeswish, line 136: Seems that the change of this file's API doesn't reduce the complexity, but make the calling more complex:

Previous design:

| Action                  | Function Call before Send | Function Call for Response |
|-------------------------|---------------------------|----------------------------|
| Access Raw Body         | No need.                  | #.Body                     |
| Access Body as Bytes    | No need.                  | #.Body()                   |
| Access Response headers | No need.                  | #.Headers                  |

New design:

| Action                  | Function Call before Send | Function Call for Response |
|-------------------------|---------------------------|----------------------------|
| Access Raw Body         | WithRawBody()             | #.RawBody()                |
| Access Body as Bytes    | No need.                  | #.Body()                   |
| Access Response headers | No need.                  | #.Response.Headers         |

There is also possibilities that people can now call `WithRawBody()` + `Body()` that will just lead to empty data, or `RawBody()` directly that will lead to `nil`. None of these can be ensured by the compiler: more possibilities to make mistakes.

What do you want to achieve by this change?

pkg/tiflash/client.go
```diff
@@ -29,21 +31,32 @@ type Client struct {
 	httpScheme   string
 	lifecycleCtx context.Context
 	timeout      time.Duration
+	isRawBody    bool
+	getEndpoints func() (map[string]struct{}, error)
 }
 
-func NewTiFlashClient(lc fx.Lifecycle, httpClient *httpc.Client, config *config.Config) *Client {
+func NewTiFlashClient(lc fx.Lifecycle, httpClient *httpc.Client, pdClient *pd.Client, config *config.Config) *Client {
+	cache := pd.NewEndpointCache()
```
breeswish, line 39: It is not reasonable to rely on PD for TiFlash.

pkg/tikv/client.go
```diff
@@ -29,21 +31,32 @@ type Client struct {
...
-func NewTiKVClient(lc fx.Lifecycle, httpClient *httpc.Client, config *config.Config) *Client {
+func NewTiKVClient(lc fx.Lifecycle, httpClient *httpc.Client, pdClient *pd.Client, config *config.Config) *Client {
+	cache := pd.NewEndpointCache()
 	client := &Client{
 		httpClient:   httpClient,
 		httpScheme:   config.GetClusterHTTPScheme(),
 		lifecycleCtx: nil,
 		timeout:      defaultTiKVStatusAPITimeout,
+		getEndpoints: func() (map[string]struct{}, error) {
+			return cache.Func("tikv_endpoints", func() (map[string]struct{}, error) {
```
breeswish, line 46: I guess you will find your code logic even more simpler without using the TTLCache, as:

1. You will only store one item
2. You want to have strong type checks

pkg/pd/api.go
```diff
@@ -0,0 +1,52 @@
...
+
+type InfoMember struct {
+	GitHash       string   `json:"git_hash"`
+	ClientUrls    []string `json:"client_urls"`
+	DeployPath    string   `json:"deploy_path"`
+	BinaryVersion string   `json:"binary_version"`
+	MemberID      uint64   `json:"member_id"`
+}
+
+func FetchMembers(c *Client) (*InfoMembers, error) {
```
breeswish, line 40: This should be called UnsafeFetchMembers, as it bypasses the target address restriction.

pkg/utils/topology/pd.go
```diff
@@ -23,25 +23,10 @@ func FetchPDTopology(pdClient *pd.Client) ([]PDInfo, error) {
 		return nil, err
 	}
 
-	data, err := pdClient.SendGetRequest("/members")
+	ds, err := pd.FetchMembers(pdClient)
```
breeswish, line 26: `FetchPDTopology` should be called `UnsafeFetchPDTopology` as it bypasses the target address restriction.

breeswish (commented):
One thing makes me very uncomfortable is that, places where "UnsafeXxx" is provided is strongly bind with the internal implementation (cause we need unsafeXxx to retrieve PD members). It doesn't look good when internal implementation causes external API changes.

pkg/httpc/client.go
shhdgit: Yes, you're right.
I wanted to unify the previous interfaces `Send` and `SendRequest`. So I learned the deisgn from go-resty and introduced a risk at the same time, which is not worth it.

Maybe renaming `SendRequest` to `SendAndReadBody` would work.

shhdgit:
Got it, a PR that does many things at once is not a good PR...

pkg/httpc/client.go
breeswish: Yes. The most uncomfortable part I fell when using go-resty is that I need to **pair** the raw request and the raw response otherwise it just won't work.

breeswish:

PR #1057, opened 2021-11-11
Refactor feature flag to support more modules

shhdgit:
ref https://github.com/pingcap/tidb-dashboard/pull/1047#discussion_r744251814


shhdgit:
/cc @baurine @breeswish 
PTAL, thanks~

pkg/apiserver/conprof/service.go
```diff
@@ -81,7 +81,7 @@ func newService(lc fx.Lifecycle, p ServiceParams) *Service {
 func registerRouter(r *gin.RouterGroup, auth *user.AuthService, s *Service) {
 	endpoint := r.Group("/continuous_profiling")
 
-	endpoint.Use(utils.MWForbidByFeatureSupport(IsFeatureSupport(s.params.Config)))
+	endpoint.Use(utils.MWForbidByFeatureSupport(FeatureFlagConprof.IsSupport(s.params.Config.FeatureVersion)))
```
breeswish, line 84: This is too long.

pkg/utils/feature_flag.go
```diff
@@ -21,10 +22,31 @@ import (
...
+	supported := isVersionSupport(targetVersion, ff.supportedVersions)
+	ff.supported = &supported
+	return supported
+}
+
 // IsVersionSupport checks if a semantic version fits within a set of constraints
 // pdVersion, standaloneVersion examples: "v5.2.2", "v5.3.0", "v5.4.0-alpha-xxx", "5.3.0" (semver can handle `v` prefix by itself)
 // constraints examples: "~5.2.2", ">= 5.3.0", see semver docs to get more information
-func IsVersionSupport(standaloneVersion string, constraints []string) bool {
+func isVersionSupport(standaloneVersion string, constraints []string) bool {
```
breeswish, line 49:
```suggestion
func isVersionSupported(standaloneVersion string, constraints []string) bool {
```

pkg/apiserver/conprof/service.go
shhdgit: How about integrating gin middleware into `FeatureFlag` struct.

pkg/utils/feature_flag.go
```diff
@@ -21,10 +22,31 @@ import (
 	"github.com/pingcap/tidb-dashboard/pkg/utils/version"
 )
 
+type FeatureFlag struct {
+	Name string
+
+	supportedVersions []string
```
breeswish, line 28: should it be called `constraints`?

pkg/apiserver/conprof/service.go
shhdgit: It's a little strange.

pkg/utils/feature_flag.go
```diff
@@ -21,10 +22,31 @@ import (
...
+	supported         *bool
```
breeswish, line 29: I think it is not a good idea to make a util stateful.

pkg/utils/feature_flag.go
```diff
@@ -21,10 +22,31 @@ import (
...
+}
+
+func NewFeatureFlag(name string, supportedVersions []string) *FeatureFlag {
+	return &FeatureFlag{Name: name, supportedVersions: supportedVersions}
+}
+
+func (ff *FeatureFlag) IsSupport(targetVersion string) bool {
```
breeswish, line 36: This implementation is broken when targetVersion is changed.

pkg/utils/feature_flag.go
```diff
@@ -8,6 +8,7 @@
 //
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```
breeswish, line 11: 🤔

pkg/apiserver/user/module.go
```diff
@@ -11,24 +11,10 @@
...
-func IsFeatureSupport(config *config.Config) (supported bool) {
-	if featureSupported != nil {
-		return *featureSupported
-	}
-
-	supported = utils.IsVersionSupport(config.FeatureVersion, supportedTiDBVersions)
-	featureSupported = &supported
-	return
-}
+var FeatureFlagNonRootLogin = utils.NewFeatureFlag("nonRootLogin", []string{">= 5.3.0"})
```
breeswish, line 20: Could you move this util to our new `util` directory? I think we'd better directly implement new utils in the `/util` instead of `/pkg/...`, in order to not being a block stone of transiting to the new `/util`.

pkg/utils/feature_flag.go
shhdgit: Agreed!

pkg/apiserver/user/module.go
shhdgit: Maybe I can move this util after https://github.com/pingcap/tidb-dashboard/pull/1052 is merged.

pkg/apiserver/conprof/service.go
shhdgit: Updated. It's still long, but the purpose is much clearer.

shhdgit:
@breeswish 
Updated, PTAL.

pkg/apiserver/info/info.go
```diff
@@ -28,8 +28,7 @@ import (
 	"github.com/pingcap/tidb-dashboard/pkg/config"
 	"github.com/pingcap/tidb-dashboard/pkg/dbstore"
 	"github.com/pingcap/tidb-dashboard/pkg/tidb"
-	"github.com/pingcap/tidb-dashboard/pkg/utils"
-	"github.com/pingcap/tidb-dashboard/pkg/utils/version"
+	"github.com/pingcap/tidb-dashboard/util/versionutil"
```
breeswish, line 31: Oh. I forgot to list when util should be called "xxx" or "xxxutil". My current convention is that:

If there is a commonly used package called "xxx", then we should use "xxxutil" to avoid naming conflict. For example, if we want to have our own tls util, since there is a [std tls package](https://pkg.go.dev/crypto/tls), we should call it "tlsutil".

Otherwise it would be just fine to call "xxx".

You can name it as "version" I think. There is no name conflict.

util/versionutil/version.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/pingcap/log"
 	"go.uber.org/zap"
 
-	"github.com/pingcap/tidb-dashboard/pkg/utils/distro"
+	"github.com/pingcap/tidb-dashboard/util/distro"
 )
 
 type Info struct {
```
breeswish, line 26: I suspect this util can only be used in TiDB Dashboard instead of TiDB Dashboard variants, as not all variants needs to specify a PD version. Is it possible to make it more general?

util/versionutil/feature_flag.go
```diff
@@ -37,8 +35,8 @@ func NewFeatureFlag(name string, constraints []string) *FeatureFlag {
 // constraints examples: "~5.2.2", ">= 5.3.0", see semver docs to get more information
 func (ff *FeatureFlag) IsSupported(targetVersion string) bool {
 	curVersion := targetVersion
-	if version.Standalone == "No" {
-		curVersion = version.PDVersion
+	if Standalone == "No" {
```
breeswish, line 38: Standalone is a concept of TiDB Dashboard instead of other variants. Can we make it more general so that we can use it in Debug Portal or other variants?

pkg/apiserver/utils/mw_feature_flag.go
```diff
@@ -18,13 +18,12 @@ import (
...
 	"github.com/gin-gonic/gin"
-
-	"github.com/pingcap/tidb-dashboard/pkg/utils"
+	"github.com/pingcap/tidb-dashboard/util/versionutil"
 )
 
 var ErrFeatureNotSupported = ErrNS.NewType("feature_not_supported")
 
-func MWForbidByFeatureFlag(featureFlags []*utils.FeatureFlag, targetVersion string) gin.HandlerFunc {
+func MWForbidByFeatureFlag(featureFlags []*versionutil.FeatureFlag, targetVersion string) gin.HandlerFunc {
```
breeswish, line 27: This middleware can be provided by the version util.

pkg/apiserver/conprof/service.go
```diff
@@ -82,7 +82,7 @@ func newService(lc fx.Lifecycle, p ServiceParams) *Service {
 func registerRouter(r *gin.RouterGroup, auth *user.AuthService, s *Service) {
 	endpoint := r.Group("/continuous_profiling")
 
-	endpoint.Use(apiutils.MWForbidByFeatureFlag([]*utils.FeatureFlag{
+	endpoint.Use(apiutils.MWForbidByFeatureFlag([]*versionutil.FeatureFlag{
```
breeswish, line 85: Is it possible to accept a calling like this:

```golang
endpoint.Use(version.FeatureFlagGateFn(featureConstraint))
```

pkg/config/config.go
```diff
@@ -40,7 +42,7 @@ func Default() *Config {
 		TiDBTLSConfig:      nil,
 		EnableTelemetry:    true,
 		EnableExperimental: false,
-		FeatureVersion:     "",
+		FeatureVersion:     version.PDVersion,
```
baurine, line 45: It's a nice improvement.

baurine (commented):
LGTM!

baurine (approved):

util/feature/flag.go
```diff
@@ -0,0 +1,57 @@
...
+		if c.Check(v) {
+			return true
+		}
+	}
+	return false
+}
+
+// Register feature.Flag to feature.Manager, this can be easily used with fx.
+// e.g. `fx.Invoke(featureFlag.Register())`.
+func (f *Flag) Register() func(m *Manager) {
```
breeswish, line 51: As a util you'd better not rely on fx.

Also this enforces a global feature manager?

breeswish (commented):
I suggest you to provide a doc, describing your designs, then we discuss designs first. As you can see, we have iterate over the architectural design for long time already. It is not efficient.

shhdgit:
Yes, it's time to have a design doc.

util/feature/flag.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/pingcap/tidb-dashboard/util/nocopy"
+)
+
+type Flag struct {
+	nocopy.NoCopy
+
+	Name string
+
+	once        sync.Once
```
breeswish, line 19: Is it possible to avoid having this `once`? You can maintain who is registered in the manager.

util/feature/flag.go
```diff
@@ -0,0 +1,57 @@
...
+	"strings"
+	"sync"
+
+	"github.com/Masterminds/semver"
+
+	"github.com/pingcap/tidb-dashboard/util/nocopy"
+)
+
+type Flag struct {
+	nocopy.NoCopy
```
breeswish, line 15: You can provide `Name` via `GetName` or `Name()`. Then `nocopy` can be avoided as `Flag` is totally immutable.

util/feature/manager.go
```diff
@@ -0,0 +1,75 @@
...
+		if _, ok := m.supportedMap[ff.Name]; !ok {
+			supported = false
+			unsupportedFeatures = append(unsupportedFeatures, ff.Name)
+			continue
+		}
+	}
+
+	return func(c *gin.Context) {
+		if !supported {
+			_ = c.Error(fmt.Errorf("unsupported features: %v", unsupportedFeatures))
```
breeswish, line 67: How about assigning an errorx type for it?

util/feature/manager_test.go
```diff
@@ -0,0 +1,79 @@
...
+	})
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest("GET", "/ping", nil)
+	r.ServeHTTP(w, req)
+
+	require.Equal(t, 200, w.Code)
+	require.Equal(t, "pong", w.Body.String())
+
+	// StatusForbidden
+	r2 := gin.Default()
```
breeswish, line 69: Could you also check whether it works with the error handler fn? (we should expect well formatted error messages at any time)

util/feature/manager.go
```diff
@@ -0,0 +1,75 @@
...
+			c.Status(http.StatusForbidden)
+			c.Abort()
```
breeswish, line 69: Will this abort causing error handler fn not handling it?

util/feature/manager.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/gin-gonic/gin"
+)
+
+type Manager struct {
+	version      string
+	flags        []*Flag
+	supportedMap map[string]struct{}
+}
+
+func NewManager(version string) *Manager {
```
breeswish, line 19: How about just calling this "VersionGuard"? Then it is normal to have a "GinFn" which does the version guard thing. We should avoid using "manager" as much as possible, if there is a more precise word to describe its behavior.

util/feature/manager.go
```diff
@@ -0,0 +1,75 @@
...
+		supportedMap: map[string]struct{}{},
+	}
+}
+
+// NewManagerProvider returns a fx.Provider.
+func NewManagerProvider(version string) func() *Manager {
+	return func() *Manager {
+		return NewManager(version)
+	}
+}
```
breeswish, line 32: I guess this doesn't reduce LOC as you expect, while increasing the difficulty to understand this code. For example, without this function, in order to use the FeatureManager in fx, it is not that hard:

```go
fx.Provide(
   ...
   feature.NewManager(version)),
   ...
)
```

While in the current code base, you forces the reader to understand what is a "Provider". There are also many different "Provider" concepts in this repo, which will confuse the readers deeper.

util/feature/manager.go
```diff
@@ -0,0 +1,75 @@
...
+	for k := range m.supportedMap {
+		sf = append(sf, k)
+	}
+	sort.Strings(sf)
+	return sf
+}
+
+// Guard returns gin.HandlerFunc as guard middleware.
+// It will determine if features are available in the current version.
+func (m *Manager) Guard(featureFlags []*Flag) gin.HandlerFunc {
```
breeswish, line 54: Flags are passed in any way here. Then why do they need to be registered first? Seems that they don't even need any registrations.

breeswish: You can accept `featureFlags ...*Flag` or `featureFlags ...Flag` to make its usage simpler.

util/feature/manager.go
shhdgit: It looks like we should provide a constructor function for `fx.Provide`.
https://pkg.go.dev/go.uber.org/fx#Provide

util/feature/manager.go
shhdgit: Yes, shouldn't abort here.

shhdgit: Nope, because we register `MWHandleErrors` before feature flag router guard. According to `Next` impl, we can still exec `MWHandleErrors` when middlewares' calls backtracking. `Abort` only abort prevents pending handlers.

pkg/apiserver/apiserver.go
```diff
@@ -100,6 +101,7 @@ func (s *Service) Start(ctx context.Context) error {
 	s.app = fx.New(
 		fx.Logger(utils.NewFxPrinter()),
 		fx.Provide(
+			featureflag.ProvideRegistry(s.config.FeatureVersion),
```
breeswish, line 104: How about:

```go
fx.Supply(featureflag.NewRegistry(s.config.FeatureVersion))
```

util/featureflag/version_guard.go
```diff
@@ -0,0 +1,36 @@
...
+			unsupportedFeatures = append(unsupportedFeatures, ff.Name())
+			continue
+		}
+	}
+
+	return func(c *gin.Context) {
+		if !supported {
+			_ = c.Error(ErrFeatureUnsupported.New("list: %v", unsupportedFeatures))
+			c.Status(http.StatusForbidden)
+			c.Abort()
```
breeswish, line 30: You can attach the forbidden to the ErrFeatureUnsupported type

pkg/apiserver/user/auth.go
```diff
@@ -88,8 +91,9 @@ func NewAuthService() *AuthService {
 	}
 
 	service := &AuthService{
-		middleware:     nil,
-		authenticators: map[utils.AuthType]Authenticator{},
+		FeatureFlagNonRootLogin: featureFlagRegistry.Register("nonRootLogin", ">= 5.3.0"),
```
breeswish, line 94: Seems that this FeatureFlagNonRootLogin is not used. It doesn't look good as this may imply that this protection is purely happening at frontend and there is no backend protection for corresponding features.

pkg/apiserver/user/auth.go
```diff
@@ -30,6 +31,8 @@ var (
 )
 
 type AuthService struct {
+	FeatureFlagNonRootLogin *featureflag.FeatureFlag
```
breeswish, line 34: This feature should be placed inside SQL auth, as only SQL auth needs it and other kind of auth does not care about the "non root login".

breeswish (commented):
the rest lgtm

util/featureflag/version_guard.go
shhdgit: That's cool!

pkg/apiserver/apiserver.go
shhdgit: Upgrade fx to 1.16.0 to support `fx.Supply`.

go.sum
```diff
@@ -450,11 +455,11 @@ golang.org/x/tools v0.0.0-20191029190741-b9c20aec41a5/go.mod h1:b+2E5dAYhXwXZwtn
 golang.org/x/tools v0.0.0-20191030062658-86caa796c7ab/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
 golang.org/x/tools v0.0.0-20191107010934-f79515f33823/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
 golang.org/x/tools v0.0.0-20191108193012-7d206e10da11/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
-golang.org/x/tools v0.0.0-20191114200427-caa0b0f7d508/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
 golang.org/x/tools v0.0.0-20191119224855-298f0cb1881e/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
 golang.org/x/tools v0.0.0-20200130002326-2f3ba24bd6e7/go.mod h1:TB2adYChydJhpapKDTa4BR/hXlZSLoq2Wpct/0txZ28=
-golang.org/x/tools v0.0.0-20210112230658-8b4aab62c064 h1:BmCFkEH4nJrYcAc2L08yX5RhYGD4j58PTMkEUDkpz2I=
 golang.org/x/tools v0.0.0-20210112230658-8b4aab62c064/go.mod h1:emZCQorbCU4vsT4fOWvOPXz4eW1wZW4PmDk9uLelYpA=
+golang.org/x/tools v0.1.5 h1:ouewzE6p+/VEB31YYnTbEJdi8pFqKp4P4n85vwo3DHA=
```
breeswish, line 461: According to my experience if you (indirectly) introduced new x/tools then compiling will fail on Go1.13.

pkg/apiserver/apiserver.go
```diff
@@ -100,6 +101,7 @@ func (s *Service) Start(ctx context.Context) error {
 	s.app = fx.New(
 		fx.Logger(utils.NewFxPrinter()),
 		fx.Provide(
+			fx.Supply(featureflag.NewRegistry(s.config.FeatureVersion)),
```
breeswish, line 104: I think your `fx.Supply` should be placed outside `fx.Provide`?

shhdgit: Yes, thank you! I'll finish the remain issues today.

go.sum
shhdgit: Oops.. I downgrade the `fx` to v1.12.0, that looks much better.

pkg/apiserver/user/auth.go
shhdgit: Create a new PR #1089 to refactor the "non root login" for better review experience.

go.sum
shhdgit: No break changes from 1.10.0 to 1.12.0. https://github.com/uber-go/fx/blob/master/CHANGELOG.md

util/featureflag/featureflag.go
```diff
@@ -0,0 +1,71 @@
...
+	return f.isSupported
+}
+
+// VersionGuard returns gin.HandlerFunc as guard middleware.
+// It will determine if features are available in the target version.
+func (f *FeatureFlag) VersionGuard() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		if !f.isSupported {
+			_ = c.Error(ErrFeatureUnsupported.New(f.name))
+			c.AbortWithStatus(http.StatusForbidden)
```
breeswish, line 43: This may be not compatible with the rest.ErrorHandlerFn. Please add a test to verify it.

util/featureflag/featureflag.go
```diff
@@ -0,0 +1,71 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/Masterminds/semver"
+	"github.com/gin-gonic/gin"
+
+	"github.com/pingcap/tidb-dashboard/util/rest"
+)
+
+var ErrFeatureUnsupported = rest.ErrForbidden.NewSubtype("feature_unsupported")
```
breeswish, line 15: Oh. I'm sorry, I misled you. There is no way to attach the error status code into the error type for now.

Please use `HTTPCodeProperty` when initializing errors:

```go
var ErrFeatureUnsupported = ErrNS.xxxx()

_ = c.Error(ErrFeatureUnsupported.NewWithNoMessage().WithProperty(HTTPCodeProperty(http.StatusForbidden)))
```

or:

```go
_ = c.Error(rest.ErrForbidden.New("<ErrorMessages>")
```

breeswish (approved):

util/featureflag/featureflag_test.go
```diff
@@ -54,39 +55,48 @@ func Test_VersionGuard(t *testing.T) {
...
-	r.Equal(200, w.Code)
+	r.Equal(http.StatusOK, w.Code)
 	r.Equal("pong", w.Body.String())
 
-	// abort with other middlewares
+	// abort
 	handled := false
 	e2 := gin.Default()
 	e2.Use(func(c *gin.Context) {
 		c.Next()
```
breeswish: You can use CreateTestContext to test context directly. No need to use a middleware to test it. Example:

```go
func TestDownload(t *testing.T) {
	handler := New()
	token := handler.mustGetDownloadToken(t, "foobar", "file.txt", time.Second*5)

	r := httptest.NewRecorder()
	c, _ := gin.CreateTestContext(r)
	c.Request, _ = http.NewRequest(http.MethodGet, "/download?token="+token, nil)
	handler.HandleDownloadRequest(c)

	require.Len(t, c.Errors, 0)
	require.Equal(t, http.StatusOK, r.Code)
	require.Equal(t, `attachment; filename="file.txt"`, r.Header().Get("Content-Disposition"))
	require.Equal(t, `application/octet-stream`, r.Header().Get("Content-Type"))
	require.Equal(t, "foobar", r.Body.String())

	// Download again
	r = httptest.NewRecorder()
	c, _ = gin.CreateTestContext(r)
	c.Request, _ = http.NewRequest(http.MethodGet, "/download?token="+token, nil)
	handler.HandleDownloadRequest(c)

	require.Len(t, c.Errors, 1)
	require.True(t, errorx.IsOfType(c.Errors[0].Err, rest.ErrBadRequest))
	require.Contains(t, c.Errors[0].Error(), "Download file not found")
}
```

util/featureflag/featureflag_test.go
```diff
@@ -54,39 +55,48 @@ func Test_VersionGuard(t *testing.T) {
...
+	e.Use(rest.ErrorHandlerFn())
+	e.Use(f.VersionGuard())
+	e.GET("/ping", func(c *gin.Context) {
+		c.String(http.StatusOK, "pong")
+	})
+	w2 := httptest.NewRecorder()
+	req2, _ := http.NewRequest("GET", "/ping", nil)
+	e.ServeHTTP(w2, req2)
+
+	r.Equal(http.StatusForbidden, w2.Code)
```
breeswish, line 101: Forbidden means Unauthorized: The HTTP 403 Forbidden response status code indicates that the server understands the request but refuses to authorize it.

However there is no way for this router to be authorized. So maybe BadRequest is better.

PR #1064, opened 2021-11-19
Integrate speedscope

YiniXu9506:
This PR achieved:

1. integrating Speedscope to TiDB dashboard with the following improvements:
   -  ignore samples with negative weights
   -  make stack name more concise
   -  set `left_heavy` as default view mode
2. unifying fetch method for all instances by adding `Content-Type:application/protobuf` to TiKV/TiFlash request header. 
3. supporting to view CPU profiling of each instance as flamegraph online by importing **protobuf** file.
4. hiding Graph temporary and will support in the near future.

Screen Shot 2021-11-19 at 11 57 01 AM

Many thanks to [speedscope](https://github.com/jlfwong/speedscope) for the amazing work, here we use the [Left Heavy](https://github.com/jlfwong/speedscope#%EF%B8%8Fleft-heavy) view mode to understand where all the time is going in situations where there are hundreds or thousands of function calls interleaved between other call stacks. Other view modes and more details can be referred [here](https://github.com/jlfwong/speedscope#speedscope).


pkg/apiserver/apiserver.go
```diff
@@ -212,6 +213,7 @@ func newAPIHandlerEngine() (apiHandlerEngine *gin.Engine, endpoint *gin.RouterGr
 	apiHandlerEngine.Use(apiutils.MWHandleErrors())
 
 	endpoint = apiHandlerEngine.Group("/dashboard/api")
+	endpoint.StaticFS("/speedscope", speedscopeFiles.Assets())
```
breeswish, line 216: I think we should not expose speedscope static files in the `/api` category. How about letting it to be `/dashboard/speedscope`?

breeswish (commented):
mostly lgtm

pkg/apiserver/profiling/model.go
```diff
@@ -80,7 +80,7 @@ func NewTask(ctx context.Context, taskGroup *TaskGroup, target model.RequestTarg
 
 func (t *Task) run() {
 	fileNameWithoutExt := fmt.Sprintf("profiling_%d_%d_%s", t.TaskGroupID, t.ID, t.Target.FileName())
-	svgFilePath, err := profileAndWriteSVG(t.ctx, t.fetchers, &t.Target, fileNameWithoutExt, t.taskGroup.ProfileDurationSecs)
+	svgFilePath, err := profileAndWritePprof(t.ctx, t.fetchers, &t.Target, fileNameWithoutExt, t.taskGroup.ProfileDurationSecs)
```
breeswish, line 83: This is no longer `svgFile`. Please update the variable name accordingly.

pkg/apiserver/profiling/fetcher.go
```diff
@@ -74,6 +77,9 @@ type tiflashFetcher struct {
 }
 
 func (f *tiflashFetcher) fetch(op *fetchOptions) ([]byte, error) {
+	f.client.WithBeforeRequest(func(req *http.Request) {
```
breeswish, line 80: This causes protobuf to be stored in the database. However there could be existing svg in the database. I think we should handle this case. Otherwise user will no longer be able to view history profilings.

pkg/tiflash/client.go
```diff
@@ -54,6 +54,11 @@ func (c Client) WithTimeout(timeout time.Duration) *Client {
 	return &c
 }
 
+func (c Client) WithBeforeRequest(callback func(req *http.Request)) *Client {
+	c.httpClient.BeforeRequest = callback
+	return &c
+}
+
```
shhdgit, line 61: There's an incorrect impl at pd/client.go. It will set `BeforeRequest` to the base http client reference.

We can fix like [this](https://github.com/pingcap/tidb-dashboard/pull/1050/files#diff-45e2e42c28256d3b1f144bc786d0e72deade31665db3a188aa9b32fe516e583eR86-R89). And invoke `c.httpClient.WithBeforeRequest` at L64.

pkg/apiserver/apiserver.go
```diff
@@ -213,7 +213,8 @@ func newAPIHandlerEngine() (apiHandlerEngine *gin.Engine, endpoint *gin.RouterGr
 	apiHandlerEngine.Use(apiutils.MWHandleErrors())
 
 	endpoint = apiHandlerEngine.Group("/dashboard/api")
-	endpoint.StaticFS("/speedscope", speedscopeFiles.Assets())
+	ssEndpoint := apiHandlerEngine.Group("/dashboard/speedscope")
+	ssEndpoint.StaticFS("/", speedscopeFiles.Assets())
```
breeswish, line 217: I think you should either move these two lines out of `apiserver`, or rename this function, as it is not an API, but clearly static assets.

breeswish (commented):
Seems that using `StaticFS` just causes you more trouble than simply putting all assets in `public/` and reuse current logics, without breaking current code guarantees :) 

This is the downside of introducing a new mechanism.

YiniXu9506:
> Seems that using `StaticFS` just causes you more trouble than simply putting all assets in `public/` and reuse current logics, without breaking current code guarantees :)
> 
> This is the downside of introducing a new mechanism.

I have been updated the integration method, which is copy all assets of speedscope from node_modules to `public/speedscope/' without changing current endpoint mechanism.

pkg/tiflash/client.go
shhdgit: Sorry, this fix still has the problem of shared pointers.

You might be blocked by this pr. https://github.com/pingcap/tidb-dashboard/pull/1069

YiniXu9506: Got it!

pkg/apiserver/profiling/model.go
```diff
@@ -23,13 +23,14 @@ const (
...
-	Error       string                  `json:"error" gorm:"type:text"`
-	StartedAt   int64                   `json:"started_at"` // The start running time, reset when retry. Used to estimate approximate profiling progress.
+	ID                uint                    `json:"id" gorm:"primary_key"`
+	TaskGroupID       uint                    `json:"task_group_id" gorm:"index"`
+	State             TaskState               `json:"state" gorm:"index"`
+	Target            model.RequestTargetNode `json:"target" gorm:"embedded;embedded_prefix:target_"`
+	FilePath          string                  `json:"file_path" gorm:"type:text"`
+	Error             string                  `json:"error" gorm:"type:text"`
+	StartedAt         int64                   `json:"started_at"` // The start running time, reset when retry. Used to estimate approximate profiling progress.
+	ProfileOutputType string                  `json:"profile_output_type"`
```
YiniXu9506, line 33: Set the ProfileOutputType as string since we will have other output types is the future. It's more flexible to use string than use bool.

pkg/apiserver/profiling/pprof.go
```diff
@@ -3,170 +3,67 @@
...
-func (b blankPprofUI) Print(i ...interface{}) {
-}
-
-func (b blankPprofUI) PrintErr(i ...interface{}) {
-}
-
-func (b blankPprofUI) IsTerminal() bool {
-	return false
-}
+	w, err := os.Create(tmpPath)
```
breeswish, line 53: You can create a temp file in this function, this is a simplified version if you just want a temp file for writing:

```go
f, err := os.CreateTemp("", op.fileNameWithoutExt)
if err != nil {
  ..
}
_, err = f.Write(resp)
..

_ = f.Close()
```

pkg/apiserver/profiling/router.go
```diff
@@ -100,10 +101,19 @@ func (s *Service) getGroupList(c *gin.Context) {
 	c.JSON(http.StatusOK, resp)
 }
 
+type ResponseTask struct {
+	ID                uint                    `json:"id" gorm:"primary_key"`
```
breeswish, line 105: As a API response, there is no need to have `gorm` tags.

pkg/apiserver/profiling/router.go
```diff
@@ -100,10 +101,19 @@ func (s *Service) getGroupList(c *gin.Context) {
...
+	State             TaskState               `json:"state" gorm:"index"`
+	Target            model.RequestTargetNode `json:"target" gorm:"embedded;embedded_prefix:target_"`
+	Error             string                  `json:"error" gorm:"type:text"`
+	StartedAt         int64                   `json:"started_at"` // The start running time, reset when retry. Used to estimate approximate profiling progress.
+	ProfileOutputType string                  `json:"profile_output_type"`
```
breeswish, line 110: As you already have this field in the db model, then this struct can be omitted (that you can use the db model directly) as there is no new information need to be added. We can have this when we need something not in the db but want to be passed to the frontend.

pkg/httpc/client.go
```diff
@@ -23,6 +23,8 @@ const (
 	defaultTimeout = time.Second * 10
 )
 
+type BeforeRequestFunc func(req *http.Request)
```
breeswish, line 26: I guess this is a merge conflict.

breeswish (commented):
The rest LGTM. There are no code logic problems.

shhdgit (approved):
LGTM.

pkg/apiserver/profiling/pprof.go
```diff
@@ -3,170 +3,60 @@
...
-
-func (f *flagSet) AddExtraUsage(eu string) {}
-
-// oswriter implements the Writer interface using a regular file.
-type oswriter struct {
-	output string
-}
+	defer func() {
+		if err := tmpfile.Close(); err != nil {
+			fmt.Printf("failed to close file, %v", err)
```
breeswish, line 44: You can use `log.Warn`, since `fmt.Printf` will break the output log format.

Or you can just discard this error :) In the rest of the code base this kind of error is usually discarded.

breeswish (approved):
The rest LGTM

breeswish (approved):

PR #1069, opened 2021-11-24
Fix request header being pinned after pd profiling

shhdgit:


shhdgit:
/cc @baurine @breeswish @YiniXu9506 
PTAL, thanks!

breeswish (approved):

PR #1073, opened 2021-11-29
Refactor: a new httpclient

breeswish:
This PR provides a new design of the HTTP client, pretty much close to [resty](https://github.com/go-resty/resty).

Primary advantages:
- No need to call `Close()`, just like resty.
- No need to process error multiple times during the calling, just like resty, one chained call for all.
- Unlike resty, No need to call and pair `SetDoNotParseResponse()` + `.RawResponse` in order to do streaming, just call `PipeBody()`.
- Unlike resty, No need to call and pair `SetResult(&Foo{})` + `.Result().(*Foo)` in order to unmarshal, just call `ReadBodyAsJSON(&foo)`.

Sample usages below :

**Get response body in bytes**:

```go
bytes, _, err := httpclient.New(httpclient.Config{}).LR().
    Get(....).
    ReadBodyAsBytes(&data)
if err != nil {
    // error handling
}
```

**Unmarshal JSON**:

```go
var data FooResponse
_, err := httpclient.New(httpclient.Config{}).LR().
    Get(....).
    ReadBodyAsJSON(&data)
if err != nil {
    // error handling
}
```

**Streaming**:

```go
writtenBytes, _, err := httpclient.New(httpclient.Config{}).LR().
    Get(....).
    PipeBody(someWriter)
if err != nil {
    // error handling
}
```

**Set Headers**:

```go
var data FooResponse
_, err := httpclient.New(httpclient.Config{}).LR().
    SetHeader("x-requested-with", "foo").
    Get(....).
    ReadBodyAsJSON(&data)
if err != nil {
    // error handling
}
```

**Reuse clients**:

```go
client := httpclient.New(httpclient.Config{})
_, _, _ = client.LR().Get(...).ReadBodyAsString()
_, _, _ = client.LR().Get(...).ReadBodyAsString()
```

**Read response headers**:

```go
bytes, rawResp, err := httpclient.New(httpclient.Config{}).LR().
    Get(....).
    ReadBodyAsBytes(&data)
if err != nil {
    // error handling
}
fmt.Println(rawResp.Header.Get("x-header"))
```


util/israce/race.go
```diff
@@ -0,0 +1,9 @@
+// Copyright 2021 PingCAP, Inc. Licensed under Apache-2.0.
+
+// +build race
+
+// Package israce reports if the Go race detector is enabled.
```
shhdgit, line 5: knowledge++++

util/client/httpclient/client.go
```diff
@@ -8,92 +8,68 @@ package httpclient
...
+	}
+	return &http.Transport{
+		Proxy:                 http.ProxyFromEnvironment,
+		DialContext:           dialer.DialContext,
+		ForceAttemptHTTP2:     true,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+		MaxIdleConnsPerHost:   runtime.GOMAXPROCS(0) + 1,
```
shhdgit, line 52: Does this mean the min value of `MaxIdleConnsPerHost` is 2?

util/client/httpclient/response.go
```diff
@@ -0,0 +1,222 @@
...
+	lResp.close()
+	return
+}
+
+func (lResp *LazyResponse) ReadBodyAsString() (data string, respNoBody *http.Response, err error) {
+	bytes, resp, err := lResp.ReadBodyAsBytes()
+	if err != nil {
+		return "", nil, err
+	}
+	return strings.TrimSpace(string(bytes)), resp, nil
```
shhdgit, line 194: Maybe it's too surprising to have `TrimSpace`, as a common http client.

breeswish: This behavior is taken from resty:

```go
// String method returns the body of the server response as String.
func (r *Response) String() string {
	if r.body == nil {
		return ""
	}
	return strings.TrimSpace(string(r.body))
}
```

shhdgit (approved):

PR #1079, opened 2021-12-01
Refactor: Move tools into a standalone module

breeswish:
This reduces the dependencies of the module github.com/pingcap/tidb-dashboard .

- [x] Merge https://github.com/pingcap/tidb-dashboard/pull/1078 first to get a nice diff.


shhdgit (approved):

scripts/go.mod
```diff
@@ -0,0 +1,14 @@
+module scripts
+
+go 1.13
```
shhdgit, line 3: What will happen if we upgrade the go version in the scripts module.

breeswish: I think there is no difference, as our CI is using Go1.13.

PR #1081, opened 2021-12-03
chore: fix profileURL of speedscope

YiniXu9506:
The `client.getBasePath` will only return API basePath in prod mode, however, speedscope profileURL requires `${apiOrigin}${apiBasePath}` both in dev and prod mode. This PR fixed api baseURL of speedscope profileURL in prod mode.

### DIFF
--- a/ui/lib/apps/InstanceProfiling/pages/Detail.tsx
+++ b/ui/lib/apps/InstanceProfiling/pages/Detail.tsx
@@ -127,7 +127,7 @@ export default function Page() {
       }
 
       // make both generated graph(svg) file and protobuf file viewable online
-      let profileURL = `${client.getBasePath()}/profiling/single/view?token=${token}`
+      let profileURL = `${client.getApiBaseURL()}/profiling/single/view?token=${token}`
 
       if (rec.profile_output_type === 'protobuf') {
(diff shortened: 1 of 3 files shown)


ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -127,7 +127,7 @@ export default function Page() {
       }
 
       // make both generated graph(svg) file and protobuf file viewable online
-      let profileURL = `${client.getBasePath()}/profiling/single/view?token=${token}`
+      let profileURL = `${client.getApiBaseURL()}/profiling/single/view?token=${token}`
```
breeswish, line 130: I recommend to call it `getAbsoluteBasePath`, to ensure the "absolute" behavior.

YiniXu9506: Make sense!  We may have the same requirement in the future, so we can support the `getAbsoluteBasePath` for future use in the next time. How about importing profiling file via relative URL instead of absolute path, see relative [PR](https://github.com/pingcap/tidb-dashboard/pull/1082).

breeswish:
Close for https://github.com/pingcap/tidb-dashboard/pull/1082

PR #1083, opened 2021-12-05
Add integration test

shhdgit:
Including:
- Integration test
- Compatibility test
- CI


Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: install_tools lint dev yarn_dependencies ui server run
+.PHONY: install_tools lint dev yarn_dependencies ui server run test
```
breeswish, line 1: unit_test and integration_test should be put in .PHONY as well

Makefile
```diff
@@ -24,6 +27,14 @@ install_tools:
 lint:
 	scripts/lint.sh
 
+test: unit_test integration_test
+
+unit_test:
```
breeswish, line 32: Unit test is failing in CI, could you take a look?

Makefile
```diff
@@ -24,6 +27,14 @@ install_tools:
...
+	$(GOTEST) $$(go list ./... | grep -v $(INTEGRATION_TEST_PACKAGES))
```
breeswish, line 33: There may be another way to do this: you can use build tags to mark different kind of test packages!

tests/run.sh
```diff
@@ -0,0 +1,21 @@
...
+
+echo "+ Start integration tests"
+SELECTED_TEST_NAME="${TEST_NAME-$(find tests -mindepth 2 -maxdepth 2 -name run.sh | cut -d/ -f2 | sort)}"
+echo -e "  - Selected test cases: \n$SELECTED_TEST_NAME"
+
+for casename in $SELECTED_TEST_NAME; do
+  script=tests/$casename/run.sh
+  echo "+ Running test $script..."
+  bash "$script" && echo "  - TEST: [$casename] success!"
+done
```
breeswish, line 21: How about simply calling `go test` to run them all?

tests/download_tools.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
breeswish, line 1: Seems that you made a `_utils` directory. Then how about moving it to this directory?

.github/workflows/test.yaml
```diff
@@ -7,8 +7,28 @@ on:
...
+      - name: Load go module cache
+        uses: actions/cache@v1
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+      - name: Test
+        run: |
+          make test
```
breeswish, line 29: 👍 I like the current way. In this way tests can be easily run in local as well.

breeswish (commented):
Cool!

.github/workflows/test.yaml
```diff
@@ -7,8 +7,28 @@ on:
       - release
 
 jobs:
+  backend:
+    name: backend
```
breeswish, line 11: I think you can make backend unit test and integration test run in parallel.

Makefile
shhdgit: Yeah, sure.

Makefile
shhdgit: Build tags seems to have some cons in static analysis: errors, code completion, etc.

How about separate different kind of tests by folder or env variables?
https://peter.bourgon.org/blog/2021/04/02/dont-use-build-tags-for-integration-tests.html

shhdgit:
PTAL, thanks!
/cc @baurine @breeswish

tests/integration/slowquery_test.go
```diff
@@ -0,0 +1,181 @@
...
+		Order("Time").
+		Find(&queryDs)
+
+	s.Equal(ds, queryDs)
+}
+
+func (s *testSlowQuerySuite) TestGetListLimitRequest() {
+	ds := s.mustQuerySlowLogList(&slowquery.GetListRequest{Limit: 5})
+
+	s.Len(ds, 5)
```
breeswish, line 108: The default order should be checked

tests/integration/slowquery_test.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func (s *testSlowQuerySuite) TestGetListSearchRequest() {
+	digest := "2375da6810d9c5a0d1c84875b1376bfd469ad952c1884f5dc1d6f36fc953b5df"
+	ds := s.mustQuerySlowLogList(&slowquery.GetListRequest{Fields: "*", Text: digest})
```
breeswish, line 113: Is `Fields: "*"` a must-have?

tests/integration/slowquery_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+	s.NotEmpty(ds)
+	for _, d := range ds {
+		s.Contains(d.Digest, digest)
+	}
```
breeswish, line 118: You can also check the number of rows, as you have a clearly defined data set. The query result should not have anything surprise.

tests/integration/slowquery_test.go
```diff
@@ -0,0 +1,181 @@
...
+)
+
+const (
+	TestSlowQueryTableName = "test.CLUSTER_SLOW_QUERY"
+)
+
+type testSlowQuerySuite struct {
+	suite.Suite
+	db               *testutil.TestDB
+	slowqueryColumns []string
```
breeswish, line 25: This part doesn't look good. It relies heavily on internal implementations. It should only occur in unit tests. The methodology of the integration test should be not relying on other internal implementation details at all. Otherwise it may be hard for others to write something similar for another module.

breeswish (commented):
Generally looks great! Some trivial ideas.

tests/integration/slowquery_test.go
```diff
@@ -0,0 +1,181 @@
+// Copyright 2021 PingCAP, Inc. Licensed under Apache-2.0.
```
breeswish, line 1: maybe you can differentiate mock-db test and real-db via the file name, e.g.:

```
tests/integrations/slow_query/mock_db_test.go
    MockDBSuite
tests/integrations/slow_query/with_db_test.go
    WithDBSuite
```

tests/integration/v4_slowquery_test.go
```diff
@@ -0,0 +1,76 @@
+// Copyright 2021 PingCAP, Inc. Licensed under Apache-2.0.
```
breeswish, line 1: It would be great if your test works for all 4.0+ versions. Otherwise it is too horrible to write one file for one db version.

tests/integration/v4_slowquery_test.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (s *testV4SlowQuerySuite) prepareSlowQuery() {
+	s.db.MustExec("SET tidb_slow_log_threshold = 0")
+	var wg sync.WaitGroup
+	for i := 1; i < 5; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			s.db.MustExec(fmt.Sprintf("SELECT count(*) FROM %s", slowquery.SlowQueryTable))
```
breeswish, line 56: If you are trying to simulate slow queries, then something like `sleep(1)` could be your friend.

tests/util/download_tools.sh
```diff
@@ -0,0 +1,36 @@
...
+
+  if [ ! -e "$BIN/tidb-lightning" ]; then
+    tar -x -f "$BIN/toolkit.tar.gz" -C "$BIN/" tidb-toolkit-nightly-linux-amd64/bin/tidb-lightning
+    mv "$BIN"/tidb-toolkit-nightly-linux-amd64/bin/tidb-lightning "$BIN/tidb-lightning"
+  fi
+
+  if [ ! -e "$BIN/dumpling" ]; then
+    tar -x -f "$BIN/toolkit.tar.gz" -C "$BIN/" tidb-toolkit-nightly-linux-amd64/bin/dumpling
+    mv "$BIN"/tidb-toolkit-nightly-linux-amd64/bin/dumpling "$BIN/dumpling"
+  fi
```
breeswish, line 33: Actually you can import a sql just using the official mysql client. It's a standard mysql usage: https://stackoverflow.com/questions/17666249/how-do-i-import-an-sql-file-using-the-command-line-in-mysql

tests/util/run_services.sh
```diff
@@ -0,0 +1,68 @@
...
+  fi
+
+  if [ -e "$BIN/dumpling" ]; then
+    echo "+ Start dump fixtures..."
+    $BIN/dumpling -u root -P 4000 -h 127.0.0.1 --filetype sql -o $FIXTRUE_DIR -T $1
+    echo "  - Dump success!"
+  else
+    echo "Tool $BIN/dumpling not exist"
+    return 1
+  fi
```
breeswish, line 67: The official mysqldump is enough.

tests/integration/v4_slowquery_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (s *testV4SlowQuerySuite) mustQuerySlowLogList(req *slowquery.GetListRequest) []slowquery.Model {
+	d, err := slowquery.QuerySlowLogList(req, s.slowqueryColumns, s.slowQuerySession())
+	s.NoError(err)
+	return d
+}
+
+func (s *testV4SlowQuerySuite) TestFieldsCompatibility() {
+	ds := s.mustQuerySlowLogList(&slowquery.GetListRequest{Fields: "*"})
+	s.NotEmpty(ds)
```
breeswish, line 75: This check is too loose. It doesn't ensure anything. It does improve the code coverage but does not actually help with preventing mistakes.

breeswish (commented):
The effectiveness of your test case can be imagined by: what bad logic could I introduce into the current code base to break the user but make my test case continue working?

tests/util/run_services.sh
shhdgit: mysqldump 8.0 looks like not supporting dump the INFORMATION_SCHEMA schema.
```
mysqldump does not dump the INFORMATION_SCHEMA schema.
```
https://dev.mysql.com/doc/refman/8.0/en/mysqldump.html

tests/integration/slowquery/compatibility_test.go
```diff
@@ -0,0 +1,113 @@
+// Copyright 2021 PingCAP, Inc. Licensed under Apache-2.0.
```
shhdgit, line 1: since there are mock and real dbs, how about use file name `compatibility_test.go`.

tests/integration/v4_slowquery_test.go
shhdgit: Maybe `sleep(1)` will slow down the tests?

tests/integration/slowquery_test.go
shhdgit: Nope, at least need the field we want to be verified.

tests/integration/slowquery_test.go
shhdgit: The default order has been tested in  the`TestGetListDefaultRequest`. Do we need to test again here?

go.mod
```diff
@@ -33,8 +33,9 @@ require (
 	github.com/pingcap/kvproto v0.0.0-20200411081810-b85805c9476c
 	github.com/pingcap/log v0.0.0-20210906054005-afc726e70354
 	github.com/rs/cors v1.7.0
+	github.com/shhdgit/testfixtures/v3 v3.6.2-0.20211219171712-c4f264d673d3
```
shhdgit, line 36: Need to adapt tidb checksum logic, I'll create a pr for upstream later

tests/util/dump/dump.go
```diff
@@ -0,0 +1,48 @@
+// Copyright 2021 PingCAP, Inc. Licensed under Apache-2.0.
```
shhdgit, line 1: Use testfixtures to dump yaml data helps we easier to modify data.

tests/integration/v4_slowquery_test.go
breeswish: One second sleep should be fine, as long as you don't have too many of them LOL

baurine (approved):
lgtm

YiniXu9506 (approved):
LGTM

PR #1085, opened 2021-12-06
chore: add graph generation

YiniXu9506:
This PR achieved:

1.  supporting view graph of CPU profiling online.
2.  making previous generated SVG files viewable online.
3.  adding multi profiling output types for user to choose by using button with dropdown menu.

The previous generated SVG can only view the Graph or Flame Graph:
image

User can view Graph, Flame Graph and download raw data of a single instance:
image


breeswish:
As the first impression I have two ideas:

1. You may want to use a button with dropdown menu, instead of a dropdown and a menu, to eliminate one click:

   image
   
   doc: https://ant.design/components/button-cn/

2. Seems that the table row is no longer vertically aligned. Could you fix it?

pkg/apiserver/profiling/protobuf_to_svg.go
```diff
@@ -0,0 +1,147 @@
...
+
+func convertProtobufToDot(content []byte, task TaskModel) ([]byte, error) {
+	args := []string{
+		"-dot",
+		// prevent printing stdout
+		"-output", "dummy",
+		"-seconds", strconv.Itoa(int(1)),
+	}
+	address := fmt.Sprintf("%s:%d", task.Target.IP, task.Target.Port)
+	args = append(args, address)
```
breeswish, line 38: I suspect this address might not be needed. Maybe it can be any address?

pkg/apiserver/profiling/router.go
```diff
@@ -303,6 +304,15 @@ func (s *Service) viewSingle(c *gin.Context) {
...
+	if outputType == "graph" {
+		svgContent, err := convertProtobufToSVG(content, task)
+		if err != nil {
+			_ = c.Error(err)
+			return
+		}
+		content = svgContent
+	}
+
 	c.Data(http.StatusOK, "image/svg+xml", content)
```
breeswish, line 316: Seems that previously the content type is set wrongly. When proto is served, its content type should not be image/svg+xml.

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -10,6 +10,16 @@ import { CardTable, DateTime, Head, Descriptions } from '@lib/components'
 import { useClientRequestWithPolling } from '@lib/utils/useClientRequest'
 import { InstanceKindName } from '@lib/utils/instanceTable'
 import useQueryParams from '@lib/utils/useQueryParams'
+import { ScrollablePane } from 'office-ui-fabric-react'
+import { MenuInfo } from 'rc-menu/lib/interface'
+
+const profilingOutputTypeOptions = {
+  flamegraph: 'Flame Graph',
+  graph: 'Graph',
+}
```
breeswish, line 19: Maybe you will find it simpler if it is an enum, which seems to roughly provides what you are trying to do:

```ts
enum ProfilingOutputType {
  FlameGraph,
  Graph,
}

// ...

    if (task.profile_output_type === 'protobuf') {
      task.profilingOutputTypeOptions = [ProfilingOutputType.FlameGraph, ProfilingOutputType.Graph]
    } else if (task.profile_output_type === '') {
      // for legacy records
      switch (task.target.kind) {
        case 'tidb':
        case 'pd':
          task.profilingOutputTypeOptions = [ProfilingOutputType.Graph]
          break
        case 'tiflash':
        case 'tikv':
          task.profilingOutputTypeOptions = [ProfilingOutputType.FlameGraph]
          break
      }
    }

```

This also makes it possible to i18n the profiling type, as profiling types are now constant strings `"FlameGraph"` and `"Graph"`, which can be then used as a translation token.

YiniXu9506: I agree with using enum to declare ProfilingOuputType, but I am not sure is it necessary to translate "FlameGraph" and "Graph" to Chinese.

pkg/apiserver/profiling/protobuf_to_svg.go
YiniXu9506: Yes, you are right. The value of address is not used here, but the arg is still required by driverPporf to convert protobuf to dot format. We can also set it to an empty string.

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
breeswish: how about 关系图? we can change it later if necessary.

pkg/apiserver/profiling/router.go
```diff
@@ -303,7 +311,29 @@ func (s *Service) viewSingle(c *gin.Context) {
...
+				_ = c.Error(err)
+				return
+			}
+			content = svgContent
+			contentType = "image/svg+xml"
+		case string(ViewOutputTypeFlameGraph):
+			contentType = "application/protobuf"
+		default:
+			// Will not handle converting protobuf to other formats except flamegraph and graph
+			log.Warn("Failed to convert protobuf to %v", zap.String("output", outputType))
```
breeswish, line 332: In this case you can use:

```go
_ = c.Error(rest.ErrBadRequest.New("Cannot output protobuf as %s", outputType))
return
```

so that from the API perspective user can receive a standard error response with a correct BadRequest status code.

breeswish (approved):
the rest LGTM. @baurine @shhdgit PTAL

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -28,6 +46,20 @@ function mapData(data) {
...
+    // set profiling output options for previous generated SVG files and protobuf files.
+    if (task.profile_output_type === ProfilingOutputTypeProtobuf) {
+      task.view_default_output_type_val = t(viewDefaultOutputTypeVal)
+    } else {
+      switch (task.target.kind) {
+        case 'tidb':
+        case 'pd':
+          task.view_default_output_type_val = t(ViewOutputTypeOptions.Graph)
+          break
+        default:
```
breeswish, line 59: This is not actually the default. It is not *semantically reasonable* to make "TiKV" and "TiFlash" to be default while TiDB is not a default. You will find `default` can work. But they are not directly related.

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -10,11 +10,29 @@ import { CardTable, DateTime, Head, Descriptions } from '@lib/components'
...
+
+enum taskState {
+  Error,
+  Running,
+  Success,
+}
+
+const ProfilingOutputTypeProtobuf = 'protobuf'
+
+let viewDefaultOutputTypeVal = ViewOutputTypeOptions.FlameGraph
```
breeswish, line 29: I think there is no "default to flame graph" or something. What we only have is `FlameGraph is preferred over Graph for protobuf data`. This definition can be very misleading.

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -10,11 +10,29 @@ import { CardTable, DateTime, Head, Descriptions } from '@lib/components'
...
 import { InstanceKindName } from '@lib/utils/instanceTable'
 import useQueryParams from '@lib/utils/useQueryParams'
+import { ScrollablePane } from 'office-ui-fabric-react'
+import { MenuInfo } from 'rc-menu/lib/interface'
 
-function mapData(data) {
+enum ViewOutputTypeOptions {
+  FlameGraph = 'instance_profiling.detail.table.columns.selection.flamegraph',
+  Graph = 'instance_profiling.detail.table.columns.selection.graph',
+}
```
breeswish, line 19: I think you can use a simple enum, instead of mixing i18n with it. I18N should happen only at the presentation layer, instead of in the models:

```ts
// Define the simplist model value. As a model, it is relied by others. It should do its best without knowing any other outside usages.
enum ViewOutputTypeOptions {
  FlameGraph,
  Graph,
}

// Use the model value in models:
task.view_outputs = [ViewOutputTypeOptions.FlameGraph, ViewOutputTypeOptions.Graph]

// Use the single model value at the presentation layer to support I18N:
    { task.view_outputs.map(vo => (
        <Menu.Item key={vo}>
          {t('instance_profiling.detail.table.columns.selection.view', {
            type: {t(`instance_profiling.detail.table.columns.selection.types.${vo}`)}
          })}
        </Menu.Item>
    )) }
```

pkg/apiserver/profiling/router.go
```diff
@@ -303,7 +311,29 @@ func (s *Service) viewSingle(c *gin.Context) {
 		_ = c.Error(err)
 		return
 	}
-	c.Data(http.StatusOK, "image/svg+xml", content)
+
+	// set default content-type for legacy profiling content.
+	contentType := "image/svg+xml"
+
+	if task.ProfileOutputType == ProfilingOutputTypeProtobuf {
```
breeswish, line 318: Prefer to explicitly add

```go
} else if task.ProfileOutputType == '' {
  contentType := "image/svg+xml"
}
```

instead of using fallbacks, since `task.ProfileOutputType == ''` is the only correct way to test whether it is a legacy record or not.

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -37,6 +69,100 @@ function isFinished(data) {
...
+    window.open(`${profileURL}`, '_blank')
+  })
+
+  const menu = () => {
+    return (
+      <Menu onClick={handleViewResultMenuClick}>
+        <Menu.Item key={ViewOutputTypeOptions.Graph}>
+          {t('instance_profiling.detail.table.columns.selection.view')}{' '}
+          {t(ViewOutputTypeOptions.Graph)}
+        </Menu.Item>
```
breeswish, line 124: This is very complicated and is not approachable when you have different source types. The previous design `view_output_options` (which is an array of options to be displayed) is more reasonable, since it is a design works for now and future.

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
shhdgit: Maybe string enums is better
```
enum ViewOutputTypeOptions {
  FlameGraph = 'flamegraph', 
  Graph = 'graph',
}
```

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -37,6 +69,100 @@ function isFinished(data) {
...
+  const isProtobuf: boolean =
+    rec.profile_raw_data_type === ProfilingRawDataTypeProtobuf
+  let token: string | undefined
+
+  const handleViewResultMenuClick = usePersistFn(async (e: MenuInfo) => {
+    switch (e.key) {
+      case 'download':
+        token = await getActionToken(rec.id, 'single_download')
+
+        window.location.href = `${client.getBasePath()}/profiling/single/download?token=${token}`
```
shhdgit, line 94: What scenario will token be empty? Maybe we shouldn't download or open a new window when token is empty?

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -37,6 +74,119 @@ function isFinished(data) {
...
+        )}
+      </Dropdown.Button>
+    )
+  }
+
+  return (
+    <>
+      {isProtobuf ? (
+        <DropdownButton />
+      ) : (
```
breeswish, line 171: I think your check can be based on `how many view_output_type_options are there`, which is more semantically correct, e.g.

```
if view_output_type_options.length > 1 {
  render as drop down button
} else {
  render as a single button 
}
```

Since the core reason that you want to display a drop down button is that there are multiple view outputs.

This also avoids you checking the underlying raw data type multiple places. You have already checked once in the `mapData()`, which is enough. The more places you introduce these logics, the more complex your code becomes.

breeswish (commented):
The rest LGTM

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -37,6 +74,119 @@ function isFinished(data) {
...
+          disabled={rec.state !== taskState.Success}
+          onClick={handleViewResultBtnClick}
+          style={{ width: 150 }}
+        >
+          {rec.state === taskState.Success
+            ? t(
+                `instance_profiling.detail.table.columns.selection.types.${rec.view_output_type_options[0]}`
+              )
+            : t(
+                `instance_profiling.detail.table.columns.selection.types.no_data`
```
breeswish, line 182: If there is no data I prefer not displaying anything at all.

pkg/apiserver/profiling/router.go
```diff
@@ -303,7 +311,29 @@ func (s *Service) viewSingle(c *gin.Context) {
...
+		switch outputType {
+		case string(ViewOutputTypeGraph):
+			svgContent, err := convertProtobufToSVG(content, task)
+			if err != nil {
+				_ = c.Error(err)
+				return
+			}
+			content = svgContent
+			contentType = "image/svg+xml"
+		case string(ViewOutputTypeFlameGraph):
```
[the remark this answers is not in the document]
YiniXu9506: I totally agree with the relationships you mentioned above, but I don't think we have to change the name `ViewOutputTypeFlameGraph` to `ViewOutputTypeProtobuf`.

Since the OriginalSVG has been checked with the field `task.RawDataType`, it is empty for legacy record, so we can just return the content that been read from the legacy record file.

And for the new records, we should use the query `output_type` to find what kind of content the users want, if they want graph, we convert protobuf to svg, if they want flamegraph, we return the protobuf format. So I think the `ViewOutputType` should indicate users requirements not profiling raw data type.

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -37,6 +74,119 @@ function isFinished(data) {
...
+        }
+
+        let profileURL = `${client.getBasePath()}/profiling/single/view?token=${token}&output_type=${
+          e.key
+        }`
+        window.open(`${profileURL}`, '_blank')
+    }
+  })
+
+  const handleViewResultBtnClick = usePersistFn(async () => {
```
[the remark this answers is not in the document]
YiniXu9506: I think we still need to check the protobuf in `openResult(...)`, in function `mapdata`, we just set the options for menuList, but there are two situations for `FlameGraph` option click and `Graph` option click: 

1 . For the legacy records, both `FlameGraph` and `Graph` click will get the originalSVG, we don't add `output_type` query to the API, then the backend will return "image/svg+xml" content to frontend.
2. For new records, we will append `&output_type=FlameGraph` and `&output_type=Graph` to FlameGraph Click and Graph Click, respectively, then the backend return "protobuf" and image/svg+xml" to frontend.

pkg/apiserver/profiling/router.go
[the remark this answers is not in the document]
YiniXu9506: I may get your points. You mean we'd better to make the backend API more reasonable, since backend does not return a flamegraph not a protobuf content.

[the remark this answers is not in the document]
breeswish: > I may get your points. You mean we'd better to make the backend API more reasonable, since backend does not return a flamegraph not a protobuf content.

Yes. That's what I mean.

breeswish (approved):

breeswish:
@baurine @shhdgit PTAL

shhdgit (approved):
LGTM! 👍

PR #1089, opened 2021-12-08
Refactor auth feature flag

shhdgit:
This PR add backend protection for non root login.

Merge #1057  first to get a nice diff.


pkg/apiserver/user/shared/feature_flag.go
```diff
@@ -0,0 +1,15 @@
+// Copyright 2021 PingCAP, Inc. Licensed under Apache-2.0.
+
+package shared
+
+import "github.com/pingcap/tidb-dashboard/util/featureflag"
+
+type UserFeatureFlags struct {
+	NonRootLogin *featureflag.FeatureFlag
```
breeswish, line 8: I don't think this is something shared. It is SQL sign-in specific thing. I would rather see SSO to rely on something of SQL sign-in, instead of introducing SQL sign-in for all authenticators.

pkg/apiserver/user/shared/error.go
```diff
@@ -0,0 +1,14 @@
...
+
+import "github.com/joomcode/errorx"
+
+var (
+	ErrNS                  = errorx.NewNamespace("error.api.user")
+	ErrUnsupportedAuthType = ErrNS.NewType("unsupported_auth_type")
+	ErrUnsupportedUser     = ErrNS.NewType("unsupported_user")
+	ErrNSSignIn            = ErrNS.NewSubNamespace("signin")
+	ErrSignInOther         = ErrNSSignIn.NewType("other")
+	ErrInsufficientPrivs   = ErrNSSignIn.NewType("insufficient_priv")
```
breeswish, line 13: This is SQL sign-in specific and should not be put in `shared`

breeswish (commented):
Some further code reorg can be performed for `shared`.

breeswish:

Considering that this is actually a bug fix, it means we need to track it anyway, instead of not picking it up all the time. So here is the tracking issue: https://github.com/pingcap/tidb-dashboard/issues/1100

PR #1091, opened 2021-12-09
Refactor ngm client

shhdgit:
Refactoring to make ngm route more common.


shhdgit:
/cc @breeswish @baurine 
PTAL, thanks~

baurine (approved):
lgtm

util/topo/ngm.go
```diff
@@ -0,0 +1,71 @@
...
+	unixTimestampNano, err := strconv.ParseUint(string(value), 10, 64)
+	if err != nil {
+		return false, ErrInvalidTopologyData.Wrap(err, "NgMonitoring TTL info parse failed")
+	}
+	t := time.Unix(0, int64(unixTimestampNano))
+	if time.Since(t) > time.Second*90 {
+		return false, nil
+	}
+	return true, nil
+}
```
breeswish, line 71: Why do we have to have different topology protocols for different components? We should use the same or limited models for all components, instead of one component one protocol .

@crazycs520

shhdgit:
This pr needs to be refactored after #1098 to make it easier to write unit test.

util/client/ngmclient/client.go
```diff
@@ -0,0 +1,112 @@
+// Copyright 2021 PingCAP, Inc. Licensed under Apache-2.0.
+
+package ngmclient
```
breeswish, line 3: Maybe call it `ngmclient.Proxy` is better, as it is actually prodiving proxies instead of client APIs..

util/client/ngmclient/client.go
```diff
@@ -0,0 +1,112 @@
...
+	"go.etcd.io/etcd/clientv3"
+	"go.uber.org/fx"
+	"golang.org/x/sync/singleflight"
+
+	"github.com/pingcap/tidb-dashboard/pkg/utils/topology"
+	"github.com/pingcap/tidb-dashboard/util/topo"
+)
+
+var (
+	NgmErrNS                 = errorx.NewNamespace("error.api.ng_monitoring")
```
breeswish, line 24: Our new error ns in `util` is meant to be as short and expressive as possible:

```suggestion
	ErrNS                 = errorx.NewNamespace("ngm_client")
```

util/client/ngmclient/client.go
```diff
@@ -0,0 +1,112 @@
...
+	ErrNgMonitoringNotDeploy = NgmErrNS.NewType("not_deploy")
+	ErrNgMonitoringNotStart  = NgmErrNS.NewType("not_start")
```
breeswish, line 26: It's wired to mix ngm and ngmonitoring.

```suggestion
	ErrNgmNotDeployed = NgmErrNS.NewType("ngm_not_deployed")
	ErrNgmNotStarted  = NgmErrNS.NewType("ngm_not_started")
```

util/client/ngmclient/client.go
```diff
@@ -0,0 +1,112 @@
...
+
+		c.Request.URL.Path = targetPath
+
+		ngMonitoringURL, _ := url.Parse(ngMonitoringAddr)
+		proxy := httputil.NewSingleHostReverseProxy(ngMonitoringURL)
+		proxy.ServeHTTP(c.Writer, c.Request)
+	}
+}
+
+func (nc *NgmClient) getNgMonitoringAddrFromCache() (string, error) {
```
breeswish, line 74: It is never a good idea to resolve the address by the client itself. It means your client will have only one limited use case. Usually proving something small, atomic, cohesive and then assembling these things is a better solution. 

Think about cloud, and other dashboard variants. None of them can provide you a resolution in this way. It effectively let your util cannot be used in these variants.

util/client/ngmclient/client.go
```diff
@@ -0,0 +1,112 @@
...
+	address string
+	err     error
+	cacheAt time.Time
+}
+
+type NgmClient struct {
+	lifecycleCtx          context.Context
+	etcdClient            *clientv3.Client
+	ngMonitoringReqGroup  singleflight.Group
+	ngMonitoringAddrCache atomic.Value
```
breeswish, line 43: Not a good idea to mix ngm and ngmonitoring. It only produce more confusion.

util/client/ngmclient/client.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func NewNgmClient(lc fx.Lifecycle, etcdClient *clientv3.Client) (*NgmClient, error) {
```
breeswish, line 46: Better to not relying on fx by default for a util. It makes your util useful in more scenarios.

util/client/ngmclient/client.go
```diff
@@ -0,0 +1,112 @@
...
+
+const (
+	ngMonitoringCacheTTL = time.Second * 5
+)
+
+type ngMonitoringAddrCacheEntity struct {
+	address string
+	err     error
+	cacheAt time.Time
+}
```
breeswish, line 37: Seems that the way you use ngm is a good case of adopting a TopologyProvider in https://github.com/pingcap/tidb-dashboard/pull/1098.

Think about:

```go
type TargetTopoProvider interface {
    GetNgm(ctx context.Context) ([]topo.NgmInfo, error)
}

func NewProxy(target TargetTopoProvider) {
   ...
}

// usage:

ngmclient.NewProxy(myCachedTopologyProvider)
// the developer is free to assemble the proxy using a non-cached topology, 
// or own ngm topo provider! So this makes it useful in more cases.
```

shhdgit: Yeah, it would be better to open a new task after Top SQL PR is finished.

shhdgit: Let me bring back the ngm proxy impl.

shhdgit:
Temporarily closed. It would be better to refactor ngm client after #1099

PR #1092, opened 2021-12-09
Feature Top SQL

shhdgit:
Introducing Top SQL UI.

Merge #1091 first to get a nice diff.

image


shhdgit:
This PR is same as #1099

PR #1094, opened 2021-12-13
support dynamic distro config

baurine:


ui/config-overrides.js
```diff
@@ -17,16 +17,31 @@ const WebpackBar = require('webpackbar')
 const MiniCssExtractPlugin = require('mini-css-extract-plugin')
 const rewireHtmlWebpackPlugin = require('react-app-rewire-html-webpack-plugin')
 
+function copyDistroRes() {
+  const distroResPath = '../bin/distro-res'
+  if (fs.existsSync(distroResPath)) {
+    fs.copySync(distroResPath, './public/distro-res')
```
baurine, line 23: Copy files to the public folder cause unnecessary git changes, this makes the solution not perfect, but I can't find a way to resolve it. But I think it is not that important because this logic only happens when you debug this feature.

util/distro/distro.go
```diff
@@ -62,3 +74,39 @@ func R() *DistributionResource {
 	}
 	return r.(*DistributionResource)
 }
+
+func StringsRes() (distroStringsRes DistributionResource) {
```
breeswish, line 78: As a utility, it should be as much common as possible, to serve the different needs of different environments, and different dashboard variants. In this background, I suggest you to make this function like this:

```go
func ReadResourceStrings(filePath string) (DistributionResource, error) {
    // ...
}
```

ui/public/distro-res/strings.json.example
```diff
@@ -0,0 +1,7 @@
+{
```
breeswish, line 1: I think you can put this example in https://github.com/PingCAP-CBG/tidb-dashboard-distro

ui/lib/utils/distroStringsRes.ts
```diff
@@ -0,0 +1,40 @@
...
+  tiflash: string
+}
+
+const defalutDistroStringsRes: DistroStringsRes = {
+  is_distro: false,
+  tidb: 'TiDB',
+  tikv: 'TiKV',
+  pd: 'PD',
+  tiflash: 'TiFlash',
+}
```
breeswish, line 15: Is it possible to not having this ts definition (types and default values) at all while keep type-check working? In this PR, people needs to modify a lot of places in order to introduce a new distribution string field, while previously what people only need to do is modifying the go source code of `type DistributionResource struct`, writing new fields and default values, and everything will work like charm. They don't need to change here and there.

ui/dashboardApp/layout/signin/index.tsx
```diff
@@ -29,11 +29,11 @@ import { getAuthURL } from '@lib/utils/authSSO'
 import { AuthTypes } from '@lib/utils/auth'
 import { isDistro } from '@lib/utils/i18n'
 import * as auth from '@lib/utils/auth'
+import { useIsFeatureSupport } from '@lib/utils/store'
+import publicPathPrefix from '@lib/utils/publicPathPrefix'
```
breeswish, line 33: Will this work in webpack dev mode?

scripts/go.mod
```diff
@@ -3,12 +3,15 @@ module scripts
...
 	github.com/shurcooL/httpfs v0.0.0-20190707220628-8d4bc4ba7749 // indirect
 	github.com/shurcooL/vfsgen v0.0.0-20200824052919-0d455de96546
 	github.com/swaggo/swag v1.6.6-0.20200529100950-7c765ddd0476
 	go.uber.org/zap v1.19.0
+	golang.org/x/lint v0.0.0-20201208152925-83fdc39ff7b5 // indirect
+	golang.org/x/net v0.0.0-20210405180319-a5a99cb37ef4 // indirect
+	golang.org/x/tools v0.0.0-20210112230658-8b4aab62c064 // indirect
 )
 
 replace github.com/pingcap/tidb-dashboard => ../
```
breeswish, line 17: You can remove this replace as well

scripts/go.mod
```diff
@@ -3,12 +3,15 @@ module scripts
...
+	github.com/pingcap/errors v0.11.5-0.20200917111840-a15ef68f753d // indirect
 	github.com/pingcap/log v0.0.0-20210906054005-afc726e70354
-	github.com/pingcap/tidb-dashboard v0.0.0-00010101000000-000000000000
 	github.com/shurcooL/httpfs v0.0.0-20190707220628-8d4bc4ba7749 // indirect
 	github.com/shurcooL/vfsgen v0.0.0-20200824052919-0d455de96546
 	github.com/swaggo/swag v1.6.6-0.20200529100950-7c765ddd0476
 	go.uber.org/zap v1.19.0
+	golang.org/x/lint v0.0.0-20201208152925-83fdc39ff7b5 // indirect
+	golang.org/x/net v0.0.0-20210405180319-a5a99cb37ef4 // indirect
+	golang.org/x/tools v0.0.0-20210112230658-8b4aab62c064 // indirect
```
breeswish, line 14: Please remove all of these newly added indirect instructions. I guess they are not needed. If you find it hard to remove, follow these steps:

1. remove `go.sum`
2. remove all `// indirect` here
3. run `go mod tidy` to generate a new `go.sum`

I usually find it working with the steps above.

scripts/go.mod
```diff
@@ -3,12 +3,15 @@ module scripts
 go 1.13
 
 require (
+	github.com/pingcap/errors v0.11.5-0.20200917111840-a15ef68f753d // indirect
 	github.com/pingcap/log v0.0.0-20210906054005-afc726e70354
-	github.com/pingcap/tidb-dashboard v0.0.0-00010101000000-000000000000
```
breeswish, line 7: This is a very nice elimination!

ui/public/index.html
```diff
@@ -3,12 +3,8 @@
   <head>
     <meta charset="utf-8" />
     <meta name="x-public-path-prefix" content="__PUBLIC_PATH_PREFIX__" />
-    <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
+    <link rel="icon" href="%PUBLIC_URL%/distro-res/favicon.ico" />
```
breeswish, line 6: But where is `__DISTRO_STRINGS_RES__`?

ui/config-overrides.js
```diff
@@ -17,16 +17,31 @@ const WebpackBar = require('webpackbar')
...
+  }
+}
+
 function injectDistroToHTML(config, env) {
```
breeswish, line 27: Deserves some comments describing the behavior, since it is different in dev mode and production mode.

ui/config-overrides.js
breeswish: Looks fine to me, since there will be git changes from the very beginning implementation of @shhdgit .

scripts/distro/write_strings.sh
```diff
@@ -1,25 +0,0 @@
...
-BUILD_TAG=""
-
-if [[ -f "${PROJECT_DIR}/internal/resource/distrores/strings.go" ]]; then
-  echo "+ Existing distribution resource is detected, using it to write strings"
-  BUILD_TAG=distro
-fi
-
-echo "+ Write resource strings"
-cd "$PROJECT_DIR"
-# FIXME: distro/write_strings needs to access the /internal package, which is not allowed to be invoked in another module
```
breeswish, line 18: Nice to see that this hack is successfully removed!

ui/public/index.html
```diff
@@ -90,7 +86,6 @@
         }
       }
     </style>
-    <title><%= htmlWebpackPlugin.options.distro_tidb %> Dashboard</title>
```
breeswish, line 93: Titles are moved. Please make sure that correct titles are set later in JS.

ui/lib/utils/distroStringsRes.ts
```diff
@@ -0,0 +1,40 @@
...
+    distro = {
+      ...defalutDistroStringsRes,
+      ...distroObj,
+    }
+  } catch (error) {
+    console.log(error)
+  }
+}
+
+document.title = `${distro.tidb} Dashboard`
```
breeswish, line 36: It may not be a good idea to modify the title in a util.

pkg/uiserver/uiserver.go
```diff
@@ -53,6 +61,63 @@ func RewriteAssets(fs http.FileSystem, cfg *config.Config, updater UpdateContent
...
+		targetFile, err := fs.Open(assetPath)
+		if err != nil {
+			// has no target asset to be overried, skip
+			return
+		}
+		defer targetFile.Close()
+
+		sourceFile, err := os.Open(path.Join(distroResDir, assetName))
+		if err != nil {
+			log.Fatal("Failed to open source file", zap.String("path", assetName), zap.Error(err))
```
breeswish, line 92: log.Fatal will crash the server. Please don't.

pkg/uiserver/uiserver.go
```diff
@@ -53,6 +61,63 @@ func RewriteAssets(fs http.FileSystem, cfg *config.Config, updater UpdateContent
...
+
+	distroResDir := path.Join(path.Dir(exePath), distro.DistroResFolderName)
+	info, err := os.Stat(distroResDir)
+	if err != nil || !info.IsDir() {
+		// just ignore
+		return
+	}
+
+	override := func(assetName string) {
+		assetPath := path.Join("/", distro.DistroResFolderName, assetName)
```
breeswish, line 82: If I remember correctly, some distributions will not substitute all files. You may need to be compatible with the behavior.

pkg/uiserver/uiserver.go
```diff
@@ -53,6 +61,63 @@ func RewriteAssets(fs http.FileSystem, cfg *config.Config, updater UpdateContent
...
 	rewrite("/index.html")
 	rewrite("/diagnoseReport.html")
+
+	overrideDistroAssetsRes(fs, cfg, updater)
+}
+
+func overrideDistroAssetsRes(fs http.FileSystem, cfg *config.Config, updater UpdateContentFunc) {
+	exePath, err := os.Executable()
+	if err != nil {
+		log.Fatal("Failed to get work dir", zap.Error(err))
```
breeswish, line 71: This is a executable path instead of working dir.

pkg/uiserver/uiserver.go
```diff
@@ -53,6 +61,63 @@ func RewriteAssets(fs http.FileSystem, cfg *config.Config, updater UpdateContent
...
+	}
+
+	distroResDir := path.Join(path.Dir(exePath), distro.DistroResFolderName)
+	info, err := os.Stat(distroResDir)
+	if err != nil || !info.IsDir() {
+		// just ignore
+		return
+	}
+
+	override := func(assetName string) {
```
breeswish, line 81: How about moving this function out? It would be better to be not a lambda function.

pkg/uiserver/uiserver.go
breeswish, line 92: I don't think it is a good behavior to panic everywhere in your logic. previous `rewrite` function are purely based on *generated embedded assets*, which has an *expected content* and *should never miss any files*. So that it is fine to panic when there are vfs IO issues because they *are unexpected programming errors*. However in distro I don't think this still stands. For example, failed to read the dynamic distro file should be tolerated, as it is a file exists on user's FS and it is very possible that the read is failed (e.g. wrong permissions).

breeswish (commented):
the rest LGTM

breeswish:
@shhdgit @YiniXu9506 PTAL, thanks!

ui/dashboardApp/layout/signin/index.tsx
baurine: Yep, I have tested it.

scripts/go.mod
baurine: Got it.

ui/public/index.html
baurine: It is generated when building.

ui/lib/utils/distroStringsRes.ts
baurine: Make sense.

pkg/uiserver/uiserver.go
baurine: Oh, yep, the error message is not correct.

pkg/uiserver/uiserver.go
baurine: Yep, they just need to supply the files that they want to replace, and we only replace the files they supply.

pkg/uiserver/uiserver.go
baurine: OK. I thought they should crash as early as possible if the distribution config is not correct, so they won't get the wrong brand names/logos.

ui/public/index.html
breeswish: I mean seems that there is no source string for replacement?

pkg/uiserver/uiserver.go
breeswish: Cool!

util/distro/distro.go
baurine: Make sense!

ui/config-overrides.js
```diff
@@ -17,16 +17,31 @@ const WebpackBar = require('webpackbar')
...
-  const distroInfo = Object.entries(require('./lib/distribution.json')).reduce(
-    (prev, [k, v]) => {
-      return {
-        ...prev,
-        [`distro_${k}`]: v,
-      }
+  let distroStringsResMeta = '__DISTRO_STRINGS_RES__'
```
baurine, line 28: hi @breeswish , the `__DISTRO_STRINGS_RES__` source string is here.

ui/lib/utils/distroStringsRes.ts
baurine: I have an idea, let me try.

ui/lib/utils/distroStringsRes.ts
```diff
@@ -0,0 +1,32 @@
+import { DistroDistributionResource } from '@lib/client'
+
+const defalutDistroStringsRes: DistroDistributionResource = {
+  is_distro: false,
+  tidb: 'TiDB',
+  tikv: 'TiKV',
+  pd: 'PD',
+  tiflash: 'TiFlash',
+}
```
baurine, line 9: Have no idea to dismiss this part duplicated code yet.

pkg/uiserver/uiserver.go
```diff
@@ -53,6 +66,61 @@ func RewriteAssets(fs http.FileSystem, cfg *config.Config, updater UpdateContent
...
+}
+
+func overrideDistroAssetsRes(fs http.FileSystem, distroResFolderPath string, updater UpdateContentFunc) {
+	info, err := os.Stat(distroResFolderPath)
+	if errors.Is(err, os.ErrNotExist) || !info.IsDir() {
+		// just ignore if the folder doesn't exist or it's not a folder
+		return
+	}
+	if err != nil {
+		log.Fatal("Failed to stat dir", zap.String("dir", distroResFolderPath), zap.Error(err))
```
breeswish, line 80: Have you confirmed that crash is the accepted behavior?

breeswish (commented):
The rest LGTM. Remember to supply tests later.

internal/resource/distrores/populate.go
```diff
@@ -3,9 +3,31 @@
...
+		log.Fatal("Failed to get executable path", zap.Error(err))
+	}
+
+	distroStringsResPath := path.Join(path.Dir(exePath), distroResFolderName, distroStringsResFileName)
+	distroStringsRes, err := distro.ReadResourceStringsFromFile(distroStringsResPath)
+	if err != nil {
+		log.Fatal("Failed to read distro strings res", zap.String("path", distroStringsResPath), zap.Error(err))
+	}
+
+	distro.ReplaceGlobal(distroStringsRes)
```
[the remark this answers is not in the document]
baurine: Got it.

pkg/uiserver/uiserver.go
baurine: No reply yet, I will DM alex soon.

ui/lib/utils/distroStringsRes.ts
baurine: Let me try it.

scripts/write_distro_strings.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+# This script outputs distribution strings to json which is required for the frontend.
+
+set -euo pipefail
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+PROJECT_DIR=$(cd "$DIR/.."; pwd)
+BUILD_TAG=""
```
baurine, line 9: TODO: remove

ui/lib/distro_strings.json
```diff
@@ -0,0 +1 @@
+{ "tidb": "TiDB", "tikv": "TiKV", "pd": "PD", "tiflash": "TiFlash" }
```
baurine, line 1: This file is auto-generated.

cmd/distro/write_strings.go
```diff
@@ -10,15 +10,14 @@ import (
...
-	_ "github.com/pingcap/tidb-dashboard/internal/resource/distrores"
 	"github.com/pingcap/tidb-dashboard/util/distro"
 )
 
 func main() {
 	outputPath := flag.String("o", "", "Output json file path")
 	flag.Parse()
 
-	d, _ := json.Marshal(distro.R())
+	d, _ := json.Marshal(distro.DefaultDistroRes)
```
baurine, line 20: Move this file to cmd/distro/write_strings.go, so the `scripts` doesn't need to depend on tidb-dashboard, not sure whether it is a good idea.

breeswish: I think it is not a good idea. It is not a runnable binary that we ever want to provide to the user, so `cmd` is not a right place for it.

Actually it's fine to make `scripts/go.mod` depending on the tidb-dashboard, as bumping the `scripts/go.sum` is as easy as running a `go mod tidy`, which can be automated and encapsulated in something like scripts.

util/distro/distro.go
```diff
@@ -20,7 +24,7 @@ type DistributionResource struct {
 	TiFlash  string `json:"tiflash,omitempty"`
 }
 
-var defaultDistroRes = DistributionResource{
+var DefaultDistroRes = DistributionResource{
```
breeswish, line 27: There is no need to expose `DefaultDistroRes` to the external, as `R()` already provides it.

util/distro/distro.go
```diff
@@ -58,7 +62,37 @@ func ReplaceGlobal(r DistributionResource) func() {
...
+func ReadResourceStringsFromFile(filePath string) (DistributionResource, error) {
+	distroStringsRes := DistributionResource{}
+
+	info, err := os.Stat(filePath)
+	if errors.Is(err, os.ErrNotExist) || info.IsDir() {
+		// ignore if file not exist or it is a folder
+		return distroStringsRes, nil
+	}
+	if err != nil {
+		// may permission or other errors
```
breeswish, line 79:
```suggestion
		// may be permission-like errors
```

ui/config-overrides.js
```diff
@@ -183,10 +204,10 @@ module.exports = override(
...
   ),
   disableMinimizeByEnv(),
-  addExtraEntries(),
   supportDynamicPublicPathPrefix(),
   overrideProcessEnv({
     REACT_APP_RELEASE_VERSION: JSON.stringify(getInternalVersion()),
   }),
-  injectDistroToHTML
+  injectDistroToHTML,
+  addExtraEntries()
```
baurine, line 212: make sure `addExtraEntries()` run before `injectDistroToHtml` (the running order is reverse), else `injectDistroToHtml` only works for index.html, but not diagnoseReport.html

breeswish (commented):
the rest LGTM

util/distro/distro.go
baurine: Oh,  yep, I got it, because the default distro res won't be replaced by customized distro res when we only run the write_strings.go

breeswish (approved):

breeswish:
Remember to also provide a PD PR after this is released.

ui/config-overrides.js
shhdgit: How about adding this comment in the code.

breeswish:
We don't need to take care of the diagnostics report and the metrics. They are out of our scope.

shhdgit:
The rest LGTM.

cmd/tidb-dashboard/main.go
```diff
@@ -147,6 +148,26 @@ func buildTLSConfig(caPath, keyPath, certPath *string) *tls.Config {
 	return tlsConfig
 }
 
+const (
+	distroResFolderName      string = "distro-res"
+	distroStringsResFileName string = "strings.json"
+)
+
+func loadDistroStringsRes() {
```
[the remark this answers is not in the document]
baurine: They are in there previously.

shhdgit (approved):

baurine:
PD side PR: https://github.com/tikv/pd/pull/4502

PR #1095, opened 2021-12-13
chore: support multiple profiling types

YiniXu9506:
### What did
1. Add multi-select for profiling types: CPU, Heap, Goroutine, Mutex
2. Update taskGroupState, the logic is:
    - there has error task in taskGroup, the groupState is Error
    - there has no error task, the groupState is Finished even if there exists skippedTasks.
3. Groupby tasks in profiling detail page. 
4. Skip Heap, Goroutine and Mutex profiling for TiKV/TiFlash.

Result can be previewed: 

image
image


breeswish:
Cool! How about adding a (?) icon and some tooltip explanation to the skipped state to tell user the reason that it skipped?

pkg/apiserver/profiling/model.go
```diff
@@ -20,21 +20,35 @@ const (
...
-	RawDataType TaskRawDataType         `json:"raw_data_type"`
+	ID            uint                    `json:"id" gorm:"primary_key"`
+	TaskGroupID   uint                    `json:"task_group_id" gorm:"index"`
+	State         TaskState               `json:"state" gorm:"index"`
+	Target        model.RequestTargetNode `json:"target" gorm:"embedded;embedded_prefix:target_"`
+	FilePath      string                  `json:"-" gorm:"type:text"`
+	Error         string                  `json:"error" gorm:"type:text"`
+	StartedAt     int64                   `json:"started_at"` // The start running time, reset when retry. Used to estimate approximate profiling progress.
+	RawDataType   TaskRawDataType         `json:"raw_data_type" gorm:"raw_data_type"`
+	ProfilingType string                  `json:"profiling_type"`
```
breeswish, line 51:
```suggestion
	ProfilingType TaskProfilingType       `json:"profiling_type"`
```

pkg/apiserver/profiling/model.go
```diff
@@ -67,14 +82,15 @@ type Task struct {
 }
 
 // NewTask creates a new profiling task.
-func NewTask(ctx context.Context, taskGroup *TaskGroup, target model.RequestTargetNode, fts *fetchers) *Task {
+func NewTask(ctx context.Context, taskGroup *TaskGroup, target model.RequestTargetNode, fts *fetchers, profilingType string) *Task {
```
breeswish, line 85:
```suggestion
func NewTask(ctx context.Context, taskGroup *TaskGroup, target model.RequestTargetNode, fts *fetchers, profilingType TaskProfilingType) *Task {
```

pkg/apiserver/profiling/model.go
```diff
@@ -84,11 +100,15 @@ func NewTask(ctx context.Context, taskGroup *TaskGroup, target model.RequestTarg
...
 
 func (t *Task) run() {
-	fileNameWithoutExt := fmt.Sprintf("profiling_%d_%d_%s", t.TaskGroupID, t.ID, t.Target.FileName())
-	protoFilePath, rawDataType, err := profileAndWritePprof(t.ctx, t.fetchers, &t.Target, fileNameWithoutExt, t.taskGroup.ProfileDurationSecs)
+	fileNameWithoutExt := fmt.Sprintf("profiling_%d_%d_%s_%s", t.TaskGroupID, t.ID, t.ProfilingType, t.Target.FileName())
+	protoFilePath, rawDataType, err := profileAndWritePprof(t.ctx, t.fetchers, &t.Target, fileNameWithoutExt, t.taskGroup.ProfileDurationSecs, t.ProfilingType)
 	if err != nil {
 		t.Error = err.Error()
-		t.State = TaskStateError
+		if err.Error() == "unsupported profiling type" {
```
breeswish, line 107: It is not a good practice to test error against its message. You can use errorx to make a typed error and use [`errorx.IsOfType(..)`](https://pkg.go.dev/github.com/joomcode/errorx#IsOfType) to test against it.

pkg/apiserver/profiling/model.go
```diff
@@ -84,11 +100,15 @@ func NewTask(ctx context.Context, taskGroup *TaskGroup, target model.RequestTarg
 }
 
 func (t *Task) run() {
-	fileNameWithoutExt := fmt.Sprintf("profiling_%d_%d_%s", t.TaskGroupID, t.ID, t.Target.FileName())
-	protoFilePath, rawDataType, err := profileAndWritePprof(t.ctx, t.fetchers, &t.Target, fileNameWithoutExt, t.taskGroup.ProfileDurationSecs)
+	fileNameWithoutExt := fmt.Sprintf("profiling_%d_%d_%s_%s", t.TaskGroupID, t.ID, t.ProfilingType, t.Target.FileName())
+	protoFilePath, rawDataType, err := profileAndWritePprof(t.ctx, t.fetchers, &t.Target, fileNameWithoutExt, t.taskGroup.ProfileDurationSecs, t.ProfilingType)
 	if err != nil {
 		t.Error = err.Error()
```
breeswish, line 106: When task is skipped, maybe there is no need to set `t.Error`, since it is a normal and expected state, instead of a actual error state.

pkg/apiserver/profiling/model.go
```diff
@@ -109,13 +129,14 @@ type TaskGroup struct {
 }
 
 // NewTaskGroup create a new profiling task group.
-func NewTaskGroup(db *dbstore.DB, profileDurationSecs uint, stats model.RequestTargetStatistics) *TaskGroup {
+func NewTaskGroup(db *dbstore.DB, profileDurationSecs uint, stats model.RequestTargetStatistics, profilingTypeList string) *TaskGroup {
```
breeswish, line 132: Maybe you can use this to make your life easier:

```go
type ProfilingTypeList []TaskProfilingType

func (r *ProfilingTypeList) Scan(src interface{}) error {
	return json.Unmarshal([]byte(src.(string)), r)
}

func (r ProfilingTypeList) Value() (driver.Value, error) {
	val, err := json.Marshal(r)
	return string(val), err
}

// ...

type TaskGroupModel struct {
	...
	RequestedProfilingTypes  ProfilingTypeList `json:"requested_profiling_types"`
	...
}
```

It relies on customized serialization / deserialization so that your model is kept to be structured.

pkg/apiserver/profiling/model.go
```diff
@@ -47,6 +61,7 @@ type TaskGroupModel struct {
 	ProfileDurationSecs uint                          `json:"profile_duration_secs"`
 	TargetStats         model.RequestTargetStatistics `json:"target_stats" gorm:"embedded;embedded_prefix:target_stats_"`
 	StartedAt           int64                         `json:"started_at"`
+	ProfilingTypeList   string                        `json:"profiling_type_list"`
```
breeswish, line 64: Better to add a comment for it, documenting how the list is stored in a string. For example:

```go
// ProfilingTypeList is a comma separated list of profiling types requested by the user.
// Each profiling type is ensured to be a variant of `TaskProfilingType`.
```

pkg/apiserver/profiling/model.go
```diff
@@ -20,33 +24,60 @@ const (
 	TaskStateRunning
 	TaskStateFinish
 	TaskPartialFinish
+	TaskSkipped
```
breeswish, line 27:
```suggestion
	TaskStatePartialFinish // Only valid for task group
	TaskStateSkipped
```

pkg/apiserver/profiling/model.go
```diff
@@ -84,11 +116,15 @@ func NewTask(ctx context.Context, taskGroup *TaskGroup, target model.RequestTarg
...
 
 func (t *Task) run() {
-	fileNameWithoutExt := fmt.Sprintf("profiling_%d_%d_%s", t.TaskGroupID, t.ID, t.Target.FileName())
-	protoFilePath, rawDataType, err := profileAndWritePprof(t.ctx, t.fetchers, &t.Target, fileNameWithoutExt, t.taskGroup.ProfileDurationSecs)
+	fileNameWithoutExt := fmt.Sprintf("profiling_%d_%d_%s_%s", t.TaskGroupID, t.ID, t.ProfilingType, t.Target.FileName())
+	protoFilePath, rawDataType, err := profileAndWritePprof(t.ctx, t.fetchers, &t.Target, fileNameWithoutExt, t.taskGroup.ProfileDurationSecs, t.ProfilingType)
 	if err != nil {
-		t.Error = err.Error()
-		t.State = TaskStateError
+		if errorx.IsOfType(err, ErrTaskSikpped) {
```
breeswish, line 122: From the "profileAndWritePprof" perspective, it should be some kind of unsupported profiling type error, which then result in the task to be skipped.

```suggestion
		if errorx.IsOfType(err, errUnsupportedProfilingType) {
```

pkg/apiserver/profiling/profile.go
```diff
@@ -9,16 +9,26 @@ import (
...
+func skipTask(profilingType string) (string, TaskRawDataType, error) {
+	return "nil", "", ErrTaskSikpped.New("task_skipped")
+}
+
+func profileAndWritePprof(ctx context.Context, fts *fetchers, target *model.RequestTargetNode, fileNameWithoutExt string, profileDurationSecs uint, profilingType TaskProfilingType) (string, TaskRawDataType, error) {
 	switch target.Kind {
 	case model.NodeKindTiKV:
-		return fetchPprof(&pprofOptions{duration: profileDurationSecs, fileNameWithoutExt: fileNameWithoutExt, target: target, fetcher: &fts.tikv})
+		if string(profilingType) != string(ProfilingTypeCPU) {
+			return skipTask(string(profilingType))
```
breeswish, line 20:
```suggestion
			return "", "", errUnsupportedProfilingType.NewWithNoMessage()
```

pkg/apiserver/profiling/profile.go
```diff
@@ -9,16 +9,26 @@ import (
...
-func profileAndWritePprof(ctx context.Context, fts *fetchers, target *model.RequestTargetNode, fileNameWithoutExt string, profileDurationSecs uint) (string, TaskRawDataType, error) {
+func skipTask(profilingType string) (string, TaskRawDataType, error) {
+	return "nil", "", ErrTaskSikpped.New("task_skipped")
+}
+
+func profileAndWritePprof(ctx context.Context, fts *fetchers, target *model.RequestTargetNode, fileNameWithoutExt string, profileDurationSecs uint, profilingType TaskProfilingType) (string, TaskRawDataType, error) {
 	switch target.Kind {
 	case model.NodeKindTiKV:
-		return fetchPprof(&pprofOptions{duration: profileDurationSecs, fileNameWithoutExt: fileNameWithoutExt, target: target, fetcher: &fts.tikv})
+		if string(profilingType) != string(ProfilingTypeCPU) {
```
breeswish, line 19:
```suggestion
		// TiKV only supports CPU Profiling
		if profilingType != ProfilingTypeCPU {
```

pkg/apiserver/profiling/service.go
```diff
@@ -150,18 +152,21 @@ func (s *Service) exclusiveExecute(ctx context.Context, req *StartRequest) (*Tas
...
 	}
 
 	tasks := make([]*Task, 0, len(req.Targets))
 	for _, target := range req.Targets {
-		t := NewTask(ctx, taskGroup, target, s.fetchers)
-		s.params.LocalStore.Create(t.TaskModel)
-		s.tasks.Store(t.ID, t)
-		tasks = append(tasks, t)
+		profileTypeList := req.RequstedProfilingTypes
+		for _, profilingType := range profileTypeList {
```
breeswish, line 164: better to check whether `profileTypeList` is a valid ProfilingType (since JSON deserialization will never check enum variants. Bad user may pass in "foo" as a profilingType). Otherwise bad request should be thrown.

breeswish (commented):
the rest LGTM

breeswish:
@baurine @shhdgit PTAL

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -325,6 +389,10 @@ export default function Page() {
             columns={columns}
             items={data?.tasks_status || []}
```
baurine, line 390: I think we should use `newRows` as items, because the the `startIndex` filed of `groups` depends on the data in the `newRows` instead of the `data?.tasks_status`.

We can have an experiment, use `['tifalsh', 'tidb', 'tikv', 'pd']` replace `['pd', 'tidb', 'tikv', 'tiflash']` in the line 244, then we will get wrong result:

image

`2379` is the PD component port.

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -217,22 +272,13 @@ export default function Page() {
...
-        },
-      },
       {
         name: t('instance_profiling.detail.table.columns.content'),
         key: 'content',
         minWidth: 150,
         maxWidth: 250,
         onRender: (record) => {
-          return `CPU Profiling - ${profileDuration}s`
+          return `${record.profiling_type} - ${profileDuration}s`
```
baurine, line 281: memory, mutex, goroutine profiles have no duration.

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -153,29 +118,86 @@ function ViewResultButton({ rec, t }) {
...
+          ViewOptions.Graph
+        }`
+
+      window.open(`${profileURL}`, '_blank')
+      break
+    case ViewOptions.Text:
+      token = await getActionToken(rec.id, 'single_view')
+      if (!token) {
+        return
+      }
```
baurine, line 182: This part code duplicates 4 times, we can put it before switch-case

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -153,29 +118,86 @@ function ViewResultButton({ rec, t }) {
...
+    case ViewOptions.FlameGraph:
+      token = await getActionToken(rec.id, 'single_view')
+      if (!token) {
+        return
+      }
+
+      profileURL = `${client.getBasePath()}/profiling/single/view?token=${token}`
+      if (isProtobuf) {
+        const titleOnTab = rec.target.kind + '_' + rec.target.display_name
+        profileURL = `/dashboard/speedscope#profileURL=${encodeURIComponent(
```
baurine, line 157: I think `/dashboard` should use `publicPathPrefix` variable.

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -203,11 +217,44 @@ export default function Page() {
...
+        key: InstanceKindName[instanceKind],
+        name: InstanceKindName[instanceKind],
+        startIndex: startIndex,
+        count: newRows.length - startIndex,
+      })
+      startIndex = newRows.length
+    }
+    setRows(newRows)
+    setGroups(newGroups)
+  }, [data])
```
baurine, line 256: It's can be replaced by `useMemo()`

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -153,29 +119,75 @@ function ViewResultButton({ rec, t }) {
...
+  const isProtobuf = rec.raw_data_type === RawDataType.Protobuf
+  let token: string | undefined = await getActionToken(rec.id, 'single_view')
+  if (!token) {
+    return
+  }
+  let profileURL: string
+
+  switch (openAs) {
+    case ViewOptions.Download:
+      token = await getActionToken(rec.id, 'single_download')
```
baurine, line 145: oh, sorry, I didn't notice this difference to get token, so the former logic is fine.

pkg/apiserver/profiling/service.go
```diff
@@ -25,14 +25,17 @@ const (
 )
 
 var (
-	ErrNS             = errorx.NewNamespace("error.profiling")
-	ErrIgnoredRequest = ErrNS.NewType("ignored_request")
-	ErrTimeout        = ErrNS.NewType("timeout")
+	ErrNS                         = errorx.NewNamespace("error.profiling")
```
shhdgit, line 28: Find a legacy typo
`errorx.NewNamespace("error.api.profiling")`

pkg/apiserver/profiling/model.go
```diff
@@ -19,34 +23,80 @@ const (
...
+	ProfilingTypeGoroutine TaskProfilingType = "goroutine"
+	ProfilingTypeMutex     TaskProfilingType = "mutex"
+)
+
+var profilingTypeList = []TaskProfilingType{
+	ProfilingTypeCPU,
+	ProfilingTypeHeap,
+	ProfilingTypeGoroutine,
+	ProfilingTypeMutex,
+}
```
shhdgit, line 63: How about using profiling type map like:
```
var profilingTypeMap = map[TaskProfilingType]struct{}{
	ProfilingTypeCPU: {},
	ProfilingTypeHeap: {},
	ProfilingTypeGoroutine: {},
	ProfilingTypeMutex: {},
}
```

So that we can easily check the valid profiling type
```
_, valid := profilingTypeMap[profilingType]
...
```

shhdgit (commented):
Rest LGTM.

baurine (commented):
lgtm

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -87,56 +98,11 @@ async function getActionToken(
...
-        profileURL =
-          profileURL = `${client.getBasePath()}/profiling/single/view?token=${token}&output_type=${
-            ViewOption.Graph
-          }`
-
-        window.open(`${profileURL}`, '_blank')
-    }
-  })
-
+function DropdownButton(props) {
```
breeswish, line 101: Is it possible to assign strong types for them? (for example, the type of `rec`). Currently they are all slightly fell into `unknown`, which is not a good practice.

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -153,29 +119,84 @@ function ViewResultButton({ rec, t }) {
...
-    return <DropdownButton />
+    return <DropdownButton rec={rec} t={t} />
   } else {
     return (
       <>
         {rec.state === taskState.Success && (
           <Button
             disabled={rec.state !== taskState.Success}
-            onClick={() => openResult(`${rec.view_options[0]}`)}
+            onClick={() => openResult(`${rec.view_options[0]}`, rec)}
```
breeswish, line 199:
```suggestion
            onClick={() => openResult(rec.view_options[0], rec)}
```

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -87,56 +98,11 @@ async function getActionToken(
...
-        window.open(`${profileURL}`, '_blank')
-    }
-  })
-
+function DropdownButton(props) {
+  const { rec, t } = props
   const menu = () => {
     return (
-      <Menu onClick={(e) => openResult(e.key as string)}>
+      <Menu onClick={(e) => openResult(e.key as string, rec)}>
```
breeswish, line 105: You can now assign a strong enum type for `openResult` I guess.

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -153,29 +119,84 @@ function ViewResultButton({ rec, t }) {
...
-        {t(
-          `instance_profiling.detail.table.columns.selection.types.${rec.view_options[0]}`
-        )}
-      </Dropdown.Button>
-    )
+  return (
+    <Dropdown.Button
+      disabled={rec.state !== taskState.Success}
+      overlay={menu}
+      onClick={() => openResult(`${rec.view_options[0]}`, rec)}
```
breeswish, line 126:
```suggestion
      onClick={() => openResult(rec.view_options[0], rec)}
```

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -87,56 +98,11 @@ async function getActionToken(
...
-          profileURL = `${client.getBasePath()}/profiling/single/view?token=${token}&output_type=${
-            ViewOption.Graph
-          }`
-
-        window.open(`${profileURL}`, '_blank')
-    }
-  })
-
+function DropdownButton(props) {
+  const { rec, t } = props
```
breeswish, line 102: You can use the hook instead of passing `t` around in a component:

```suggestion
function DropdownButton({ rec }) {
  const { t } = useTranslation()
```

breeswish (approved):
The rest LGTM

baurine (approved):
the rest lgtm

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
baurine: I extracted the ActionsButton to the lib/components in PR #1102 , we can use this component later.

ui/lib/apps/InstanceProfiling/pages/Detail.tsx
```diff
@@ -257,6 +155,51 @@ export default function Page() {
     return [newRows, newGroups]
   }, [data])
 
+  const openResult = usePersistFn(
+    async (openAs: string, rec: ActionButtonProps) => {
```
breeswish, line 159: Better to call it something like `IRecord`. It is the ActionButton that relying on record's data, instead of the record relying on the action button's data.

PR #1098, opened 2021-12-13
Refactor: Add TopologyProvider

breeswish:
To resolve https://github.com/pingcap/tidb-dashboard/pull/1050 in new utils elegantly, an interface called `TopologyProvider` is introduced. Over which, there are two implementors:

- `TopologyFromPD`: A topology provider that provide topology from PD
- `CachedTopology`: A topology provider that caches the result of the underlying topology provider


util/topo/provider_cached.go
```diff
@@ -0,0 +1,123 @@
...
+		return data, nil
+	}
+	// TODO: use singleflight.
+	src, err := backSource()
+	if err != nil {
+		// Error is never cached.
+		return nil, err
+	}
+	_ = c.cache.Set(key, src)
+	runtime.KeepAlive(c)
```
shhdgit, line 47: Ok...This is too elaborate. Then we also need to be careful about using `c.cache` outside of `getOrFillCache` in the future.

shhdgit: Probably won't use `c.cache` anymore.

shhdgit: Cool, knowledge++

etc/go.mod
```diff
@@ -0,0 +1,3 @@
+module ignore_etc // a hack to ignore this directory in go commands
```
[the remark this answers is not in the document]
breeswish: See https://github.com/golang/go/issues/30058

util/client/pdclient/pd_api_highlevel.go
```diff
@@ -5,12 +5,15 @@
...
+	"context"
 	"sort"
 	"strings"
 )
 
-func (api *APIClient) HLGetStores() ([]GetStoresResponseStore, error) {
-	resp, err := api.GetStores()
+// HLGetStores returns all stores in PD in order.
+// An optional ctx can be passed in to override the default context. To keep the default context, pass nil.
+func (api *APIClient) HLGetStores(ctx context.Context) ([]GetStoresResponseStore, error) {
```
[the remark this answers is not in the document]
breeswish: No. Context is always required. The full name of the parameter is `overrideCtx context.Context` here which makes it optional. Nil means not override and use the context specified in the client.

util/topo/provider_cached.go
breeswish: Yeah. Probably I should add a comment: Don't touch this unless you are professional about `runtime.SetFinalizer` behavior.. LOL

shhdgit (approved):

PR #1102, opened 2021-12-16
Refine conprof

baurine:
What did:

- update wording
- rename i18n key "continue_profiling" to "conprof"
- update check ngm ready logic
- replace the time range component
- align the detail page UI and UX with manual profiling detail page

image

image

image


pkg/apiserver/conprof/service.go
```diff
@@ -111,7 +111,8 @@ func (s *Service) reverseProxy(targetPath string) gin.HandlerFunc {
 				_ = c.Error(rest.ErrBadRequest.WrapWithNoMessage(err))
 				return
 			}
-			c.Request.URL.RawQuery = queryStr
+
+			c.Request.URL.RawQuery = strings.Replace(c.Request.URL.RawQuery, "token="+token, queryStr, 1)
```
baurine, line 115: Refine this part logic, after parsing the token , we only replace the token part, and keep other query params.

/cc @shhdgit

ui/lib/apps/ContinuousProfiling/pages/Detail.tsx
```diff
@@ -94,10 +201,28 @@ export default function Page() {
...
+        const profileURL = `${client.getBasePath()}/continuous_profiling/single_profile/view?token=${token}&data_format=protobuf`
+        const speedscopeURL = `${publicPathPrefix}/speedscope#profileURL=${encodeURIComponent(
+          profileURL
+        )}&title=${speedscopeTitle}`
+        window.open(speedscopeURL, '_blank')
+        return
+      }
+
+      if (action === 'download') {
+        window.location.href = `${client.getBasePath()}/continuous_profiling/download?token=${token}`
```
baurine, line 223: @crazycs520 will help implement this API in the ngm side (it doesn't support download single profile before).

ui/lib/apps/ContinuousProfiling/pages/List.tsx
```diff
@@ -65,42 +98,41 @@ export default function Page() {
...
+                text={t('conprof.list.table.status.failed')}
               />
             )
           } else if (rec.state === 'success') {
             // all success
             return (
               <Badge
                 status="success"
-                text={t('continuous_profiling.list.table.status.finished')}
+                text={t('conprof.list.table.status.finished')}
```
baurine, line 120: Update the state after PR https://github.com/pingcap/ng-monitoring/pull/50 is merged.

pkg/apiserver/conprof/service.go
shhdgit: How about use `URL.Values` to ser/deser the query values.

ui/lib/apps/ContinuousProfiling/pages/List.tsx
```diff
@@ -75,32 +108,30 @@ export default function Page() {
             return (
               <Badge
                 status="error"
-                text={t('continuous_profiling.list.table.status.failed')}
+                text={t('conprof.list.table.status.failed')}
               />
             )
           } else if (rec.state === 'success') {
```
baurine, line 114: Update the state after PR pingcap/ng-monitoring#50 is merged.

pkg/apiserver/conprof/service.go
baurine: I found a way that doesn't need to change this part logic, so I reverted the changes.

breeswish:
Cool! Could you prepare a preview (merging with changes from https://github.com/pingcap/tidb-dashboard/pull/1095) so that @lilyjazz can evaluate and provide some UI design comments?

baurine:
> Cool! Could you also prepare a preview (merging with changes from #1095) so that @lilyjazz can evaluate and provide some UI design comments? Thanks a lot!

OK

ui/lib/apps/ContinuousProfiling/pages/List.tsx
baurine: Updated:

image

baurine:
@breeswish @shhdgit PTAL, thanks!

ui/lib/components/ActionsButton/index.tsx
```diff
@@ -21,11 +21,6 @@ export default function ActionsButton({
     throw new Error('actions should at least have one action')
   }
 
-  // hide actions if the task does not succceed
-  if (disabled) {
```
YiniXu9506, line 25: @breeswish recommended to hide the actions if the task does not succeed. It would be better to make the behavior consistent with the one on manual profiling. But you can still keep the display way until PMs want to change it.

baurine: OK, got it.

ui/lib/apps/ContinuousProfiling/pages/Detail.tsx
```diff
@@ -58,53 +106,99 @@ export default function Page() {
...
+      },
+      {
+        name: t('conprof.detail.table.columns.actions'),
+        key: 'actions',
+        minWidth: 150,
+        maxWidth: 200,
+        onRender: (record) => {
+          const rec = record as ConprofProfileDetail
+          let actionsKey = TEXT_ACTIONS
+          if (rec.profile_type !== 'goroutine') {
```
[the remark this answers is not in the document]
baurine: I confirmed with @crazycs520 , it is different with manual profiling for mutex profile type. in manual profiling, it profiles with `debug=1` params and get txt result, it would affect the cluster performance, so in the conprof, we don't attach the `debug=1` params, then we get protobuf result.

ui/lib/apps/ContinuousProfiling/pages/Detail.tsx
```diff
@@ -26,26 +37,63 @@ export default function Page() {
...
+    const newGroups: IGroup[] = []
+
+    let startIndex = 0
+    const profiles = groupProfileDetail?.target_profiles || []
+    // rename profile to cpu for profile_type for easier sort
+    profiles.forEach((p) => {
+      if (p.profile_type === 'profile') {
+        p.profile_type = 'cpu'
+      }
+    })
```
shhdgit, line 51: Can we sort profile type by specific enum order.

ui/lib/apps/ContinuousProfiling/pages/Detail.tsx
```diff
@@ -58,53 +106,99 @@ export default function Page() {
...
-      )
+
+      if (action === 'view_graph' || action === 'view_text') {
+        const profileURL = `${client.getBasePath()}/continuous_profiling/single_profile/view?token=${token}`
+        window.open(profileURL, '_blank')
+        return
+      }
+
+      if (action === 'view_flamegraph') {
+        // view flamegraph by speedscope
```
shhdgit, line 181: Maybe we can have a helper to opens speedscope between manual profiling and conprof.

ui/lib/apps/ContinuousProfiling/pages/Detail.tsx
baurine: How can we do it, do you have an example?

ui/lib/apps/ContinuousProfiling/pages/Detail.tsx
baurine: Yep, we can refine it later.

ui/lib/apps/ContinuousProfiling/pages/Detail.tsx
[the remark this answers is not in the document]
YiniXu9506: Got it! 👍

ui/lib/apps/ContinuousProfiling/pages/Detail.tsx
shhdgit: Do we need to specify the ordering here? Maybe I misunderstood it before.

baurine: Yep, the design spec requires.

breeswish: Maybe keeping a specific order is fine. There is no need to make CPU the first one? Please confirm with PM.

Even if you want to sort CPU as the first, what you should do is defining a sort key array and sort by the sort key.

baurine: Confirmed with PM, it is required, let me refine the sort method.

baurine: done!

baurine:
all done, PTAL, thanks!

breeswish (approved):

PR #1099, opened 2021-12-13
Feature Top SQL

shhdgit:
- [x] Top SQL List & Detail
- [x] Remove `react-query` dep
- [x] Fix fluent-ui DetailList Selection bug

image

only one plan:
image

multi plans:
image
image

PR #1105, opened 2021-12-22
refactor(ui): auto refresh button

shhdgit:
Related PR: #1099


shhdgit:
/cc @baurine @breeswish 
PTAL, thanks!

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+    )
+    return () => clearTimeout(timer)
+  }, [autoRefreshSeconds, isLoading, getRemainingRefreshSeconds()])
+
+  // reset auto refresh when onRefresh function update
+  useEffect(() => {
+    clearTimeout(timer)
+    setTimer(undefined)
+    setRemainingRefreshSeconds(autoRefreshSeconds)
+  }, [onRefresh])
```
breeswish, line 123: I don't even need to take a close look and will know there will be problems for people ever want to write interval by himself :)

For example, if people are not wrapping onRefresh in a memo then this component will be broken due to onRefresh is changing all the time and there will be never a timer successfully set. This is not a good public API.

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+      setTimer(undefined)
+      return
+    }
+
+    clearTimeout(timer)
+    setTimer(
+      setTimeout(() => {
+        if (isLoading) {
+          return
+        }
```
breeswish, line 105: It may not be a good idea to let the external drive the component, since your component itself is driving the externals (it emits refresh ticks periodically). In this case, someone may just very easy to break it.

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+  // reset auto refresh when onRefresh function update
+  useEffect(() => {
+    clearTimeout(timer)
+    setTimer(undefined)
+    setRemainingRefreshSeconds(autoRefreshSeconds)
+  }, [onRefresh])
+
+  return (
+    <Space>
```
breeswish, line 126: This makes your "Button" not just a button. Please avoid doing this. Otherwise you will find its use case highly limited.

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+  const handleRefresh = useCallback(() => {
+    if (isLoading) {
+      return
+    }
+    onRefresh()
+  }, [isLoading, onRefresh])
+
+  // Auto refresh
+  const [timer, setTimer] = useState<number | undefined>()
```
breeswish, line 85: There is a specific type for the return value of `setTimeout` and maybe you can use it.

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+        })}
+      </Menu.ItemGroup>
+    </Menu>
+  )
+
+  const handleRefresh = useCallback(() => {
+    if (isLoading) {
+      return
+    }
+    onRefresh()
```
breeswish, line 81: onRefresh may be unspecified

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+  }, [isLoading, onRefresh])
+
+  // Auto refresh
+  const [timer, setTimer] = useState<number | undefined>()
+  const [getRemainingRefreshSeconds, setRemainingRefreshSeconds] =
+    useGetSet(autoRefreshSeconds)
+
+  useEffect(() => {
+    setRemainingRefreshSeconds(autoRefreshSeconds)
```
breeswish, line 90: This means when there are for example 5s remains, changing auto-refresh to 30s will reset it to 30s. Is it possible to keep the 5s update to finish ticking?

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+  }, [autoRefreshSeconds])
+
+  useEffect(() => {
+    if (autoRefreshSeconds === 0) {
+      clearTimeout(timer)
+      setTimer(undefined)
+      return
+    }
```
breeswish, line 98: What's the behavior of setting autoRefreshSeconds to 0? I think it needs to be clearly defined (or forbidden)

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+    clearTimeout(timer)
```
breeswish, line 100: the `timer` is not in the deps list, which means you may access a stale `timer`. Is it fine?

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+        if (getRemainingRefreshSeconds() === 0) {
+          setRemainingRefreshSeconds(autoRefreshSeconds)
+          handleRefresh()
+        } else {
+          setRemainingRefreshSeconds((c) => c - 1)
+        }
+      }, 1000) as unknown as number
+    )
+    return () => clearTimeout(timer)
+  }, [autoRefreshSeconds, isLoading, getRemainingRefreshSeconds()])
```
breeswish, line 116: The CI is passing without complaining about missing deps.. It doesn't look right. Could you take a look?  @baurine

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+  // reset auto refresh when onRefresh function update
```
breeswish, line 118: why do we need this behavior?

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+    clearTimeout(timer)
+    setTimer(undefined)
+    setRemainingRefreshSeconds(autoRefreshSeconds)
+  }, [onRefresh])
+
+  return (
+    <Space>
+      <Dropdown.Button
+        disabled={!enabled}
+        onClick={() => handleRefresh()}
```
breeswish, line 129:
```suggestion
        onClick={handleRefresh}
```

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+import { Spin, Dropdown, Menu, Space } from 'antd'
+import { useSpring, animated } from 'react-spring'
+import { getValueFormat } from '@baurine/grafana-value-formats'
+import { useTranslation } from 'react-i18next'
+import { addTranslationResource } from '@lib/utils/i18n'
+import { useGetSet } from 'react-use'
+
+interface AutoRefreshButtonProps {
+  // use seconds as options
+  options: number[]
```
breeswish, line 12:
```suggestion
  refreshSecs: number[]
```

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+  autoRefreshSeconds: number
+  onAutoRefreshSecondsChange: (v: number) => void
+  onRefresh: () => void
+  isLoading: boolean
```
breeswish, line 16: this is actually duplicated with enabled I think.

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
shhdgit: Let me redesign the interface.

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
baurine: OK, let me check it.

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+import { getValueFormat } from '@baurine/grafana-value-formats'
+import { useTranslation } from 'react-i18next'
+import { addTranslationResource } from '@lib/utils/i18n'
+import { useGetSet } from 'react-use'
+
+interface AutoRefreshButtonProps {
+  autoRefreshSecondsOptions: number[]
+  autoRefreshSeconds: number
+  onAutoRefreshSecondsChange: (v: number) => void
+  onRefresh: () => Promise<void>
```
breeswish, line 14: I think you actually don't need it to be async. `disabled` can be already used to express the information about whether the auto refresh ticking should continue. BTW your current implementation keep ticking when disabled = true, which is incorrect.

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+    if (isLoading) {
+      return
+    }
+    setIsLoading(true)
+    await onRefresh()
+    setIsLoading(false)
+  }, [isLoading, onRefresh])
+
+  // Auto refresh
+  const [getTimer, setTimer] = useGetSet<NodeJS.Timer | undefined>(undefined)
```
breeswish, line 91:
```suggestion
  const [getTimer, setTimer] = useGetSet<ReturnType<typeof setTimeout> | undefined>(undefined)
```

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const handleRefresh = useCallback(async () => {
+    if (isLoading) {
+      return
+    }
+    setIsLoading(true)
+    await onRefresh()
+    setIsLoading(false)
+  }, [isLoading, onRefresh])
+
+  // Auto refresh
```
breeswish, line 90: getTimer and setTimer is a state. Is it a correct behavior that when timer is changed or finished the component needs a re-render? If this behavior is not needed, then you may want a `useRef` instead.

breeswish (commented):
I think a correct behavior of your component should be:

1. When auto refresh is on, a tick starts.
2. At any time it is set to disabled, the tick paused.
3. At any time it is set to enabled, the tick continues.
4. When auto refresh settings is changed, the tick skips if the new refresh interval is set to a value smaller than remaining ticks.
5. When auto refresh is off, the tick stops.

The loading behavior can be naturally expressed with the `disabled` property, as you should not allow user to click refresh button again when there is a request on-going. In this case, you will want to set the `disabled` when using this component. This also applies to the auto refresh: there is no need for the tick to continue if there is a request on-going, since the last request is not finished yet. This means, `disabled` should be applied to both manual refresh and auto refresh. One property for all, which is great.

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,203 @@
...
+    timer.current = setTimeout(() => {
+      if (getRemainingRefreshSeconds() === 0) {
+        setRemainingRefreshSeconds(autoRefreshSeconds)
+        handleRefresh()
+      } else {
+        setRemainingRefreshSeconds((c) => c - 1)
+      }
+    }, 1000)
+    return () => clearTimeout(timer.current!)
+  }, [autoRefreshSeconds, disabled, getRemainingRefreshSeconds()])
```
baurine, line 131: Should depend on `getRemainingRefreshSeconds` instead of `getRemainingRefreshSeconds()`?

baurine (approved):

ui/lib/apps/KeyViz/components/KeyViz.tsx
```diff
@@ -99,6 +98,14 @@ const KeyViz = () => {
 
   const enabled = config?.auto_collection_disabled !== true
 
+  const resetAutoRefresh = useCallback(() => {
+    const prevAutoRefreshSeconds = getAutoRefreshSeconds()
+    setAutoRefreshSeconds(0)
+    setTimeout(() => {
+      setAutoRefreshSeconds(prevAutoRefreshSeconds)
+    })
+  }, [getAutoRefreshSeconds()])
```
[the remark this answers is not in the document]
shhdgit: There's a reset behavior in L121-L123. Any good ideas?
```
  const updateHeatmap = useCallback(async () => {
    if (getAutoRefreshSeconds() > 0) {
      setRemainingRefreshSeconds(getAutoRefreshSeconds())
    }
```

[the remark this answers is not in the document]
breeswish: When heatmeap is updated, it resets the **remaining** seconds to the refresh second options, instead of not changing anything at all.

ui/lib/apps/KeyViz/components/KeyViz.tsx
```diff
@@ -174,6 +182,14 @@ const KeyViz = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
 
+  const onChangeMetricType = useCallback(
+    (v: DataTag) => {
+      setMetricType(v)
+      resetAutoRefresh()
```
breeswish, line 188: I believe it is intentional that the there is no need to reset the refresh when metric type is changed, since there will be no data fetching operations during this.

ui/lib/apps/KeyViz/components/KeyViz.tsx
```diff
@@ -143,11 +147,15 @@ const KeyViz = () => {
...
   }, [])
 
-  const onChangeDateRange = useCallback((v: number) => {
-    setDateRange(v)
-    setSelection(null)
+  const onChangeDateRange = useCallback(
+    (v: number) => {
+      setDateRange(v)
+      setSelection(null)
+      resetAutoRefresh()
```
breeswish, line 154: I think you should revert these changes and only reset the timer in `updateHeatmap` or places like that, since it is the "request" behavior that causes a timer to be reset.

ui/lib/apps/KeyViz/components/KeyVizToolbar.tsx
```diff
@@ -105,7 +41,7 @@ class KeyVizToolbar extends Component<IKeyVizToolbarProps & WithTranslation> {
     this.props.onRefresh()
   }
 
-  handleAutoRefreshMenuClick = ({ key }) => {
+  handleAutoRefreshMenuClick = (key) => {
     this.props.onChangeAutoRefresh(parseInt(key))
```
breeswish, line 45: No need to parseInt?

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -0,0 +1,203 @@
...
+    if (!autoRefreshSeconds || disabled) {
+      if (!!timer.current) {
+        clearTimeout(timer.current)
+        timer.current = undefined
+      }
+      return
+    }
+
+    timer.current = setTimeout(() => {
+      if (getRemainingRefreshSeconds() === 0) {
```
breeswish, line 123: I guess you may like to have something like `onRemainingSecondsChange` to let the outside to control your remaining seconds. Otherwise it is not possible for the outside th reset the remaining seconds.

shhdgit: Very helpful! It's looks like the concept of `remaining seconds` is unavoidable.

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
baurine, line 131: I feel this kind of way to implement the timer is a little hacky, and not accurate. There is a gap between the effects, maybe use interval is better.

shhdgit: Yes, it's not accurate and it's on purpose. This may be necessary if we want to pause or resume instantly. Any good ideas?

ui/lib/apps/KeyViz/components/KeyViz.tsx
shhdgit: It looks better to recover the remaining seconds. And there's data fetching logic caused by metric type at L195-L200. Is this as expected? PTAL.

breeswish (approved):

PR #1106, opened 2021-12-22
fix(distro): check distro_strings.json fmt by prettier

shhdgit:
Since the distro_strings.json is generated by scripts when dev, there is no need for prettier to check and fmt.


shhdgit:
/cc @breeswish @baurine 
PTAL, thanks!

breeswish (approved):

PR #1107, opened 2021-12-24
fix generate assets

baurine:
What did:

- Fix `scripts/generate_assets.go` crash when building by `DASHBOARD=COMPILE make pd-server` in pd repo

### DIFF
--- a/scripts/embed_ui_assets.sh
+++ b/scripts/embed_ui_assets.sh
@@ -32,7 +32,7 @@ fi
 echo "+ Embed UI assets"
 
 cd "$PROJECT_DIR/scripts"
-go run generate_assets.go "$PROJECT_DIR/ui/build" $BUILD_TAG_PARAMETER
+go run generate_assets.go "$PROJECT_DIR/ui/build" "$BUILD_TAG_PARAMETER"
 
 HANDLER_PATH=$PROJECT_DIR/pkg/uiserver/embedded_assets_handler.go
 mv assets_vfsdata.go $HANDLER_PATH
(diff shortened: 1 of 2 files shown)


scripts/embed_ui_assets.sh
```diff
@@ -32,7 +32,7 @@ fi
 echo "+ Embed UI assets"
 
 cd "$PROJECT_DIR/scripts"
-go run generate_assets.go "$PROJECT_DIR/ui/build" $BUILD_TAG_PARAMETER
+go run generate_assets.go "$PROJECT_DIR/ui/build" $BUILD_TAG_PARAMETER
```
baurine, line 35: `$BUILD_TAG_PARAMETER` may be empty

scripts/generate_assets.go
```diff
@@ -16,7 +16,10 @@ func main() {
 		log.Fatal("Require 2 args")
 	}
 	directory := os.Args[1]
-	buildTag := os.Args[2]
+	buildTag := ""
+	if len(os.Args) > 2 {
+		buildTag = os.Args[2]
+	}
```
baurine, line 22: When `$BUILD_TAG_PARAMETER` is empty, `len(os.Args) == 2`, and `os.Args[2]` would crash.

breeswish: Could you try with `"$BUILD_TAG_PARAMETER"` instead of `$BUILD_TAG_PARAMETER`?

baurine: Yep, that's would be better, let me try it.

scripts/generate_assets.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 func main() {
-	if len(os.Args) < 2 {
+	if len(os.Args) < 3 {
```
baurine, line 15: `os.Args[0]` is the command bin self.

breeswish (approved):

PR #1109, opened 2021-12-27
chore: add e2e test

YiniXu9506:
What this PR did?

- Add E2E test with Cypress framework
- Add User login test
- Integrate GitHub Action
- Test TiDB Dashboard latest features and lower version compatibility.


.github/workflows/e2e-test-compatibility.yaml
```diff
@@ -0,0 +1,88 @@
...
+
+on:
+  pull_request:
+    branches:
+      - master
+      - release
+
+jobs:
+  e2e_test:
+    name: UI
```
shhdgit, line 11: `name: E2E Test`

ui/dashboardApp/layout/signin/index.tsx
```diff
@@ -215,6 +215,9 @@ const LAST_LOGIN_USERNAME_KEY = 'dashboard_last_login_username'
 
 function TiDBSignInForm({ successRoute, onClickAlternative }) {
   const supportNonRootLogin = useIsFeatureSupport('nonRootLogin')
+
+  console.log('supportNonRootLogin:', supportNonRootLogin)
```
shhdgit, line 219: `console.log`

ui/dashboardApp/layout/signin/index.tsx
```diff
@@ -459,6 +464,7 @@ function App({ registry }) {
     async function run() {
       try {
         const resp = await client.getInstance().userGetLoginInfo()
+        console.log('resp', resp)
```
shhdgit, line 467: ditto

.github/workflows/e2e-test-compatibility.yaml
```diff
@@ -0,0 +1,88 @@
...
+          key: ${{ runner.os }}-tiup
+          restore-keys: |
+            ${{ runner.os }}-tiup
+      - name: Install and run TiUP in the background
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://tiup-mirrors.pingcap.com/install.sh | sh
+          source /home/runner/.profile
+          tiup update playground
+          source /home/runner/.profile
+          tiup playground nightly --tiflash=0 &
```
shhdgit, line 52: Maybe we can share some common scripts.

https://github.com/pingcap/tidb-dashboard/pull/1083/files#diff-4502f11bfc8852994bb2dccf25ef4ee07afd342f40de01feecf9b452f3ba7113

YiniXu9506: Got it! I will take a look.

breeswish:
I have merged https://github.com/pingcap/tidb-dashboard/pull/1083. You can use common scripts now.

.github/workflows/test.yaml
```diff
@@ -137,11 +141,16 @@ jobs:
           make run &
         env:
           UI: 1
-      - name: Run e2e test
-        run: |
-          cd ui/tests
-          yarn
-          yarn test
+          TEST_COMPATIBILITY: "${{ matrix.feature_flag == '5.0.0' && '1' || '0' }}"
```
breeswish, line 144: How about using `FEATURE_VERSION: ${{ matrix.feature_version }}`? It seems to be more clear and depends less on the Makefile.

.github/workflows/test.yaml
```diff
@@ -137,11 +141,16 @@ jobs:
...
+      - name: Run e2e test features
```
breeswish, line 145: You can run test functions for all feature versions, it would be better!

.github/workflows/test.yaml
```diff
@@ -65,9 +65,13 @@ jobs:
...
           verbose: true
   e2e_test:
-    name: E2E
+    name: E2E Test
     runs-on: ubuntu-latest
     timeout-minutes: 10
+    strategy:
+      matrix:
+        # test latest features and compatibility of lower version
+        feature_flag: [6.0.0, 5.0.0]
```
breeswish, line 74: Maybe you will want this matrix, so that `tiup playground $TIDB_VERSION  --tiflash=0` can be covered and there is no magic check when running compatibility tests:

```yaml
strategy:
  matrix:
    include:
      - feature_version: 6.0.0
        tidb_version: nightly
        run_compat_test: no
      - feature_version: 5.0.0
        tidb_version: v5.0.0
        run_compat_test: yes
```

ui/package.json
```diff
@@ -30,6 +30,7 @@
     "i18next-browser-languagedetector": "^5.0.0",
     "lodash": "^4.17.21",
     "moize": "^5.4.7",
+    "mysql": "mysqljs/mysql",
```
breeswish, line 33: I believe you should use https://github.com/sidorares/node-mysql2 which is more up-to-date.

breeswish: Also this should be a dev dependency I guess, as it is test related.

ui/dashboardApp/layout/signin/index.tsx
```diff
@@ -264,12 +265,14 @@ function TiDBSignInForm({ successRoute, onClickAlternative }) {
             <h2>{t('signin.form.tidb_auth.title')}</h2>
           </Form.Item>
           <Form.Item
+            data-e2e="signin_username_form_item"
```
breeswish, line 268: Seems that this e2e is unused.

ui/cypress/integration/features/user_login_spec.js
```diff
@@ -0,0 +1,102 @@
...
+
+      // {enter} causes the form to submit
+      cy.get('[data-e2e="signin_password_input"]').type(
+        'incorrect_password{enter}'
+      )
+      cy.wait('@login').then(() => {
+        cy.get('[data-e2e="signin_password_form_item"]').should(
+          'have.class',
+          'ant-form-item-has-error'
+        )
```
breeswish, line 52: How about testing the error message? You can expect error messages are shown in the form.

ui/cypress/integration/features/user_login_spec.js
```diff
@@ -0,0 +1,102 @@
...
+      cy.task('queryDB', { query, password })
+    })
+
+    it('nonRoot user with correct password', function () {
+      cy.get('[data-e2e=signin_username_input]').clear().type('test')
+      cy.get('[data-e2e="signin_password_input"]').type('test_pwd{enter}')
+      cy.url().should('include', '/overview')
+    })
+
+    it('nonRoot user with incorrect password', function () {
```
breeswish, line 93: You can also remove existing e2e test cases that has been covered by your new test cases.

breeswish (commented):
Generally looks cool! @baurine @shhdgit PTAL

breeswish (approved):

breeswish:
@shhdgit PTAL whether it is possible to have same compatibility matrix. This can be improved in future PRs. Now integration test has latest + v4.0.1 and E2E has nightly + v5.0.0. It would be better to make them identical.

ui/cypress/integration/common-features/user_login_spec.js
```diff
@@ -0,0 +1,53 @@
...
+// Test User login
+describe('Root User Login', () => {
+  beforeEach(function () {
+    cy.fixture('uri.json').as('uri')
+    cy.visit('@uri.root')
+  })
+
+  it('authenticated redirect', function () {
+    // Redirect to login
+    cy.visit(this.uri.root)
```
shhdgit, line 12: Can these two uses be the same style? For example, both use `@uri.root` or `this.uri.root`. I don't know if this is possible.

ui/cypress/integration/compatibility-features/user_login_spec.js
```diff
@@ -0,0 +1,74 @@
...
+
+      // Direct to login page
+      cy.visit('@uri.root')
+    })
+
+    it('noRootLogin is supported', () => {
+      cy.log('FEATURE_VERSION is: ', Cypress.env('FEATURE_VERSION'))
+      cy.request(`${Cypress.env('apiUrl')}info/info`).then((resp) => {
+        expect(resp.status).to.eq(200)
+        expect(resp.body.supported_features).include('nonRootLogin')
```
shhdgit, line 34: Do we really need to test the api response? From the user's side, maybe BDD is enough. If the business process doesn't change, the test is valid. But the implementation may change.

shhdgit (approved):
The rest LGTM. Can't wait to write the e2e test, Looks interesting!

shhdgit:
> @shhdgit PTAL whether it is possible to have same compatibility matrix. This can be improved in future PRs. Now integration test has latest + v4.0.1 and E2E has nightly + v5.0.0. It would be better to make them identical.

Ok

ui/cypress/integration/common-features/user_login_spec.js
YiniXu9506: Both `@uri.root` and `this.uri.root` work fine, but you are right, we'd better use one style. I chose `this.uri.root`, since it will be easier to use in assertion.

ui/cypress/integration/compatibility-features/user_login_spec.js
YiniXu9506: I agree that from the user side, this check may not be necessary, but here we want to check whether the TiDB Dashboard with `FEATURE_VERSION >= 5.3.0` support `nonRootLogin` or not.  This is a real API test instead of mocking data to assert the supported_features.

PR #1112, opened 2021-12-30
refine: address feedback

YiniXu9506:
What this PR did?

Address feedback for manual profiling:

1. update download file/folder name, eg. `cpu_pd_127.0.0.1:2379_1234567.proto`
2. add `README.md` to download zip folder.
3. add download text source file.

### DIFF
--- a/pkg/apiserver/profiling/router.go
+++ b/pkg/apiserver/profiling/router.go
@@ -3,9 +3,13 @@
 package profiling
 
 import (
+	"archive/zip"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
+	"os"
+	"path/filepath"
 	"strconv"
 	"time"
 
@@ -17,7 +21,6 @@ import (
 	"github.com/pingcap/tidb-dashboard/pkg/apiserver/utils"
 	"github.com/pingcap/tidb-dashboard/pkg/config"
 	"github.com/pingcap/tidb-dashboard/util/rest"
-	"github.com/pingcap/tidb-dashboard/util/ziputil"
 )
 
 // Register register the handlers to the service.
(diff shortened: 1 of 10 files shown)


util/ziputil/zip_writer.go
```diff
@@ -26,6 +27,27 @@ func WriteZipFromFiles(w io.Writer, files []string, compress bool) error {
 		}
 	}
 
+	if withREADME {
```
breeswish, line 30: This place doesn't sounds to be right, as it is a "zip utility", and should have no logics about "cpu" or "profiling" or "readme".

ui/lib/apps/InstanceProfiling/translations/zh.yaml
```diff
@@ -52,7 +52,8 @@ instance_profiling:
             text: 查看文本
       status:
         finished: 完成
-        skipped: 不适用
+        skipped: 不支持
+        to_be_supported: 待支持
```
baurine, line 56: It seems the text should be unified to `不适用` by PM.

util/ziputil/zip_writer.go
```diff
@@ -7,12 +7,41 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"time"
 )
 
+type Options func(zw *zip.Writer) error
```
baurine, line 13: I just feel like `Options` is not a good type name for a function, and it should not be plural.

util/ziputil/zip_writer.go
```diff
@@ -7,12 +7,41 @@ import (
...
+			return err
+		}
+
+		_, err = zipFile.Write([]byte(downloadREADME))
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+}
```
breeswish, line 39: This is not utility at all. For example, you will never find a "zip" tool in your PC that contains a settings like [x] Add README..

breeswish (approved):

pkg/apiserver/profiling/router.go
```diff
@@ -260,13 +273,90 @@ func (s *Service) downloadSingle(c *gin.Context) {
...
 	c.Writer.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=\"%s\"", fileName))
-	err = ziputil.WriteZipFromFiles(c.Writer, []string{task.FilePath}, true)
+	zw := zip.NewWriter(c.Writer)
+	defer func() {
+		_ = zw.Close()
+	}()
+
+	err = writeZipFromFiles(zw, []string{task.FilePath}, true)
 	if err != nil {
 		log.Error("Stream zip pack failed", zap.Error(err))
```
breeswish, line 286: How about abort the stream and return?

pkg/apiserver/profiling/router.go
```diff
@@ -260,13 +273,90 @@ func (s *Service) downloadSingle(c *gin.Context) {
...
 	}
+
+	err = zipREADME(zw)
+	if err != nil {
+		log.Error("Stream zip README file failed", zap.Error(err))
```
breeswish, line 291: How about abort the stream and return?

pkg/apiserver/profiling/router.go
```diff
@@ -260,13 +273,92 @@ func (s *Service) downloadSingle(c *gin.Context) {
...
+		return err
+	}
+
+	_, err = zipFile.Write([]byte(downloadREADME))
 	if err != nil {
-		log.Error("Stream zip pack failed", zap.Error(err))
+		return err
 	}
+	return nil
 }
```
baurine, line 362: How about passing the content and file name from outside, so that become a utility method. We can continue to refine it later.

YiniXu9506: Yes, we can refine this later.

PR #1115, opened 2021-12-30
update ui mod name

baurine:
Update ui mod name to make PD can compile with tidb-dashboard.

### DIFF
--- a/ui/go.mod
+++ b/ui/go.mod
@@ -1,3 +1,3 @@
-module ignore_ui // a hack to ignore this directory in go commands
+module github.com/pingcap/tidb-dashboard/ui // a hack to ignore this directory in go commands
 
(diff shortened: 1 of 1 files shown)


ui/go.mod
```diff
@@ -1,3 +1,4 @@
-module ignore_ui // a hack to ignore this directory in go commands
+// module ignore_ui // a hack to ignore this directory in go commands
+module github.com/pingcap/tidb-dashboard/ui // a hack to ignore this directory in go commands
```
breeswish, line 2:
```suggestion
module github.com/pingcap/tidb-dashboard/ui

// a hack to ignore this directory in go commands
// note: this is still an path that can be imported, as the source may be need in
// some other places, like PD.
```

PR #1117, opened 2021-12-30
update scripts mod name

baurine:
update scripts and ui mod name for compiling with PD.


ui/ui.go
```diff
@@ -0,0 +1 @@
+package ui
```
breeswish, line 1: Please add comments about why do we need this

baurine: sure

baurine:
Close this PR because we fix the PD compile issue by the better solution in PR https://github.com/tikv/pd/pull/4502

PR #1118, opened 2021-12-30
fix: ignore override tidb endpoint when profiling

shhdgit:
We need to ignore `TIDB_OVERRIDE_STATUS_ENDPOINT` env var when profiling so that we can profile the correct target even if `TIDB_OVERRIDE_STATUS_ENDPOINT` is set.


breeswish (approved):
Please add it to the TODO test case list to cover.

PR #1122, opened 2021-12-31
refine(topsql): ux

shhdgit:
Some ux improvements.


shhdgit:
/cc @baurine @breeswish 
PTAL, thanks!

breeswish:
I made reviews for the last PR first in https://github.com/pingcap/tidb-dashboard/pull/1099#pullrequestreview-842162652. Please take a look to see if they have already been handled in this PR. If not, maybe you need to address them.

shhdgit:
How about solving the ux problem in this pr and opening another pr for other issues?

ui/lib/apps/TopSQL/utils/specialRecord.ts
```diff
@@ -0,0 +1,55 @@
...
+import type { PlanRecord } from '../pages/List/ListDetail/ListDetailTable'
+import type { SQLRecord } from '../pages/List/ListTable'
+
+const OVERALL_IDENTIFIER = '__OVERALL_IDENTIFIER__'
+
+export const isOverallRecord = (r: PlanRecord) => {
+  return r.plan_digest === OVERALL_IDENTIFIER
+}
+
+export const createOverallRecord = (records: PlanRecord[]) => {
```
breeswish, line 12: I think this should be handled by ngm @zhongzc 
The front-end should not have any logic about the computation.

ui/lib/components/AutoRefreshButton/index.module.less
```diff
@@ -0,0 +1,12 @@
+.auto_refresh_btn {
+  :global {
+    .ant-dropdown-trigger {
+      width: auto;
+      padding: 0 10px;
+    }
+  }
+
+  .auto_refresh_secs {
+    font-size: 14px;
```
breeswish, line 10: Please use rem.

BTW I have no idea why this need to be specified explicitly. Maybe it should be something like button text size?

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
```diff
@@ -133,11 +135,23 @@ export function AutoRefreshButton({
...
     <Dropdown.Button
+      className={styles.auto_refresh_btn}
       disabled={disabled}
       onClick={handleRefresh}
       overlay={autoRefreshMenu}
       trigger={['click']}
-      icon={<DownOutlined />}
+      icon={
+        <>
+          {autoRefreshSeconds ? (
```
breeswish, line 145: Maybe you should move the circle ring to here.

ui/lib/apps/TopSQL/utils/specialRecord.ts
shhdgit: Oh, yes, there is wrong computation here.

zhongzc: > I think this should be handled by ngm

https://github.com/pingcap/ng-monitoring/pull/93

shhdgit:
Use table rows as legends.


https://user-images.githubusercontent.com/11549583/148167216-f200a4de-22d8-43f3-bef1-1c7699bbdbdb.mov

ui/lib/components/AutoRefreshButton/AutoRefreshButton.tsx
shhdgit: But I feel like it looks more balanced now and closer to grafana.

image
image
image

PR #1123, opened 2022-01-04
fix: add a slash (/) before the hashtag for cross browser compatibility.

YiniXu9506:
What this PR did?

1. Remove extra rows for instances with empty tasks.
image

2. There is a cross browser compatibility about hashtag. The profile URL with hashtag works unstable on safari. Sometimes different profile URLs load the same speedscope page, which is the previous loaded speedscope page.  The URL with tashtag has following behaviors on different browsers:

    `www.example.com#item`: this hashtag works fine on firefox and chrome, but work unstable on safari.
    `www.example.com/#item`: url with a back slash before hashtag works fine on firefox, chrome and safari.

      This issue can also be referred [here](https://stackoverflow.com/questions/18663410/safari-ignoring-removing-anchors-or-hashtags-when-clicking-hyperlinks/24778997#24778997)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -60,4 +60,4 @@ linters-settings:
       - "^\\s*@"
   goheader:
     template: |-
-      Copyright {{ YEAR }} PingCAP, Inc. Licensed under Apache-2.0.
+      Copyright {{ YEAR-RANGE }} PingCAP, Inc. Licensed under Apache-2.0.
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -60,4 +60,4 @@ linters-settings:
       - "^\\s*@"
   goheader:
     template: |-
-      Copyright {{ YEAR-RANGE }} PingCAP, Inc. Licensed under Apache-2.0.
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -60,4 +60,4 @@ linters-settings:
       - "^\\s*@"
   goheader:
     template: |-
-      Copyright {{ YEAR }} PingCAP, Inc. Licensed under Apache-2.0.
+      Copyright {{ YEAR-RANGE }} PingCAP, Inc. Licensed under Apache-2.0.
```
breeswish, line 63: We can bump year in a new PR.

```suggestion
      Copyright {{ YEAR }} PingCAP, Inc. Licensed under Apache-2.0.
```

breeswish (approved):

.golangci.yml
YiniXu9506: > We can bump year in a new PR.

Should we bump all years in the goheader from 2021 to 2022?

PR #1128, opened 2022-01-10
Update some NGM & Top SQL text

shhdgit:


ui/lib/apps/TopSQL/translations/en.yaml
```diff
@@ -1,6 +1,6 @@
 topsql:
   nav_title: Top SQL
-  ngm_not_started: A required component `NgMonitoring` is not started in this cluster. Top SQL can not work.
+  ngm_not_started: A required component `NgMonitoring` is not started in this cluster. This feature is not available.
```
breeswish, line 3: I think you also need to unify the text notification of Conprof.

breeswish (commented):
the rest LGTM

PR #1131, opened 2022-01-10
chore(stmt): use common time range picker

shhdgit:
According to user feedback, common time range picker may be easier to use. Any better ideas?

/cc @breeswish @baurine @lilyjazz

From:
image

To:
image

### DIFF
--- a/ui/lib/apps/Statement/pages/List/index.tsx
+++ b/ui/lib/apps/Statement/pages/List/index.tsx
@@ -22,12 +22,17 @@ import { ScrollablePane } from 'office-ui-fabric-react/lib/ScrollablePane'
 import { useTranslation } from 'react-i18next'
 
 import { CacheContext } from '@lib/utils/useCache'
-import { Card, ColumnsSelector, Toolbar, MultiSelect } from '@lib/components'
+import {
+  Card,
+  ColumnsSelector,
+  Toolbar,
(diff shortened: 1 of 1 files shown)


baurine:
Or maybe we can keep the slider component, but not fix the value after the user selects the range?

Currently, people say it is not convenient, because when the user selects a time range, the selected time range may not be a suitable range, for example, 18:20 ~ 18:50, then we will auto fix it to 18:30 ~ 19:00, I have forgotten why we do this before, maybe we can remove this logic and keep the original selected range.

breeswish:
The reason we need to fix 18:20 ~ 18:50 into 18:30 ~ 19:00 is that TiDB generates report at this interval. If you specify a condition like ts_begin >= 18:20 and ts_end <= 18:50 then no record will be returned.

baurine:
> The reason we need to fix 18:20 ~ 18:50 into 18:30 ~ 19:00 is that TiDB generates report at this interval. If you specify a condition like ts_begin >= 18:20 and ts_end <= 18:50 then no record will be returned as the record will have ts_begin=18:30, ts_end=19:00.

Oh, I remember it. So maybe we can do this work in the background or in the backend?

breeswish:
We have discussed about how a time selection picker UI should be designed. Close for a new PR.

PR #1132, opened 2022-01-10
make more convinient for developing in the remote machine

baurine:
When a user develops TiDB dashboard in a remote machine and accesses the frontend page in the local, he will get the following error by accessing `http://{remote_addr}:3001/dashboard` in the local browser: 

image

Because the frontend can't access the backend API by `http://127.0.0.0:12333/dashboard/api`.

Solutions:

1. Forward local port 12333 to the remote, for example: `ssh -L 12333:locahost:12333 remote_addr`. 

1. Run TiDB dashboard backend by `bin/tidb-dashboard --host 0.0.0.0` and set `REACT_APP_DASHBOARD_API_URL=http://{remote_addr}:12333` in the `.env` file

1. Run TiDB dashboard backend by `bin/tidb-dashboard --host 0.0.0.0` and use `http://${window.location.hostname}:12333/dashboard` instead of `http://127.0.0.01:12333/dashboard` as API addr prefix.

This PR implements solution 3, and I think this would bring a little more convenience for the developers to develop in a remote machine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ endif
 
 .PHONY: run
 run:
-	bin/tidb-dashboard --debug --experimental --feature-version "$(FEATURE_VERSION)"
+	bin/tidb-dashboard --debug --experimental --feature-version "$(FEATURE_VERSION)" --host 0.0.0.0
 
 test_e2e_compat_features:
 	cd ui &&\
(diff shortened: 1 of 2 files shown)


breeswish (approved):

breeswish:
Nice trick!

PR #1136, opened 2022-01-11
misc: upgrade swag

breeswish:


baurine:
I did the test, upgrade swag to 1.7.6 will cause swag can't parse the type from dependent lib without `--parseDependency --parseInternal --parseDepth 1` options, but with these options, it will increase the generate time from current `1~2` seconds to `7~8` seconds. Although this can be resolved by `swaggertype` tag, but it also causes some other issues, I haven't dived into it yet.

related issues: 
- https://github.com/swaggo/swag/issues/817
- https://github.com/swaggo/swag/issues/808

breeswish:
> I did the test, upgrade swag to 1.7.6 will cause swag can't parse the type from dependent lib without `--parseDependency --parseInternal --parseDepth 1` options, but with these options, it will increase the generate time from current `1~2` seconds to `7~8` seconds. Although this can be resolved by `swaggertype` tag, but it also causes some other issues, I haven't dived into it yet.
> 
> related issues:
> 
> * [Version 1.6.9 - cannot find type definition swaggo/swag#817](https://github.com/swaggo/swag/issues/817)
> * [Commit 67cb0975876f155725732d1aeb87fb8310e6aaf9 breaks parsing of external types swaggo/swag#808](https://github.com/swaggo/swag/issues/808)

Yes, I meet a bug that is only fixed in later swag versions.

I guess the `swaggertype` tag is not a correct way to resolve problems you met, as I discovered that it is a per-field configuration instead of per-type configuration, which means you may need to add this annotation to everywhere you use the type and there is even no way to ensure that every place is covered. It is basically a bad designed feature that may be better not use at all.

PR #1148, opened 2022-01-20
profiling: Migrate to the new architecture

breezewish:
1. Separate profiling module into:
- profutil (this can be useful to be used in ngm to fetch profilings)
- view
- model (tidb dashboard model implementation)

2. Update frontend accordingly. A new `<InstanceSelectV2>` is introduced, which will use the instance list provided by the profiling backend API and pass back the enveloped instance selection. We should switch to use InstanceSelectV2 one by one.

3. Update the Conprof and Manual Profiling UI from a dropdown list to a link list in order to avoid overflowing the table row:

image


breezewish:
The UI change has been put in a standalone PR https://github.com/pingcap/tidb-dashboard/pull/1148

PR #1149, opened 2022-01-20
misc: upgrade to swag v1.7.6

breeswish:
This is extracted from https://github.com/pingcap/tidb-dashboard/pull/1148

Update [swag](https://github.com/swaggo/swag) to v1.7.6.

Note: the latest swag is v1.7.8. However this version drops the support of go1.13 so that we cannot use it for now.

To view the diff of generated swagger spec `diff swagger_old.yaml swagger_new.yaml`: https://www.diffchecker.com/Kv9FYbhx

The latest swag fixed some bugs when dealing with type aliases. A typical example would be:

```go
type (
	TaskProfilingType     string
	TaskProfilingTypeList []TaskProfilingType
)

type StartRequest struct {
	Targets                []model.RequestTargetNode `json:"targets"`
	DurationSecs           uint                      `json:"duration_secs"`
	RequstedProfilingTypes TaskProfilingTypeList     `json:"requsted_profiling_types"`
}
```

Old: The RequstedProfilingTypes becomes `[]<unresolved_symbol>`

image

image

New:  `[]string`

image


baurine (approved):
tested in esbuild branch as well, it works fine, cool~

PR #1150, opened 2022-01-20
fix: missing fields in slow query

shhdgit:
This is a pr to fix the bug of missing fields in slow query. After this, when it comes time to refactor slowquery, we can replace this part of the logic with the `VirtualView`.

Related:
- The previous virtual view PR https://github.com/pingcap/tidb-dashboard/pull/1028
- AskTUG: https://asktug.com/t/topic/183588


breeswish:
Please add tests to verify that your fix works (and the previous version without the fix doesn't work)

shhdgit:
Yeah, sure.

shhdgit:
/cc @breeswish @YiniXu9506 
PTAL, thanks~

ui/cypress/integration/slow_query/list.compat_spec.js
```diff
@@ -0,0 +1,99 @@
...
+          'digest',
+          'instance',
+          'db',
+          'connection_id',
+          'timestamp',
+
+          'query_time',
+          'parse_time',
+          'compile_time',
+          // 'rewrite_time',
```
YiniXu9506, line 36: How about remove the commented fields if we don't check them.

YiniXu9506 (approved):
LGTM

ui/cypress/integration/slow_query/list.compat_spec.js
shhdgit: I don't know if this will help in the review. In this way we can easily know which fields are to be displayed or not

shhdgit: Maybe I should extract these as a test case.

YiniXu9506: Got you! But actually cypress helps us to check these fields, I think we don't need to check by ourselves again. Maybe we can add the comparison of length of availableFields and the length of fields shows on popover content list.

```
cy.get(`[data-e2e^="columns_selector_field_"]`).should('have.length', availableFields.length)
```

ui/cypress/integration/slow_query/list.compat_spec.js
```diff
@@ -0,0 +1,99 @@
...
+    cy.url().should('include', this.uri.slow_query)
+  })
+
+  describe('Available fields', () => {
+    skipOn(Cypress.env('FEATURE_VERSION') !== '6.0.0', () => {
+      it('Show all available fields', () => {
+        cy.intercept('/dashboard/api/slow_query/available_fields').as(
+          'getAvailableFields'
+        )
+        cy.wait('@getAvailableFields')
```
YiniXu9506, line 23: I think we don't need to intercept the request manually since when the request will be sent when the slow query page loaded.

ui/cypress/integration/slow_query/list.compat_spec.js
shhdgit: There might be a problem if the length of availableFields and the length of fields is the same but the fields are wrong.

YiniXu9506: I think check both the existence of columns_selector_field_xx and length of column selector fields can be more safe.

ui/cypress/integration/slow_query/list.compat_spec.js
```diff
@@ -0,0 +1,99 @@
...
+          'txn_start_ts',
+          'success',
+          'is_internal',
+          'index_names',
+          'stats',
+          'backoff_types',
+
+          // 'wait_ts',
+          // 'cop_time',
+          // 'lock_keys_time',
```
breeswish, line 52: Looks like we should instead have them in the list!

shhdgit:
Some fields that were not added will be tracked continuously in https://github.com/pingcap/tidb-dashboard/issues/1185.

shhdgit:
I will merge this pr first, and the information related to slow query and statements fields will be tracked in https://github.com/pingcap/tidb-dashboard/issues/1185.

PR #1154, opened 2022-01-24
test: add upload e2e snapshots workflow

shhdgit:


breeswish (approved):

.github/workflows/upload-e2e-snapshots.yaml
```diff
@@ -0,0 +1,110 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    strategy:
+      matrix:
+        # test latest features and compatibility of lower version
+        include:
+          - feature_version: 6.0.0
+            tidb_version: nightly
+          - feature_version: 5.0.0
+            tidb_version: v5.0.0
```
breeswish, line 23: Are you expecting that the E2E artifacts are different in different version matrixes?

shhdgit: No. But there's different test cases in different version matrixes. May have different use for snapshot.

PR #1162, opened 2022-02-07
Fix broken styles caused by antd update

shhdgit:
Prev menu ✅:
image

Current menu ❌:
image

Prev debug api collapse ✅:
image

Current debug api collapse ❌:
image

### DIFF
--- a/ui/lib/apps/DebugAPI/apilist/ApiList.module.less
+++ b/ui/lib/apps/DebugAPI/apilist/ApiList.module.less
@@ -1,5 +1,19 @@
-.collapse_panel:not(:last-child) {
-  border-bottom: 1px solid #eee;
+.collapse_panel {
+  &:not(:last-child) {
+    border-bottom: 1px solid #eee;
+  }
+
+  :global {
+    .ant-collapse-header {
+      padding-left: 40px !important;
+    }
+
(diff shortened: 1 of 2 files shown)


shhdgit:
/cc @baurine @breeswish 
PTAL, thanks!

ui/lib/apps/DebugAPI/apilist/ApiList.module.less
```diff
@@ -1,5 +1,19 @@
-.collapse_panel:not(:last-child) {
-  border-bottom: 1px solid #eee;
+.collapse_panel {
```
breeswish, line 1: It looks pretty dangerous for us to rely on such styles. Our styles may be broken in future again when antd is upgraded.

breeswish (approved):

ui/lib/apps/DebugAPI/apilist/ApiList.module.less
shhdgit: Yes, we have a lot of hacking of antd style in our project.

breeswish: Snapshot tests seems to be really helpful in these cases!

PR #1178, opened 2022-02-10
fix ui doesn't work when running backend without 'feature-version' option

### LINKED ISSUE #1177: TiDB Dashboard cannot start with default cli args
## Bug Report

**What did you do?**

```
$ bin/tidb-dashboard
```

TiDB Dashboard UI works

**What did you see instead?**

UI cannot work, always showing "Error: Invalid Semantic Version".

master

baurine:
close #1177 

Try to dismiss the `FEATURE_VERSION` conception, use PDVersion directly to reduce the understanding confusion.


pkg/apiserver/info/info.go
```diff
@@ -110,6 +94,37 @@ func (s *Service) infoHandler(c *gin.Context) {
 	c.JSON(http.StatusOK, resp)
 }
 
+func (s *Service) checkNgmState() (utils.NgmState, error) {
+	ngmState := utils.NgmStateNotSupported
+
+	// The FeatureVersion value is "N/A" when running `make && bin/tidb-dashboard`
+	// The FeatureVersion value is "Unknown" when running `go run cmd/tidb-dashboard/main.go`
+	featureSupported :=
```
breeswish, line 102: It's interesting, why do we have these two different values? Is it possible to be simply `""`, which is the well-known default value for a missing string?

pkg/apiserver/info/info.go
```diff
@@ -110,6 +94,37 @@ func (s *Service) infoHandler(c *gin.Context) {
...
+		s.params.Config.FeatureVersion == "N/A" ||
+			s.params.Config.FeatureVersion == "Unknown"
+	if !featureSupported {
+		// drop "-alpha-xxx" suffix
+		versionWithoutSuffix := strings.Split(s.params.Config.FeatureVersion, "-")[0]
+		v, err := semver.NewVersion(versionWithoutSuffix)
+		if err != nil {
+			return ngmState, err
+		}
+		constraint, _ := semver.NewConstraint(">= v5.4.0")
```
breeswish, line 112: Why do we need to hard-code a 5.4.0 constraint here?

pkg/apiserver/info/info.go
baurine: The config.FeatureVersion is assigned to version.PDVersion default.

```
func Default() *Config {
	return &Config{
		DataDir:            "/tmp/dashboard-data",
		TempDir:            "",
		PDEndPoint:         "http://127.0.0.1:2379",
		PublicPathPrefix:   defaultPublicPathPrefix,
		ClusterTLSConfig:   nil,
		TiDBTLSConfig:      nil,
		EnableTelemetry:    true,
		EnableExperimental: false,
		FeatureVersion:     version.PDVersion,
	}
}
```

and the version.PDVeresion default value is "Unknown".

```
// Version information. It will be overwritten by LDFLAGS.
var (
	InternalVersion = "Unknown"
	Standalone      = "Unknown" // Unknown, Yes or No
	PDVersion       = "Unknown"
	BuildTime       = "Unknown"
	BuildGitHash    = "Unknown"
)
```

Its value is overwritten to "N/A" by LDFLAGS when building in dev mode.

```
LDFLAGS += -X "$(DASHBOARD_PKG)/pkg/utils/version.PDVersion=N/A"
```

Let me try to assign to `""` if PDVersion is "N/A" and "Unknown".

pkg/apiserver/info/info.go
baurine: @shhdgit why we don't register it to the feature flag?

shhdgit: Yes, we can use the feature flag to declare whether ngm is supported or not.

pkg/apiserver/info/info.go
shhdgit: Can we unify the PDVersion and the FEATURE_VERSION? Maybe we can specify the pd version during development.

breeswish: How about not assigning `version.PDVersion` to `Default()`? Will it cause any problems?

baurine: > How about not assigning `version.PDVersion` to `Default()`? Will it cause any problems?

Nope, my initial implementation is just to assign the FeatureVersion `""` default value, but later we need to use `FeatureVersion` and `version.PDVersion` both, to decide which version is preferable to use.

baurine: > How about not assigning `version.PDVersion` to `Default()`? Will it cause any problems?

I think it should be the same effect. we eventually need to use the `version.PDVersion`, before or later.

cmd/tidb-dashboard/main.go
```diff
@@ -53,6 +54,10 @@ type DashboardCLIConfig struct {
 
 // NewCLIConfig generates the configuration of the dashboard in standalone mode.
 func NewCLIConfig() *DashboardCLIConfig {
+	if _, err := semver.NewVersion(version.PDVersion); err != nil {
+		log.Fatal("Invalid sematic PD Version", zap.Error(err), zap.String("pd_version", version.PDVersion))
+	}
```
baurine, line 59: Check the PDVersion when starting, expose the error as early as possible.

PR #1187, opened 2022-02-16
feat: init optimizer trace

shhdgit:
- visit optimizer trace tool page via the specified url: `/optimizer_trace`

image


ui/lib/apps/OptimizerTrace/components/LogicalOperatorTree.tsx
```diff
@@ -0,0 +1,86 @@
+import React, { useEffect, useRef, useMemo } from 'react'
+import Viz from 'viz.js'
```
breeswish, line 2: 1. viz.js [is not maintained](https://github.com/mdaines/viz.js/). I doubt whether it will cause problems for us.

2. How large will the frontend compiled code become when we are now carrying viz.js in? Will it be too slow to load in cloud environments>

shhdgit: 1. Yes! viz.js is probably not a good choice. I've replaced it with d3-graphviz.

2. Optimizer trace has been split into a separate module. So it shouldn't affect first screen load.

3. d3-graphviz use wasm to compute the graphviz layout in web-worker so that it won't freeze the browser.

shhdgit:
/cc @breeswish @baurine 
PTAL, thanks~

ui/lib/apps/OptimizerTrace/components/LogicalOperatorTree.tsx
breeswish: Cool, d3-graphviz looks promising!

ui/lib/apps/OptimizerTrace/index.meta.ts
```diff
@@ -0,0 +1,8 @@
+import translations from './translations'
+
+export default {
+  id: 'optimizer_trace',
+  routerPrefix: '/optimizer_trace',
+  translations,
+  reactRoot: () => import(/* webpackChunkName: "app_optimizer_trace" */ '.'),
```
baurine, line 7: We don't need the `webpackChunkName` comment now.

baurine (approved):

ui/lib/apps/OptimizerTrace/index.meta.ts
shhdgit: Gotcha!

ui/lib/apps/OptimizerTrace/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+              accept=".json"
+              maxCount={1}
+            >
+              <Button icon={<UploadOutlined />}>Select File</Button>
+            </Upload>
+          </Space>
+        </Toolbar>
+      </Card>
+
+      {data && (
```
YiniXu9506, line 80: It will be better check the data validation, if the json fields are not those defined in interface, you may need to give a warning to users.

YiniXu9506 (approved):
Feedback about user experience has been delivered. The code LGTM.

PR #1193, opened 2022-03-01
feat: bring back mixpanel for telemetry

shhdgit:
- bring back mixpanel


ui/.env.development
```diff
@@ -5,3 +5,4 @@ REACT_APP_SENTRY_DSN=https://10930d3b0a8d427cad2147d6d845be56@o226447.ingest.sen
 REACT_APP_SENTRY_ENABLED=false
 # REACT_APP_RELEASE_VERSION is set in config-overrides.js
 REACT_APP_RELEASE_VERSION=unknown
+REACT_APP_MIXPANEL_TOKEN=
```
breeswish, line 8: @baurine Could you provide the token for this?

ui/.env.production
```diff
@@ -5,3 +5,4 @@ REACT_APP_SENTRY_DSN=https://10930d3b0a8d427cad2147d6d845be56@o226447.ingest.sen
 REACT_APP_SENTRY_ENABLED=true
 # REACT_APP_RELEASE_VERSION is set in config-overrides.js
 REACT_APP_RELEASE_VERSION=unknown
+REACT_APP_MIXPANEL_TOKEN=
```
breeswish, line 8: @baurine Could you provide the token for this?

shhdgit: It looks like we need to provide the token from PD's environment to prevent the token leaks. There are some unpredictable telemetry in the previous mixpanel project.

ui/.env.development
baurine: We can get it in the history commit.

ui/.env.production
shhdgit: The previous prod token was leaked https://github.com/pingcap/tidb-dashboard/pull/895/files#diff-f7e6bf7e30ccb470c218b3ce574c31e07e802d50c78d4a9802fa7c7ce2fcd14eL3

I found PD was embedded the dashboard static assets from tidb-dashboard's release, it seems we can add env variables to tidb-dashboard repo's workflow instead of pd.

breeswish: We need to make sure that the telemetry works.

ui/lib/utils/telemetry.ts
```diff
@@ -0,0 +1,37 @@
...
+    batch_requests: true,
+    persistence: 'localStorage',
+    property_blacklist: [
+      '$initial_referrer',
+      '$initial_referring_domain',
+      '$referrer',
+      '$referring_domain',
+    ],
+  }
+  const apiHost = process.env.REACT_APP_MIXPANEL_HOST
```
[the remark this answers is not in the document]
shhdgit: If the host API haven't been set, there will be a default API address like `https://api.mixpanel.com`

baurine:
`https://telemetry.pingcap.com/api/v1/dashboard/report` is not used to prevent being blocked by Ad Blockers, but is a proxy to store and analyze the logs by ourselves, not sure whether the service still works.

shhdgit:
> `https://telemetry.pingcap.com/api/v1/dashboard/report` is not used to prevent being blocked by Ad Blockers, but is a proxy to store and analyze the logs by ourselves, not sure whether the service still works.

Cool. I tested it and it still works.

ui/lib/utils/telemetry.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+  mixpanel.init(token, options)
+  // disable mixpanel to report data immediately
+  mixpanel.opt_out_tracking()
+  if (info?.enable_telemetry) {
+    mixpanel.register({
+      $current_url: getPathInLocationHash(),
+      dashboard_version: info.version?.internal_version,
+    })
+    mixpanel.opt_in_tracking()
```
breeswish, line 36: Remember to test the case that user can dynamically enable or disable the mixpanel by updating the PD configuration. It means when the page refreshes (while local storage is unchanged) the `enable_telemetry` may change from `true` to `false`. Please make sure that this change will work (both at first start up and when the web page is already opened once before). You may like some tests to ensure this behavior.

ui/lib/apps/TopSQL/utils/telemetry.ts
```diff
@@ -0,0 +1,48 @@
...
+import { TimeRange } from '@lib/components'
+import { TopsqlEditableConfig } from '@lib/client'
+
+export const telemetry = {
+  openSelectInstance() {
+    mixpanel.track('TopSQL: Open Select Instance')
+  },
+  finishSelectInstance(type: string) {
+    mixpanel.track('TopSQL: Finish Select Instance', { type })
+  },
```
breeswish, line 12: Will this leak user data?

ui/lib/apps/TopSQL/utils/telemetry.ts
```diff
@@ -0,0 +1,48 @@
...
+  clickSettings(type: 'firstTimeTips' | 'settingIcon' | 'bannerTips') {
+    mixpanel.track('TopSQL: Click Settings', { type })
+  },
+  saveSettings(settings: TopsqlEditableConfig) {
+    mixpanel.track('TopSQL: Save Settings', { settings })
+  },
+  clickStatement(index: number) {
+    mixpanel.track('TopSQL: Click Statement', {
+      rank: index === 5 ? 'other' : index + 1,
```
breeswish, line 42: This magic number is horrible.

ui/lib/apps/TopSQL/utils/telemetry.ts
shhdgit: No, we only collect instance types instead of other instance information.

ui/lib/apps/TopSQL/utils/telemetry.ts
shhdgit: Nice catch! let me refine it.

PR #1197, opened 2022-03-15
fix: conProf Profile URL

YiniXu9506:
The issue is:

Two different profiling results of one component have same Flame Graph on Speedscope, this caused by the same issue, which had been fixed [here](https://github.com/pingcap/tidb-dashboard/pull/1123).


breeswish (approved):

breeswish:
@shhdgit Please organize a release. Thanks!

shhdgit (approved):

PR #1179, opened 2022-02-10
Cleanup webpack

baurine:
close #1171 

Remove the unused files and configs caused by replacing webpack by esbuild.

PR #1063, opened 2021-11-16
Build the frontend by esbuild instead of webpack

baurine:
## What did:

- Build the whole frontend by esbuild instead of webpack, and keep the functions and style exact same as before.

### Details:

- Upgrade antd from 4.8 to 4.16 to eliminate tons of compilation errors
- Upgarde dayjs to dismiss the type check errors
- Hack the esbuild-plugin-postcss2
  - Fix the bug that can't find `.scss` files when compiling fluent-ui
  - Fix the bug that can't find `bulma/dist/bulma.css`
  - Add the cache to reduce incremental build time from more than 4s to near 600ms
  - Log the process time
- Fix broken styles
- Fix broken BaseSelect component
- Fix runtime crash: sentry, formatSQL, path, process.cwd
- Support diagnoseReport proxy
- Remove webpack-only methods to handle i18n
- ...

## Result:

Speedup the compilation time nearly by 9x. In a high-performance desktop machine (CPU: Intel Core i5-9400F @ 6x 4.1GHz, Memory: 64GB), the full build costs near 70s by webpack while only near 8s by esbuild, and only near 5s to build js and css.

In the dev mode, after changing js or CSS code, esbuild needs near 600ms to make an incremental build.

|                 |   production | dev incremental |
|----------|-------------|------------|
| webpack | 70.68s         | fast by hrm |
| esbuild    | 7.74s            | 600ms |

Full build time with webpack (1.13 min, aka **67.8s** for `webpack:build` stage):

```
[09:57:43] Finished 'swagger:generate_client' after 1.51 s
[09:57:43] Finished 'swagger:generate' after 1.82 s
[09:57:43] Starting 'webpack:build'...
[09:58:51] Finished 'webpack:build' after 1.13 min
[09:58:51] Finished 'build' after 1.15 min
Done in 70.68s.
```

Full build time with esbuild (**5.26s** for `esbuild:build` stage):

```
[22:19:27] Finished 'swagger:generate_client' after 1.58 s
[22:19:27] Finished 'swagger:generate' after 1.9 s
[22:19:27] Starting 'esbuild:build'...
Build started
...
Build ended: 4795ms
[22:19:32] Finished 'esbuild:build' after 5.26 s
[22:19:32] Finished 'build' after 7.16 s
Done in 7.74s.
```

In dev mode, esbuild cost near **600ms** to make incremental build:

```
Build ended: 4633ms
Serving "build" at http://127.0.0.1:3001
Ready for changes
...
CSS change detected build/dashboard.css
[...]

PR #1199, opened 2022-03-16
Release v2022.03.16.1

shhdgit:
Main changes:
- Migrate front-end build tools from webpack to esbuild #1063 #1179
- Fix conprof profile URL #1197 
- Fix missing fields in slow query #1150 
- Telemetries of top SQL and conprof #1193
- Add some tests
- Update some deps


breeswish (approved):
What about the telemetry?

shhdgit:
> What about the telemetry?

Done~

PR #1208, opened 2022-04-02
invalidate assets cache

### LINKED ISSUE #1206: Release build must reload cache manually to work
## Bug Report

**What did you do?**

Load TiDB Dashboard webpage.

It works.

**What did you see instead?**

It will fail to work due to cache is not forced to invalidate in the latest HTML:

```html
    <link rel="stylesheet" href="/dashboard/dashboardApp.css" />
  </head>
  <body>
    <noscript>You need to enable JavaScript to run this app.</noscript>
    <script src="/dashboard/compat.js"></script>
    <div id="dashboard_page_spinner"><div class="dot-flashing"></div></div>
    <div id="root"></div>
    <script type="module" src="/dashboard/dashboardApp.js"></script>
```

master

baurine:
close #1206 , add timestamp suffix to the assets links.


ui/builder.js
```diff
@@ -164,11 +164,16 @@ if (isE2E) {
 function buildHtml(inputFilename, outputFilename) {
   let result = fs.readFileSync(inputFilename).toString()
 
-  const placeholders = ['PUBLIC_URL']
+  let placeholders = ['PUBLIC_URL']
   placeholders.forEach((key) => {
     result = result.replace(new RegExp(`%${key}%`, 'g'), process.env[key])
   })
 
+  placeholders = ['TIME_PLACE_HOLDER']
```
shhdgit, line 172: Why do we need to mutate the `placeholders` var? This seems to introduce an additional cost of understanding.

shhdgit: How about just replacing the `TIME_PLACE_HOLDER` when we replace the time instead of all placeholders.

shhdgit:
It looks like we didn't cover the code splitting file.

How about using the content hash. I forgot a benefit about the content hash is that we can use cache with unchanged code.

ui/builder.js
baurine: Make sense, let me update.

baurine:
> It looks like we didn't cover the code splitting file.
> 
> How about using the content hash for each file. I forgot that a benefit of the content hash is that we can use cache with unchanged code.

I have checked the splitted files are already added the content hash suffix.

image

shhdgit:
Oops.. So that it looks like  #1206 was caused by html cache instead of js/css...

baurine:
> Oops.. Maybe #1206 was caused by the index.html cache instead of js/css path inside the html?

Or both? OK, let me do some more checks.

baurine:
also fix the disro res assets will be cached by the browser after changing the distor res assets.

baurine:
> Oops.. Maybe #1206 was caused by the index.html cache instead of js/css path inside the html?

To disable index.html cache, I just add one line code in the index.html according to [StackOverflow answer](https://stackoverflow.com/a/34851492/2998877)

```
<meta http-equiv="Cache-control" content="no-cache, no-store, must-revalidate">
```

I don't know whether it really works because I didn't find index.html cached by the browser situation before and after adding this line code. (test method: running `make ui && UI=1 make`, then accessing `localhost:12333/dashboard`)

And I think maybe it needs the backend (PD side) support to resolve it totally, for example, adding `Cache-control: max-age=0` header in the response for index.html

breeswish:
Looks like distro is something we are likely to miss. Could we have a e2e or integration test to cover it? In this way our future work can be easier.

baurine:
> Looks like distro is something we are likely to miss. Could we have a e2e or integration test to cover it? In this way our future work can be easier.

Yep, let me create an issue to add e2e test for distro.

shhdgit (approved):

shhdgit:
> > Oops.. Maybe #1206 was caused by the index.html cache instead of js/css path inside the html?
> 
> To disable index.html cache, I just add one line code in the index.html according to [StackOverflow answer](https://stackoverflow.com/a/34851492/2998877)
> 
> ```
> <meta http-equiv="Cache-control" content="no-cache, no-store, must-revalidate">
> ```
> 
> I don't know whether it really works because I didn't find index.html cached by the browser situation before and after adding this line code. (test method: running `make ui && UI=1 make`, then accessing `localhost:12333/dashboard`)
> 
> And I think maybe it needs the backend (PD side) support to resolve it totally, for example, adding `Cache-control: max-age=0` header in the response for index.html

I misunderstand the `dashboardApp.js` meaning before. So adding the timestamp query to the js resource does solve the problem.

PR #1214, opened 2022-04-15
fix formatSQL error

### LINKED ISSUE #1157: Add unit test for `formatSql()` method
Related discussion: https://github.com/pingcap/tidb-dashboard/pull/1063#discussion_r791467106

baurine:
Fix formatSQL() error by `sql-formatter` and close #1157 as well

Currently some SQLs are failed to format, and it reports the following error:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'slice')
    at parseKey (Tokenizer.js:191:20)
    at Tokenizer2.getPlaceholderTokenWithKey (Tokenizer.js:235:21)
    at Tokenizer2.getIdentNamedPlaceholderToken (Tokenizer.js:187:19)
    at Tokenizer2.getPlaceholderToken (Tokenizer.js:182:19)
    at Tokenizer2.getNextToken (Tokenizer.js:127:147)
    at Tokenizer2.tokenize (Tokenizer.js:107:24)
    at MySqlFormatter2.format7 (Formatter.js:87:38)
    at format8 (sqlFormatter.js:74:29)
    at formatSql (index.ts:15:19)
    at index.ts:32:23
```

The root cause is discussed here: https://github.com/zeroturnaround/sql-formatter/issues/147 , and a forked repo - [prettier-sql](https://github.com/inferrinizzard/prettier-sql) resolved it.

---

I did some comparisons between `sql-formatter` and `sql-formatter-plus-plus`.

Target SQL:

```
select sleep(1);
```

|                         | webpack | vite(dev) | vite(prod) | esbuild |
| ----------------------- | ------- | --------- | ---------- | ------- |
| sql-formatter           | v       | v         | v          | v       |
| sql-formatter-plus-plus | v       | x         | x          | x       |

It reports the following error when formating the above target sql by `sql-formatter-plus-plus` with vite(dev/prod) and esbuild, reported issue to esbuild: https://github.com/evanw/esbuild/issues/2181

```
rewrite-pattern.js:50 Uncaught Error: Failed to recognize value `undefined` for property `Alphabetic`.
    at getUnicodePropertyValueSet (rewrite-pattern.js:50:9)
    at handleLoneUnicodePropertyNameOrValue (rewrite-pattern.js:68:9)
    at getUnicodePropertyEscapeSet (rewrite-pattern.js:76:9)
    at processCharacterClass (rewrite-pattern.js:160:13)
    at processTerm (rewrite-pattern.js:205:11)
    at rewrite-pattern.js:259:12
    at Array.map (<anonymous>)
    at processTerm (rewrite-pattern.js:258:26)
    at rewrite-pattern.js:259:12
    at Array.map (<anonymous>)
```

Target SQL:

```
select `topics` . `id` from `topics` left outer join `categories` on `categories` . `id` = `topics` . `category_id` where ( `topics` . `archetype` <> ? ) and ( coalesce ( `categories` . `topic_id` , ? ) <> `topics` . `id` ) and `topics` . `visible` = true and ( `topics` . `deleted_at` is ? ) and ( `topics` . `category_id` is ? or `topics` . `category_id` in ( ... ) ) and ( `topics` . `category_id` != ? ) and `topics` . `closed` = false and `topics` . `archived` = false and ( `topics` . `created_at` > ? ) order by `rand` ( ) limit ?
```

|                         | webpack | vite(dev) | vite(prod) | esbuild |
| ----------------------- | ------- | --------- | ---------- | ------- |
| sql-formatter           | x       | x         | x          | x       |
| sql-formatter-plus-plus | v       | x         | x          | x       |

It reports the beginning error by `sql-formatter` with all (webpack/vite/esbuild).

---

About unit test.

I found it's not easy to do unit test when we using esbuild. if we use `jest`, it will bring babel back, and the test result may not be correct. See above table, `sql-formatter-plus-plus` works fine with wepback but not with vite/esbuild.

So I did some a simple way, test it directly under the method definition, but only run it in dev and CI mode.


breeswish:
> It reports the following error when formating the above target sql by sql-formatter-plus-plus with vite(dev/prod) and esbuild.

So what's wrong with the things produced by vite and esbuild? Are there bugs in esbuild? What is the real expected behavior?

baurine:
> > It reports the following error when formating the above target sql by sql-formatter-plus-plus with vite(dev/prod) and esbuild.
> 
> So what's wrong with the things produced by vite and esbuild? Are there bugs in esbuild? What is the real expected behavior? Is it possible to fix esbuild, or make workarounds in sql-formatter-plus-plus to skip the bug?

I think so, I digged it but have no idea how to figure it out, I have reported issue to esbuild: https://github.com/evanw/esbuild/issues/2181 (it seems has one reply now, let me have a look)

baurine:
Thanks for the esbuild issue reply, now I have an idea to resolve it by `sql-formatter-plus-plus`, and I have PoC it successfully by the following code:

```diff
+ import { format } from 'sql-formatter-plus-plus/dist/sql-formatter'
- import { format } from 'sql-formatter-plus-plus"
```

we import from the bundled sql-formatter-plus-plus then we can avoid the dynamic require issue.

I can forked the `sql-formatter-plus-plus` to make the default entry `dist/xxx.js` instead of `lib/xxx.js`, and I can add our tidb SQL dialect formatted, will try it soon.

breeswish:
Cool! It would be better if we can dig into why esbuild cannot work with a library, by finding out a minimal reproducible example. In this way esbuild can be improved (if there are bugs) and the community can be benefited.

ui/lib/utils/sqlFormatter/index.ts
```diff
@@ -1,17 +1,9 @@
-// import TiDBSQLFormatter from './TiDBSQLFormatter'
-
-// const mySqlFormatter = new TiDBSQLFormatter({ uppercase: true })
-
-// export default function formatSql(sql?: string): string {
-//   return mySqlFormatter.format(sql || '')
-// }
-
-import { format } from 'sql-formatter'
+import { format } from '@baurine/sql-formatter-plus'
```
breeswish, line 1: We should ensure that tests are contained in order to avoid regression.

breeswish (approved):

ui/lib/utils/sqlFormatter/index.ts
baurine: ok, I will add them back in the dashboard. I added them in the `@baurine/sql-formatter-plus` as well.

ui/.env.development
```diff
@@ -1,4 +1,4 @@
-PORT=3001
+PORT=3002
```
shhdgit, line 1: Why do we need to update the env variable PORT?

ui/.env.development
```diff
@@ -7,3 +7,64 @@ REACT_APP_SENTRY_ENABLED=false
 REACT_APP_RELEASE_VERSION=unknown
 REACT_APP_MIXPANEL_HOST=https://telemetry.pingcap.com/api/v1/dashboard/report
 REACT_APP_MIXPANEL_TOKEN=
+
+INSERT INTO
```
shhdgit, line 11: What's the meaning of this SQL in the env file.

ui/.env.development
baurine: sorry, a fault operation, commit the test file.

shhdgit (approved):

PR #1216, opened 2022-04-19
fix: solve overflow doesn't work on FireFox

YiniXu9506:
Since the `<pre>` tag is inside a table row, and `overflow: auto` not working in table when using FireFox, we need to add a div inside table td with attribute `overflow: auto`. 

image

A reference about this issue can be checked [here](https://stackoverflow.com/questions/16328233/overflow-auto-does-not-work-in-firefox).


shhdgit (approved):

baurine (approved):
lgtm

PR #1217, opened 2022-04-20
enable query when conprof is disabled

baurine:
What did:
- Enable query history conprof results by changing a date even if the conprof feature is disabled


shhdgit (approved):
LGTM.

PR #1219, opened 2022-04-25
tweak(keyviz): update tidb table map immediately

shhdgit:


pkg/keyvisual/decorator/tidb.go
```diff
@@ -68,11 +68,12 @@ func (s *tidbLabelStrategy) Background(ctx context.Context) {
 	ticker := time.NewTicker(time.Minute)
 	defer ticker.Stop()
 	for {
+		s.updateMap(ctx)
```
breeswish, line 71: When PD is just started, TiDB is not available I think.

PR #1245, opened 2022-05-08
feat: multi-line logs by default

breezewish:
The time and component labels are now displayed inline, and resizeable via header:

image

Messages are displayed for 8 lines by default (expandable via clicking):

image

Log level column is removed. Border-color is now used to indicate the log level, which can reduce space occupation:

image

Components are colorized by its type and its address, so that users can easily distinguish logs from different components via its color:

image


ngaut:
Cool.

PR #1142, opened 2022-01-17
test(topsql): init

shhdgit:
Add top sql tests

PR #1250, opened 2022-05-09
fix(topsql): unstable effect hooks

shhdgit:
- Fix unstable effect hooks and reduce the use of `eslint-disable-next-line react-hooks/exhaustive-deps` to make the code more react-style.
- Fix the bug that won't fetch data when the date time is updated but the window size has not changed. It can be easily reproduced by changing the date only and fixing the date time.

I'll create a test pr for fetch data bug after #1142 merge.


ui/lib/apps/TopSQL/pages/List/List.tsx
```diff
@@ -51,92 +58,90 @@ export function TopSQLList() {
...
-      const value: [number, number] = [Math.ceil(tr[0]), Math.floor(tr[1])]
-      setTimeRange({
+      const _timeRange: TimeRange = {
         type: 'absolute',
         value,
-      })
+      }
+
+      setTimeRange(_timeRange)
+      updateData(instance, _timeRange)
```
breezewish, line 124: Maybe you can let `updateData` to be an effect of:

- selected time range is changed
- selected instance is changed

This is how useEffect build for.

Refresh can be effectively implemented as the following:

```js
setTimeRange(r => [...r])
```

shhdgit:
PTAL, thanks!
/cc @breezewish @YiniXu9506

ui/lib/apps/TopSQL/utils/useTimeWindowSize.ts
```diff
@@ -8,10 +9,11 @@ export const createUseTimeWindowSize = (barWidth: number) => {
...
     const computeTimeWindowSize = (
       screenWidth: number,
-      totalTimeRange: number
+      [min, max]: TimeRangeValue
     ) => {
-      const windowSize = (barWidth * totalTimeRange) / screenWidth
-      setTimeWindowSize(Math.ceil(windowSize))
+      const windowSize = Math.ceil((barWidth * (max - min)) / screenWidth)
+      setTimeWindowSize(windowSize)
+      return windowSize
```
breezewish, line 16: I have extracted a `useChartHandle.calcIntervalSec`, which may be useful for you. BTW the time window size should be a ref instead of a state, because when it is changed nothing should be re-rendered. The time window size is only something used at the request moment I believe.

breezewish (commented):
Mostly good. Some optional engineering suggestions.

ui/lib/apps/TopSQL/pages/List/List.tsx
```diff
@@ -51,92 +58,90 @@ export function TopSQLList() {
...
+      }
 
-    if (!instance || !timeWindowSize) {
-      return
-    }
+      // Update container width when refresh, then trigger the update of time window size to refresh.
+      const cw = containerRef.current?.offsetWidth || containerWidth
+      if (cw !== containerWidth) {
+        setContainerWidth(containerRef.current?.offsetWidth || 0)
+      }
```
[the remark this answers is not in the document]
shhdgit: Yes! ref is most accurate what I want(instead of global bars). I’m really poor at react. 😭

ui/lib/apps/TopSQL/pages/List/List.tsx
shhdgit: Impressive! I was extremes of observer and event emit. Poor react again.

ui/lib/apps/TopSQL/pages/List/List.tsx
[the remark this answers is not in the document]
breezewish: I found myself easier to judge in many cases when thinking that state will cause re-render and ref will not.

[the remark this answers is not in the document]
shhdgit: Very helpful!

ui/lib/apps/TopSQL/utils/useTimeWindowSize.ts
shhdgit: There's a feature that can resize the bar series width and adjust the window size depending on the container width.

ui/lib/apps/TopSQL/pages/List/List.tsx
```diff
@@ -392,42 +352,46 @@ const useTopSQLConfig = () => {
...
+    .topsqlInstancesGet(String(end), String(start))
+  const instances = resp.data.data || []
+  instances.sort((a, b) => {
+    const localCompare = a.instance_type!.localeCompare(b.instance_type!)
+    if (localCompare === 0) {
+      return a.instance!.localeCompare(b.instance!)
+    }
+    return localCompare
+  })
+  return instances
```
breezewish, line 375: May be simplified to:
```js
return sortBy(instance, ['instance_type', 'instance])
```

ui/lib/apps/TopSQL/pages/List/List.tsx
```diff
@@ -335,10 +291,14 @@ const useTopSQLData = () => {
 
       setTopSQLData(data)
     },
-    []
+    [containerRef]
```
breezewish, line 294: a ref doesn't need to be placed in dep array

ui/lib/apps/TopSQL/pages/List/List.tsx
```diff
@@ -282,34 +225,47 @@ export function TopSQLList() {
...
-      timestamps,
-      timeWindowSize,
-      topN,
-    }: {
-      instance: TopsqlInstanceItem
-      timestamps: [number, number]
-      timeWindowSize: number
-      topN: number
-    }) => {
+    async (_instance: TopsqlInstanceItem | null, _timeRange: TimeRange) => {
```
breezewish, line 246: just curious, in what case will you add a `_` prefix? It is usually used to indicate the parameter is not used I think.

breezewish (approved):

ui/lib/apps/TopSQL/pages/List/List.tsx
shhdgit: Cool..

ui/lib/apps/TopSQL/pages/List/List.tsx
shhdgit: emm.. I use it to implicit an internal variable if there is the same variable outside. So that the variables won't be missed when refactor.

PR #1251, opened 2022-05-10
prevent enter overview page before auth

### LINKED ISSUE #227: UCP: Avoid screen flash when session is timedout
## Description

When session is not valid, Dashboard will show overview page in an instant, then redirect to sign in page. This "flash" can be improved.

## Score

* 300

## Mentor(s)

* @baurine 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* React
* Axios Library

baurine:
close #227


ui/dashboardApp/index.ts
```diff
@@ -156,9 +156,9 @@ async function webPageStart() {
   // NOTE: Don't remove above comment line, it is a placeholder for code generator
 
   try {
-    await reloadWhoAmI()
+    const ok = await reloadWhoAmI()
```
breezewish, line 159: The spinner has already been removed in L65, will user still see a flash screen during these two requests?

breezewish (commented):
Interesting! How does it work?

shhdgit:
Oh, it's like we caught the error but did not throw the error. So the caller will never meet an error and go through to the next.

ui/lib/utils/store.ts
```diff
@@ -45,10 +45,12 @@ export async function reloadWhoAmI() {
     store.update((s) => {
       s.whoAmI = resp.data
     })
+    return true
   } catch (ex) {
     store.update((s) => {
       s.whoAmI = undefined
     })
```
shhdgit, line 52: How about throwing the error directly?

ui/lib/utils/store.ts
```diff
@@ -30,12 +30,12 @@ export const useIsFeatureSupport = (feature: string) =>
 
 export const useNgmState = () => store.useState((s) => s.appInfo?.ngm_state)
 
-export async function reloadWhoAmI() {
+export async function reloadWhoAmI(): Promise<boolean> {
   if (!getAuthToken()) {
     store.update((s) => {
       s.whoAmI = undefined
     })
-    return
```
shhdgit, line 38: Same here.

baurine:
> Interesting! How does it work?

At before, when auth token is empty, `reloadWhoamI()` will return directly  by the following code:

```
  if (!getAuthToken()) {
    store.update((s) => {
      s.whoAmI = undefined
    })
    return
  }
```

So the page will redirect from `/` to `/overview`, then the `/overview` page APIs will get 401 immediately, then jump back to `/signin` quickly.

Use boolean value as `reloadWhoamI()` return value to indicate whether request user info successfully, if failed, we don't redirect to `/overview` to avoid no-effort work.

ui/lib/utils/store.ts
baurine: I think using boolean as the return value may be more explicit.

Issue #1209, opened 2022-04-06
supports QPS/CPU dimension in keyviz


nolouch:
## Feature Request

**Is your feature request related to a problem? Please describe:**
Currently，Keyviz has `keys` and `bytes` dimensions. In some situations, it cannot find the hotspot that the CPU is high caused by QPS, the `keys` and `bytes` maybe 0. For example:
```
create table t(id int);
insert into table t values (1)(2)....(1000);
// random read out bound
select * from t where id = 1000 + rand(0,1000)
```

**Describe the feature you'd like:**

- add QPS dimension
- add CPU dimension

**Describe alternatives you've considered:**

we can add QPS first. CPU after the TopSQL enable by default.

**Teachability, Documentation, Adoption, Migration Strategy:**

- Go

nolouch:
cc @qidi1 @IcePigZDB

breeswish:
Currently we have undecided about the future iteration of KeyViz, due to its requirement of professional ability. According to the scenario you described, how about simply using Top SQL to observe such hot spots directly?

nolouch:
Yes, TopSQL can tell me that this SQL is consuming a lot of CPU. It's useful to the user, but it's not enough for an advanced user or expert. TopSQL cannot say it is a hotspot, the data may access randomly, and there is enough data to be distributed evenly on the TiKV nodes. so this type of SQL does consume a lot of CPU, but it may not cause a hotspot. We still need `KeyViz` to know the workload pattern to help understand the hotspot. 
But I agree that it requires professional ability. Here, I think I propose to add the dimension not break the current design. and there are two community people who are interesting in it can help to add it. What do you think about letting them have a try?
cc @breeswish

breeswish:
Thanks for the supply information. I totally agree "We still need KeyViz to show the workload pattern to help understand the hotspot". This looks to be extremely useful for PD developers to understand why such hotspots are not scattered.

> there are two community people who are interesting in it can help to add it. What do you think about letting them have a try?

Contributions are always welcome :) We can help reviewing PRs. I'm not sure whether there are CPU usage information in TiKV heartbeat. If such information is available, then the change may be small.

PR #1252, opened 2022-05-11
add QPS in tidb_dashboard ref:#1209

qidi1:
Signed-off-by: qidi1 <1083369179@qq.com>

### What problem does this PR solve?

Currently，Keyviz has `keys` and `bytes` dimensions. In some situations, it cannot find the hotspot that the CPU is high caused by QPS, the `keys` and `bytes` may be 0. For example:

```sql
create table t(id int);
insert into table t values (1)(2)....(1000);
// random read out bound，suppose there are many requests.
select * from t where id = 1000 + rand(0,1000)
```

## 
Issue Number:  #1209
### What is changed and how does it work?

 Return QPS data in PD.

### Check List

Tests

- Manual test 

Related changes
- https://github.com/tikv/pd/pull/4922
- https://github.com/tikv/pd/pull/4921
- https://github.com/pingcap/kvproto/pull/914
### Release note

```release-note
None
```

## Manual test
We ran a small cluster locally. On the cluster, test data was generated using the tiup bench tool. The results are shown in the figure below.
image
image

### DIFF
--- a/pkg/keyvisual/input/api.go
+++ b/pkg/keyvisual/input/api.go
@@ -29,6 +29,8 @@ type RegionInfo struct {
 	ReadBytes       uint64 `json:"read_bytes"`
 	WrittenKeys     uint64 `json:"written_keys"`
 	ReadKeys        uint64 `json:"read_keys"`
+	WriteQueryNum   uint64 `json:"write_query_num"`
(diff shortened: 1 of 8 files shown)


breezewish:
This is very cool! Will you also propose the CPU dimension in this PR, or in later PRs?

PR #1234, opened 2022-05-04
feat: improve statement

breezewish:
- Fix #1229
- Fix #1143, fix #848: By checking for overlap time ranges instead of exact equality
- Fix #858
- Ref #1233 (implements it for Statement)

The UI now looks like:

image

Notable UI changes:

1. The refresh button is replaced by a more notable "Query" button. It will be easier to use for #1233
2. A hint string about current time range is added for #848

If the request takes longer than **2 seconds** to finish, the on-the-fly request will be disabled. User can manually send a request by clicking the "Query" button after changing TimeRange, Filters or VisibleColumns. A notice will be shown as:

image

PR #1254, opened 2022-05-15
feat: do not autoload data when network slow in slow query

### LINKED ISSUE #862: Slow query refresh doesn't work
## Bug Report

**What did you do?**

1. Open slow query page, it is empty first
2. Run `select sleep(1);` in the MySQL client
3. Click the refresh button on the slow query page
4. We can't see the expected slow query

We should see a non-empty slow query list with at least one item: `select sleep(1)`

**What did you see instead?**

An empty slow query list

master

### LINKED ISSUE #1233: Disable on-the-fly request in bad network conditions
## Feature Request

When TiDB responds slowly, or the network is slow, it would be annoying if users sees a Skeleton for  long time what ever change it makes in Statement and Slow Query.

It would be better to allow users to manually send a request in such condition. When user modifies the options (like filters), on-the-fly request should not happen and current data are preserved.

breezewish:
This PR brings #1234 to slow query.

Fix #862

Resolve #1233


ui/lib/components/TimeRangeSelector/index.tsx
```diff
@@ -55,7 +55,7 @@ export function toTimeRangeValue(timeRange?: TimeRange): TimeRangeValue {
     return [...t2.value]
   } else {
     const now = dayjs().unix()
-    return [now - t2.value, now]
+    return [now - t2.value, now + 1]
```
breezewish, line 58: This seems to be the root cause that you need some wait() in integration tests before. @YiniXu9506 @baurine @shhdgit 

The reason is simple: the current second (e.g. ts=123.56) is not yet finished, so that data happen in this second got truncated when using this second (e.g. ts=123.0) as the end time.

ui/cypress/integration/slow_query/01-list.spec.js
```diff
@@ -537,6 +538,102 @@ describe('SlowQuery list page', () => {
...
   })
 
+  // FIXME: The following tests will break slow-query details E2E since it executes a SQL.
+  // Fix the slow-query details E2E first.
+
+  // describe('Slow network condition', () => {
+  //   const slowNetworkText = 'On-the-fly update is disabled'
+
+  //   it('Does not show slow information when network is fast', () => {
+  //     cy.intercept(`${Cypress.env('apiBasePath')}slow_query/list*`).as(
```
breezewish, line 548: These tests are not introduced, since introducing new tests simply breaks the slow query details page E2E. This is horrible. I will not add new E2E tests to slow query until it is fixed.

cc @baurine @YiniXu9506

ui/lib/components/TimeRangeSelector/index.tsx
shhdgit, line 58: Does `+1` caused by the grafana time range rules?

shhdgit: But that's wired we change what the recent means. When the user chooses the recent 30s option, we use the recent 31s indeed.

breezewish: No, we are not changing the start time. It is still recent 30s.

shhdgit: Oh, I got the wait issue here. Maybe we can use `cy.clock` to specify the global clock.

breezewish: cy.clock will not change what TiDB stores I guess.

shhdgit: Yes, but we delayed the timing of the request by one second. emm, it is just like what `wait` do but don't wait.

shhdgit: > No, we are not changing the start time. It is still recent 30s.

I got you.. Should recent "30s" imply that there will be 30 "1s" intervals?

shhdgit: > > No, we are not changing the start time. It is still recent 30s.
> 
> I got you.. Should recent "30s" imply that there will be 30 "1s" intervals?

It doesn't seem to be necessary...

shhdgit: But if we consider the network latency, the user will get the recent 31s data. We have no idea how a common component will be used, so I thought it would be appropriate to guarantee the semantics of the 30 "1s" interval.

breezewish: The real problem is that we are truncating sub-seconds in this component, e.g. 1.466s becomes 1s. This introduces precision lost in the previous "recent X" options.

Additionally, we didn't specify the open-close rule for the range. Is it right closed or right open? We don't know.

If we want to precisely control the "recent" behavior, the above two questions must be resolved first.

breezewish: As a result, as you can see from the test result, the "recent X minutes" does not include the things just happened in this second. This is a bug that doesn't conform to the "recent" semantic totally.

shhdgit: > As a result, as you can see from the test result, the "recent X minutes" does not include the things just happened in this second. This is a bug that doesn't conform to the "recent" semantic totally.

I found the "recent" bug too. We should fix the "recent" issue before we do some "recent" related test...

PR #1259, opened 2022-05-16
misc: new landing image

breezewish:
As we have so many changes in recent versions (since v5.4), I dropped a new landing image, to bring user more impression about we have some big changes.

image

Also this new mesh-gradient landing image does not have a TiDB icon on it, makes it easier for 3rd-party distributions.


shhdgit (approved):
Wow! Amazing!

PR #1264, opened 2022-05-18
fix: unstable tests

### LINKED ISSUE #1174: Switch to use a stable toolkit in tests
In `download_tools.sh`, a nightly toolkit is introduced. However the nightly is not ensured to be working, simply as TiDB Dashboard. It exposes risks to our test stability.

shhdgit:
- fix unstable tests
- close #1174


ui/cypress/integration/topsql/topsql.spec.ts
```diff
@@ -167,13 +167,11 @@ skipOn(Cypress.env('TIDB_VERSION') !== 'nightly', () => {
         cy.wait('@getTopsqlSummary')
```
breezewish, line 167: Maybe you should test with seconds with sub-seconds for better coverage, e.g. now * 1000 + 456.

BTW there is indeed time range problems in Top SQL related with subseconds, as the displayed time range will vary within 59s, 60s, 61s when using the brush tool.

.github/workflows/test.yaml
```diff
@@ -77,10 +77,10 @@ jobs:
         # test latest features and compatibility of lower version
         include:
           - feature_version: 6.0.0
-            tidb_version: nightly
+            tidb_version: latest
```
breezewish, line 80: latest and nightly are different. latest should be pointing to stable release only I guess.

breezewish (commented):
It's interesting that the image size is also changed.. I'm not sure where it comes from.

.github/workflows/test.yaml
breezewish: How about introducing both latest and nightly to our matrix?

IMO currently the source of unstable comes from our tests, instead of nightly.

shhdgit: Make sense.

I found some unstable cases when restarting tiup playground nightly in m1. It can be resolved by the cypress default env variable.

shhdgit:
> It's interesting that the image size is also changed.. I'm not sure where it comes from.

Oh, I updated the snapshots for the stability of screenshots. It's will be clipped, if it is not the default viewport configuration.

breezewish (approved):

breezewish:
Merging, cause v6.0.0 branch looks to be stabled.

PR #1268, opened 2022-05-19
fix(keyviz): find nearest id if not found

shhdgit:


pkg/keyvisual/decorator/tidb.go
```diff
@@ -52,14 +53,95 @@ type tidbLabelStrategy struct {
...
+	right := tLen
+	for pivot > left {
+		prevID := inOrder.tables[pivot-1].ID
+		id := inOrder.tables[pivot].ID
+		// find approaching id near the fromId
+		// table_1 ------- table_3 ------- table_5
+		//           	^
+		//       search table_2
+		// approaching result: table_3
+		if prevID < fromID && id >= fromID {
```
breezewish, line 105: id >= toID？

breezewish:
Please fix the lint.

Also I believe you need more tests to cover the "findOne" function. For such complicated logic function, maybe better to ensure 100% branch coverage.

breezewish:
How about using sort.Search to discover a lower bound, and then whether it meets upper bound? In this way you don't need to manually write the binary search or lower_bound algorithm.

Here is an example of using sort.Search to implement lower_bound: https://go-review.googlesource.com/c/go/+/29333/3/src/sort/example_search_test.go#45

PR #1271, opened 2022-05-20
tweak: cy.exec wait tiup restart timeout

shhdgit:
The default timeout of exec task `wait_tiup_playground` is 60000ms which is not enough to restart tiup playground.


breezewish (approved):

shhdgit:
Emm.. Still failed. Let me investigate it.

breezewish:
The failed E2E is not related with tiup timeout. We can fix the unstable case by case in different PRs.

PR #1280, opened 2022-06-07
pkg: make interval of keyviz fetcher be configurable

nolouch:
In some cases, we want to speed up the keyviz shows. add a config item to control the  interval of keyviz fetcher.
Signed-off-by: nolouch <nolouch@gmail.com>

### DIFF
--- a/pkg/keyvisual/input/periodic.go
+++ b/pkg/keyvisual/input/periodic.go
@@ -15,11 +15,13 @@ import (
 
 type periodicInput struct {
 	PeriodicGetter region.RegionsInfoGenerator
+	interval       time.Duration
 }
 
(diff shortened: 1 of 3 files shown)


pkg/keyvisual/input/periodic.go
```diff
@@ -28,7 +30,8 @@ func (input *periodicInput) GetStartTime() time.Time {
 }
 
 func (input *periodicInput) Background(ctx context.Context, stat *storage.Stat) {
-	ticker := time.NewTicker(time.Minute)
+	ticker := time.NewTicker(input.interval)
+	log.Info("keyvisual load regions from periodic input", zap.Duration("interval", input.interval))
```
breezewish, line 34: How about using the dynamic config? Currently all KeyViz configurations are placed there, so that it can be updated dynamically.

PR #1327, opened 2022-07-12
fix: compatibility on safari

YiniXu9506:


shhdgit:
brilliant job! does this mean we can't use `position: absolute` inside the foreignObject.

shhdgit (approved):

PR #1338, opened 2022-07-18
Refine metrics

shhdgit:
- we will find orphan data points caused by `null` value

**Want**:
image

**Now**:
image


shhdgit:
PTAL, thanks!
/cc @breezewish @YiniXu9506

YiniXu9506 (approved):
LGTM

breezewish:
Why is it a correct behavior to fill null as 0? IMO if there is no requests, latency should not be 0, but should be null.

PR #1346, opened 2022-07-19
feat: tidb dashboard for clinic

baurine:
What did:
- Make tidb dashboard reusable for clinic


ui/packages/tidb-dashboard-for-clinic/.env.production
```diff
@@ -0,0 +1,5 @@
+REACT_APP_VERSION=$npm_package_version
+REACT_APP_MIXPANEL_HOST=https://telemetry.pingcap.com/api/v1/dashboard/report
+# TODO: Move to GitHub Secrets
+# DO NOT COPY THE TOKEN TO OTHER PROJECTS
+REACT_APP_MIXPANEL_TOKEN=
```
shhdgit, line 5: How do we distinguish tokens of different projects in a monolithic dashboard repository?

ui/packages/clinic-client/swagger/spec.json
```diff
@@ -0,0 +1,180 @@
+{
```
shhdgit, line 1: This means we need to sync the clinic or cloud api specs from those projects. If we still need to maintain the multi projects in separate repos, why don't we put the dashboard-for-x codes in the `x` repository?

ui/packages/tidb-dashboard-for-clinic/src/client/index.tsx
```diff
@@ -0,0 +1,124 @@
+import React from 'react'
```
shhdgit, line 1: Like swagger specs, we still need many duplicated codes between project `x` and dashboard-for-x.

ui/packages/tidb-dashboard-for-clinic/.env.production
baurine: oh, I missed its value.

PR #1348, opened 2022-07-20
fix conflict

srstack:


Makefile
```diff
@@ -4,7 +4,7 @@ BUILD_TAGS ?=
 
 LDFLAGS ?=
 
-FEATURE_VERSION ?= 6.2.0
```
shhdgit, line 7: Why do we need to roll back the default value of FEATURE_VERSION here?

PR #1374, opened 2022-07-27
refine: build scenes

shhdgit:
According to https://pnpm.io/filtering#--filter-package_name-1, we can run the command only on the filtered package and all of its dependencies.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -12,6 +12,8 @@
     "prepare": "cd .. && husky install ui/.husky",
     "build:lib": "pnpm -r --filter @pingcap/tidb-dashboard-lib build",
     "dev": "pnpm build:lib && pnpm -r --parallel dev",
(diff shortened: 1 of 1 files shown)


ui/package.json
```diff
@@ -12,6 +12,8 @@
     "prepare": "cd .. && husky install ui/.husky",
     "build:lib": "pnpm -r --filter @pingcap/tidb-dashboard-lib build",
     "dev": "pnpm build:lib && pnpm -r --parallel dev",
+    "dev:op": "pnpm build:lib && pnpm -r --parallel --filter @pingcap/tidb-dashboard-for-op... dev",
+    "dev:dbaas": "pnpm build:lib && pnpm -r --parallel --filter @pingcap/tidb-dashboard-for-dbaas... dev",
```
baurine, line 16: hi @shhdgit , can you attach some doc links to present the `...` meaning in the PR description. Although we can google it, it is more convenient for reviewers.

shhdgit: That's right! Thanks!

baurine (approved):

PR #1378, opened 2022-07-28
fix api/visualplan/generate Content-Type

srstack:
image

image


shhdgit (approved):

pkg/apiserver/visualplan/service.go
```diff
@@ -63,5 +63,5 @@ func (s *Service) GenerateVisualPlan(c *gin.Context) {
 		rest.Error(c, rest.ErrBadRequest.New("generate visual plan failed: %v", err))
 		return
 	}
-	c.String(http.StatusOK, vp)
+	c.Data(http.StatusOK, gin.MIMEJSON, []byte(vp))
```
baurine, line 66: why not `c.JSON()` directly?

PR #1382, opened 2022-08-02
*:upgrade mockery to make sure this project builds under go 1.18

longfangsong:
close #1366 (replace it)

Current version of `mockery` used by this project [does not support go 1.18](https://github.com/vektra/mockery/issues/434). According to the [suggestion](https://github.com/pingcap/tidb-dashboard/pull/1366#issuecomment-1200749996) of the maitainer of this project, we should upgrade it and drop the support of old Golang compiler.

This PR do the upgrade work and push the minium supported Golang version to 1.18


breezewish:
I guess the Golang compiler version specified in the CI yaml also need to be updated.

scripts/lint.sh
```diff
@@ -26,6 +26,7 @@ if [ "${NEED_DOWNLOAD}" = true ]; then
 fi
 
 echo "+ Run lints for Go source code"
+go get ./...
```
longfangsong, line 29: ref golangci-lint#865

coderplay (approved):
Thanks for adding this

PR #1384, opened 2022-08-08
chore: modify primary color

baurine:
What did:
- modify tidb dashboard primary color, to keep consistency with dbaas ui

Preview:

image


ui/packages/tidb-dashboard-for-dbaas/builder.js
```diff
@@ -131,6 +114,22 @@ function handleAssets() {
   updateHtmlFiles(htmlFiles)
 }
 
+function copyAssets() {
+  // copy out dir to dbaas ui repo
+  // why we copy to dbaas ui publich folder instead of dist folder
```
YiniXu9506, line 119:
```suggestion
  // why we copy to dbaas ui public folder instead of dist folder
```

ui/packages/tidb-dashboard-for-dbaas/builder.js
```diff
@@ -131,6 +114,22 @@ function handleAssets() {
...
+  // because dbaas ui use create-react-app, it doesn't write output to disk in dev mode
```
YiniXu9506, line 120:
```suggestion
  // because dbaas ui use create-react-app, it doesn't write output to dist in dev mode
```

baurine: here `disk` means local file system, create-react-app put the output files in the memory instead of local files when in dev mode, so we can't copy the `tidb-dashboard-for-dbaas/dist` to the dbaas ui dist folder, because it has no dist folder. while for clinic, we copy the `tidb-dashboard-for-clinc/dist` to the clinic ui dist folder directly.

shhdgit (approved):

PR #1393, opened 2022-08-22
feat: support set and show time zone

baurine:
## What did

Support to set and show timezone information for date-time, to keep consistency with dbaas.

## Preview

Local time: 12:00 UTC+8:00，DBaas timezone: UTC+0:00

image

image

Related doc: https://day.js.org/docs/zh-CN/manipulate/utc-offset


ui/packages/tidb-dashboard-lib/src/utils/timezone.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+  return _tz
+}
+
+function getTimeZoneStr() {
+  const z = getTimeZone()
+  if (getTimeZone() >= 0) {
+    return `utc+${z}`
+  }
+  return `utc-${z}`
```
baurine, line 28: TODO: fix

ui/packages/tidb-dashboard-lib/src/utils/timezone.ts
```diff
@@ -0,0 +1,39 @@
+import dayjs from 'dayjs'
+import utc from 'dayjs/plugin/utc'
+
+dayjs.extend(utc)
+
+// value range: -16 ~ 16
```
shhdgit, line 6: Why the timezone range is -16 ~ 16? Are there any references?

baurine: https://day.js.org/docs/en/manipulate/utc-offset

> If the input is less than 16 and greater than -16, it will interpret your input as hours instead.

shhdgit (approved):

YiniXu9506 (approved):
LGTM

PR #1400, opened 2022-09-01
fix: recover diagnosis report for clinic cloud

baurine:


ui/packages/tidb-dashboard-lib/src/apps/SystemReport/context/index.ts
```diff
@@ -29,8 +29,13 @@ export interface ISystemReportDataSource {
   ): AxiosPromise<DiagnoseReport>
 }
 
+export interface ISystemReportEvent {
+  getFullReportLink(reportID: string): string
+}
```
shhdgit, line 34: Why `getFullReportLink` is an event? Since no one subscribed it out of the app.

ui/packages/tidb-dashboard-lib/src/apps/SystemReport/context/index.ts
```diff
@@ -29,8 +29,13 @@ export interface ISystemReportDataSource {
...
+
 export interface ISystemReportContext {
   ds: ISystemReportDataSource
+  event: ISystemReportEvent
   cfg: IContextConfig & { publicPathBase: string }
```
shhdgit, line 39: How about put the `getFullReportLink ` into cfg?

ui/packages/tidb-dashboard-lib/src/apps/SystemReport/context/index.ts
baurine: Yep, it's a bad name, do you have any suggestion for a better name?

ui/packages/tidb-dashboard-lib/src/apps/SystemReport/context/index.ts
baurine: Good idea!

PR #1407, opened 2022-09-13
refine: misc refines

baurine:
## What Did

- Handle get alert error custom, not show notification error
- Show statement/slow query search result count
- Show empty status for slow query

## Preview

image

image

image


ui/packages/tidb-dashboard-lib/src/apps/SlowQuery/pages/List/index.tsx
```diff
@@ -9,13 +9,16 @@ import {
   Menu,
   Dropdown,
   Alert,
-  Tooltip
+  Tooltip,
+  Result,
+  Button
```
YiniXu9506, line 14: Seems the `Button` not been used.

YiniXu9506 (approved):
LGTM

ui/packages/tidb-dashboard-lib/src/apps/SlowQuery/translations/en.yaml
```diff
@@ -109,6 +109,10 @@ slow_query:
     status:
       success: Success
       error: Failed
+  overview:
+    empty_result: Great, the cluster has no any related slow queries!
```
shhdgit, line 113: Why do we congratulate the user when they did not get the result they want? How about showing empty result tips only?

shhdgit (approved):
The rest LGTM.

ui/packages/tidb-dashboard-lib/src/apps/SlowQuery/translations/en.yaml
baurine: Is it not a good news if the cluster has no slow queries? ^_^

PR #1414, opened 2022-09-28
refine clinic cloud variant

baurine:
## What did

- Extend the start options, make it more configurable from outside, to support cloud ngm slow query and top sql app


ui/packages/tidb-dashboard-for-clinic-cloud/src/apps/SlowQuery/index.tsx
```diff
@@ -1,10 +1,13 @@
 import React from 'react'
 import { SlowQueryApp, SlowQueryProvider } from '@pingcap/tidb-dashboard-lib'
 import { ctx } from './context'
+import { getStartOptions } from '~/uilts/appOptions'
```
baurine, line 4: TODO: refine

PR #1417, opened 2022-10-14
fix: make debug api compatible with old tidb-dashboard backend api

baurine:


baurine:
remain todo: (will fix in another PR)

- fix i18n

image

PR #1427, opened 2022-10-26
Misc refines

baurine:
## What Did

- add deployType header for o11y slow query
- rename StartOptions to GlobalConfig for tidb-dashboard-for-clinic-cloud
- skip query slow query detail api to improve experience if list return detail
- fix show statement result length error
- let backend order for slow query


ui/packages/tidb-dashboard-lib/src/apps/SlowQuery/utils/useSlowQueryTableController.ts
```diff
@@ -191,7 +191,7 @@ export default function useSlowQueryTableController({
       // The cache key is built over queryOptions, instead of evaluated one.
       // So that when passing in same relative times options (e.g. Recent 15min)
       // the cache can be reused.
-      const cacheKey = JSON.stringify(queryOptions)
+      const cacheKey = JSON.stringify({ queryOptions, orderOptions })
```
baurine, line 194: For triggering request api if order options change

ui/packages/tidb-dashboard-for-clinic-cloud/src/apps/SlowQuery/context.ts
```diff
@@ -52,31 +55,59 @@ class DataSource implements ISlowQueryDataSource {
...
   slowQueryDownloadTokenPost(request: any, options?: ReqConfig) {
     return client.getInstance().slowQueryDownloadTokenPost({ request }, options)
   }
 }
 
-const ds = new DataSource()
+class EventHandler implements ISlowQueryEvent {
+  constructor(
+    public listApiReturnDetail: boolean,
+    public cache: SlowqueryModel[]
```
shhdgit, line 88: In which case we need an array cache？

baurine: You mean why the cache is array while we only need to cache one object, right? Because we need to share a same reference between DataSource and EventHandler, if we use `public cache: SlowqueryModel | undefined = undefined`, it can't be shared between DataSource and EventHandler. So I choose to use an array, Of course, map is fine as well.

shhdgit (approved):

PR #1428, opened 2022-10-26
fix: make new ui compatiable with old backend api for statement

baurine:


baurine:
This PR is replaced by #1431 , so close it.

PR #927, opened 2021-05-17
refine(ui): debug api

shhdgit:
- [x] debug api layout
- [x] table/db widget
- [x] add new tidb/pd endpoints
- [ ] add descriptions

https://user-images.githubusercontent.com/11549583/118844643-b18d5b80-b8fd-11eb-83fd-d5c4b9f2f5d1.mov

PR #1431, opened 2022-10-28
refine: improve compatibility

baurine:
## What Did

Make new ui can work with old tidb-dashboard backend

- improve compatibility for statement, work with old tidb-dashboard backend early to at least 5.0.0
- improve compatibility for debug api, work with old-tidb-dashboard backend early to 5.0.3
- improve compatibility for sign in

statement changes its api by PR #1234 , brings to >=5.4.2 and >=6.1.0

debug api is first introduced in 5.0.0 and changes its api in 5.0.3 ( by PR #927 ) and 5.4.0 ( by PR #1103 )

### DIFF
--- a/ui/packages/tidb-dashboard-for-clinic-cloud/src/apps/Statement/context.ts
+++ b/ui/packages/tidb-dashboard-for-clinic-cloud/src/apps/Statement/context.ts
@@ -1,10 +1,8 @@
 import {
   IStatementDataSource,
   IStatementContext,
-  ReqConfig,
-  StatementTimeRange
+  ReqConfig
 } from '@pingcap/tidb-dashboard-lib'
-import { AxiosPromise } from 'axios'
 
 import client, {
   StatementEditableConfig,
--- a/ui/packages/tidb-dashboard-for-clinic-cloud/src/apps/Statement/context.ts
+++ b/ui/packages/tidb-dashboard-for-clinic-cloud/src/apps/Statement/context.ts
@@ -1,8 +1,10 @@
 import {
   IStatementDataSource,
   IStatementContext,
-  ReqConfig
+  ReqConfig,
+  StatementTimeRange
 } from '@pingcap/tidb-dashboard-lib'
+import { AxiosPromise } from 'axios'
 
 import client, {
   StatementEditableConfig,
@@ -101,6 +103,10 @@ class DataSource implements IStatementDataSource {
 
   statementsStmtTypesGet(options?: ReqConfig) {
     return client.getInstance().statementsStmtTypesGet(options)
+  }
+
+  statementsTimeRangesGet(options?: ReqConfig) {
+    return client.getAxiosInstance().get('/statements/time_ranges', options)
   }
 
   // slow query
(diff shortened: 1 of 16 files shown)


ui/packages/tidb-dashboard-lib/src/apps/Statement/utils/useStatementTableController.ts
```diff
@@ -235,6 +248,11 @@ export default function useStatementTableController({
       setDataLoading(true)
 
       const timeRange = toTimeRangeValue(queryOptions.timeRange)
+      // enlarge the time range automatically for old tidb-dashboard backend
+      if (oldBackend) {
+        timeRange[0] -= windowSize
+        timeRange[1] += windowSize
+      }
```
shhdgit, line 255: Is that ok for the latest api at first? If it is good, why not just use the expanded time range?

ui/packages/tidb-dashboard-lib/src/apps/Statement/utils/useStatementTableController.ts
```diff
@@ -250,6 +268,12 @@ export default function useStatementTableController({
           list: res?.data || [],
           timeRange
         }
+        // if we have checked it is new backend (aka oldBackend === false)
+        // we don't check it again
+        if (oldBackend) {
+          // old backend api has no `summary_begin_time` field
+          setOldBackend(data.list[0]?.summary_begin_time === undefined)
+        }
```
shhdgit, line 276: The `oldBackend` will be set to `true` when `data.list` is empty at the latest backend. Is that ok for the next round request?

baurine: oh, nice catch! need to fix.

ui/packages/tidb-dashboard-lib/src/apps/Statement/utils/useStatementTableController.ts
baurine: > Is that ok for the latest api at first?

for the new api, after we check it is the new api by the first request, it will change `oldBackend` from `true` to `false`, and trigger re-render() and re-request again. so the list result can be fixed very quickly.

> why not just use the exapended time range?

I think use the original time range for the new api can make the result more accurate.

ui/packages/tidb-dashboard-lib/src/apps/Statement/utils/useStatementTableController.ts
baurine: and I am thinking to get new api or old api by checking api `/time_ranges` exist or not.

shhdgit: > and I am thinking to get new api or old api by checking api /time_ranges exist or not.

Agree.

shhdgit (approved):

ui/packages/tidb-dashboard-for-clinic-cloud/src/apps/Statement/context.ts
```diff
@@ -1,8 +1,10 @@
 import {
   IStatementDataSource,
   IStatementContext,
-  ReqConfig
+  ReqConfig,
+  StatementTimeRange
```
baurine, line 5: TODO: remove

PR #1469, opened 2023-01-13
chore: update tooltip when failed to get host info

### LINKED ISSUE #926: cluster-info: disks showing host down message for tidb
## Bug Report

**What did you do?**

Deploy TiDB with the TiDB Operator on AWS EKS. Then use the dashboard.

Disk info for all node types

**What did you see instead?**

"Host information is unavailable due to nstances on the host is down"

image

In the debugger it does show partitions. So the host is not down.

The top left corner shows "PD v5.0.1"

baurine:
close #926 

## What Did

- Update tooltip when failed to get host info, because the reason is not correct in some cases

## Preview

screenshot-20230113-181225


baurine:
This fix need to bring to TiDB 6.6 
/cc @shhdgit @SabaPing

shhdgit (approved):

PR #1496, opened 2023-02-23
Hotifx Index Advisor

YiniXu9506:
- update index advisor list table pagination.
- give tips when user click checkup, and skip the tips when user choose `Don't remind me again`.
- update basic information content.


ui/packages/tidb-dashboard-lib/src/apps/SQLAdvisor/component/IndexInsightList.tsx
```diff
@@ -34,8 +35,11 @@ const IndexInsightList = ({
   const [showDeactivateModal, setShowDeactivateModal] = useState<boolean>(false)
   const [showSetting, setShowSetting] = useState(false)
   const [showCheckUpModal, setShowCheckUpModal] = useState(false)
-
+  const [dontRemindCheckUpNotice, setDontRemindCheckUpNotice] = useState<
+    string | boolean
+  >(localStorage.getItem('dont_remind_checkup_notice') || false)
```
baurine, line 40: I feel access local storage is a little expensive in every render, so maybe  it's better to use `useState(() => localStorage.getItem('...') || false)`, it only run one time, instead of every time when re-rendering.

baurine: It's better to add some prefix to the local storage key, for example `index_sight`, so we  can know the value works for which module.

baurine: what it will happen if `localStorage.getItem('...')` is 'false'. or you should use `JSON.parse(localStorage.getItem('...') || 'false')`?

YiniXu9506: Good catch! 👍

baurine (approved):

YiniXu9506:
status/can-merge

Issue #1515, opened 2023-04-20
Dashboard throws 500 for slow query detail page with invalid utf-8 error


AndreMouche:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

**What did you do?**
In TiDB, I can set an invalid utf-8 string to a binary column successfully.
![img_v2_c10f98b3-451a-4303-b4d7-9c01c04851fg](https://user-images.githubusercontent.com/1085697/233307777-ea5e1900-fdbd-447d-9e6e-7dc34294b7fa.jpg)

However when the above log shows in slow-log,  I try to find more details for this slow log on dashboard 

**What did you expect to see?**

the details for the slow log

**What did you see instead?**

500 
and see the error log ```tipb.ExplainOperator.operator_info contains invalid UTF-8```
details:
```
[...]

PR #1518, opened 2023-04-27
slowquery: fix 500 when parsing binary plan fails

mornyx:
close: #1515 

# What's changed

Fix the issue where parsing the binary plan results in an error and return 500 in such cases.


shhdgit (approved):

mornyx:
Also fixed in 7.1.x by https://github.com/tikv/pd/pull/6416

PR #1530, opened 2023-06-05
chore: remove slow query v2 pages

baurine:


AndreMouche:
Do we have plan to merge this PR? @breezewish

shhdgit (approved):

baurine:
> Do we have plan to merge this PR? @breezewish

This is only a test PR. 

One user reported that his Chrome browse says fails to load `web worker`, so the dashboard doesn't work normally. After investigating, I found only the `@antv/g2plot` npm package which is used in the slow query v2 page would create `web worker`, so I removed this part of code for him to test, but not received replies yet.

baurine:
The user confirms that this PR works, so I will merge this PR.

PR #1569, opened 2023-08-13
feat: allows ngm timeout to be configured

mornyx:
This PR allows us to configure ngm timeout, so if the user finds that the query is slow and 502 occurs, the timeout can be appropriately increased.

<del>The PR also updates `pnpm-lock.yaml` (just by running `make package`), so the number of code lines doesn't look real..</del>

NOTE: This PR needs to be cherry picked to v6.1.


baurine:
> The PR also updates pnpm-lock.yaml (just by running make package), so the number of code lines doesn't look real..

I think you can revert the `pnpm-lock.yaml` changes when you commit the code.

baurine (approved):

PR #1579, opened 2023-08-21
keyvisual: use scanRegions instead of fetch all regions

nolouch:
PD memory:
before
image

after
image

test with new apis:
image


pkg/keyvisual/input/api.go
```diff
@@ -110,10 +114,55 @@ func read(data []byte) (*RegionsInfo, error) {
...
+}
+
+func scanRegions(pdclient *pd.Client, key, endKey string, limit int) (*RegionsInfo, error) {
+	values := url.Values{
+		"key":     {key},
+		"end_key": {endKey},
+		"limit":   {fmt.Sprintf("%d", limit)},
+	}
+
+	url := "/regions/key" + "?" + values.Encode()
```
baurine, line 155: Does this api work in the lower PD versions, likes 6.1 ?

nolouch: yes, this API exists for a long time, I believe it should be accessible on versions 4.x, and above.

mornyx (approved):

baurine (approved):

PR #1581, opened 2023-08-23
feat(conprof): show conprof download link for TiCDC

baurine:
## What Did

This PR supports to view TiCDC conprof status.

~~TODO: test compatibility with old versions (7.1, 6.5. 6.1)~~

## Test

Test in 7.3.0, 7.1.0, 6.5.3, 6.1.0.

6.1.0 doesn't show TiCDC component in conprof, because the ngm of this version doesn't support TiCDC.

others all show TiCDC component in conprof.

## Preview

Run dev env:

```
$ tiup playground 
$ tiup cdc server --addr 127.0.0.1:8301
```

The cluster and cdc are both 7.3.0 version.

image

image

**Note:**  in my test environment, the ngm can't get the TiCDC profiles successfully, don't know why, so in the above screenshot, the status is `Partial Finished`, and in the detail page, the status for `TiCDC` is `Error`, clicking the `Error Information`, it will show the error reason in the modal.

### DIFF
--- a/pkg/apiserver/conprof/service.go
+++ b/pkg/apiserver/conprof/service.go
@@ -144,6 +144,7 @@ type ComponentNum struct {
 	PD      int `json:"pd"`
 	TiKV    int `json:"tikv"`
 	TiFlash int `json:"tiflash"`
+	TiCDC   int `json:"ticdc"`
 }
 
 type GroupProfiles struct {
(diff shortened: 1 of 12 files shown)


shhdgit:
LGTM, but why the second pic shows the connection refused? What does that mean?

baurine:
> LGTM, but why the second pic shows the connection refused? What does that mean?

It's the ngm service error, I don't know why it doesn't work normally, need @mornyx 's help. But just for frontend, it shows the correct content from the API.

mornyx:
I successfully deployed on the feature branch and no errors were repeated. I think we can assume that this feature is OK, the problem is the environment.

image
image
image

So LGTM

mornyx (approved):

PR #1583, opened 2023-08-29
feat(security): encrypt login password

baurine:
## What Did

Encrypt the SQL Auth password in the frontend side by the RSA public key, and decrypt it in the backend side.

## Preview

image

image


mornyx:
Backend LGTM. BTW, Should we provide a configuration item for this feature (maybe disabled by default), rather than as the only choice?

ui/packages/tidb-dashboard-for-op/src/dashboardApp/layout/signin/index.tsx
```diff
@@ -517,6 +540,7 @@ function App({ registry }) {
             <TiDBSignInForm
               successRoute={successRoute}
               onClickAlternative={handleClickAlternative}
+              publicKey={publicKey}
```
shhdgit, line 543: Why not pass `publicKey={loginInfo.sql_auth_public_key}` directly?

shhdgit (approved):
LGTM.

baurine:
> Backend LGTM. BTW, Should we provide a configuration item for this feature (maybe disabled by default), rather than as the only choice?

After we can encrypt the login password, I don't think the user want to see the plain password again.

PR #1584, opened 2023-08-30
feat(debug-api): support pagination for ddl history

baurine:
## What Did

This PR supports to download the ddl history by pagination.

## Preview

image

### DIFF
--- a/ui/packages/tidb-dashboard-lib/src/apps/DebugAPI/apilist/ApiForm.tsx
+++ b/ui/packages/tidb-dashboard-lib/src/apps/DebugAPI/apilist/ApiForm.tsx
@@ -47,8 +47,6 @@ export default function ApiForm({
 
   const download = useCallback(
     async (values: any) => {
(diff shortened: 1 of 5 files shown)


ui/packages/tidb-dashboard-lib/src/apps/DebugAPI/apilist/ApiForm.tsx
```diff
@@ -47,14 +47,23 @@ export default function ApiForm({
 
   const download = useCallback(
     async (values: any) => {
+      console.log('values:', values)
+      console.log('params:', params)
```
baurine, line 51: Remove it.

shhdgit (approved):
So we don't need to provide `DDL History - All` anymore?

baurine:
> So we don't need to provide `DDL History - All` anymore?

Yep, I think so, according the discussion. And if they really want to download the all, they can set a very large `limit` value.

PR #1587, opened 2023-09-08
test(login): fix login integration test

baurine:


pkg/apiserver/user/auth.go
```diff
@@ -122,6 +122,15 @@ func NewAuthService(featureFlags *featureflag.Registry) *AuthService {
...
 			}
+			if form.Type == 0 {
+				// generate new rsa key pair for each sql auth login
+				privateKey, publicKey, err := GenerateKey()
+				// if generate successfully, replace the old key pair
+				if err == nil {
+					service.RsaPrivateKey = privateKey
+					service.RsaPublicKey = publicKey
+				}
+			}
```
shhdgit, line 133: Will there be issues during concurrency?

baurine: Yep, this is a potential issue, but for tidb-dashboard use case, it's a very low possibility. If the user meet and think it's a question, we can consider to improve it later.

baurine: Hold this PR and think it clearer.

shhdgit (approved):

pkg/apiserver/user/auth.go
baurine: Revert the key logic and change this PR to `fix login integration test`.

PR #1593, opened 2023-09-15
hide full stacktrace info in error message

mornyx:
We have been returning the full stack traces to the frontend, however this is not necessary, especially since they may reveal some internal information.


breezewish (approved):

baurine (commented):
LGTM, but need to fix the unit test

PR #1599, opened 2023-10-10
chore(sso): improve client secret security

baurine:
Related PR: https://github.com/pingcap/tidb-dashboard/pull/1567, it adds support for "Client Secret Post" authentication method. (When you create a `regular web app` in auth0 console, it uses the "Client Secret Post" authentication method).

This PR handle the compatibility with old version, and hide the client secret for security.

### Test

1. Login [Auth0](https://auth0.com/) by google account
2. Create a "Regular Web App" and "Single Web App" applications in the Auth0
image

4. Set the `Allowed Callback URLs` and `Allowed Logout URLs` for the above applications, following the steps in the https://docs.pingcap.com/tidb/stable/dashboard-session-sso#example-2-use-auth0-for-tidb-dashboard-sso-sign-in
image

6. In tidb-dashboard, go to sso configuration page, fill the `OIDC Client ID`. If the `Client ID` comes from auth0 single web app, we don't need to fill `OIDC Client Secret`. If the `Client ID` comes from auth0 regular web app, we need to fill the `OIDC Client Secret` as well.
7. Logout and Login via SSO.

### Demo

https://github.com/pingcap/tidb-dashboard/assets/1284531/cd6a658e-8c53-466c-98fc-d3cb448e63d1

### DIFF
--- a/ui/packages/tidb-dashboard-lib/src/apps/UserProfile/components/Form.SSO.tsx
+++ b/ui/packages/tidb-dashboard-lib/src/apps/UserProfile/components/Form.SSO.tsx
@@ -252,44 +252,47 @@ export function SSOForm() {
                   label={t('user_profile.sso.form.client_id')}
                   rules={[{ required: true }]}
                 >
-                  <Input
-                    width={400}
-                    disabled={!isWriteable}
-                    style={DEFAULT_FORM_ITEM_STYLE}
-                  />
+                  <Input disabled={!isWriteable} style={{ width: 320 }} />
                 </Form.Item>
-                <Form.Item
-                  name="client_secret"
-                  label={t('user_profile.sso.form.client_secret')}
-                  rules={[{ required: false }]}
-                >
-                  <Input
-                    width={400}
-                    type="password"
-                    disabled={!isWriteable}
-                    style={DEFAULT_FORM_ITEM_STYLE}
-                  />
-                </Form.Item>
-                {/* <Form.Item
-                  name="scopes"
-                  label={t('user_profile.sso.form.scopes')}
-                  rules={[{ required: false }]}
-                >
-                  <Input
-                    disabled={!isWriteable}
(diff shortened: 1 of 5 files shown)


ui/packages/tidb-dashboard-lib/src/apps/UserProfile/components/Form.SSO.tsx
```diff
@@ -252,40 +252,47 @@ export function SSOForm() {
...
+                    label={t('user_profile.sso.form.scopes')}
+                    rules={[{ required: false }]}
+                  >
+                    <Input
+                      disabled={!isWriteable}
+                      style={{width: 320}}
+                      placeholder="openid profile email"
+                    />
+                  </Form.Item> */
+                }
```
baurine, line 287: hi @horus , in your case, do you think you really need to customize `scopes` field?

mornyx (approved):

shhdgit (approved):

ui/packages/tidb-dashboard-lib/src/apps/UserProfile/components/Form.SSO.tsx
horus: Yes, otherwise I wouldn't bother.

Talking about scopes, `profile` and `email` are all [optional](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims), mind you. The original source got them hardcoded, since they're well-supported I don't think there'll be serious problems.

In some(that is, my) situation, one could use the scopes to implement access control, as I've already stated in the previous PR. These are fairly common and straight-forward, I let them there for the flexibility and/or future-proof.

Hope this clarifies.

baurine: Ok, got it, I will update it later.

PR #1608, opened 2023-10-31
tweak: explicitly set session max_execution_time to 600s

mornyx:
close: https://github.com/pingcap/tidb-dashboard/issues/1604

This PR explicitly sets the session-level `max_execution_time` to avoid being affected by the global-level `max_execution_time`.

### DIFF
--- a/pkg/tidb/client.go
+++ b/pkg/tidb/client.go
@@ -33,7 +33,8 @@ var (
 )
 
 const (
-	defaultTiDBStatusAPITimeout = time.Second * 10
+	defaultTiDBStatusAPITimeout      = time.Second * 10
+	defaultTiDBSQLExecutionTimeoutMs = 600000 // 600s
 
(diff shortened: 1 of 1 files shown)


pkg/tidb/client.go
```diff
@@ -33,7 +33,8 @@ var (
 )
 
 const (
-	defaultTiDBStatusAPITimeout = time.Second * 10
+	defaultTiDBStatusAPITimeout      = time.Second * 10
+	defaultTiDBSQLExecutionTimeoutMs = 600000 // 600s
```
jackysp, line 37: It is 600s on this line, but the PR title is 60s.

jackysp (approved):
LGTM

pkg/tidb/client.go
mornyx: Sry, it's a PR title's typo, should be 600s.

baurine (approved):

PR #1614, opened 2023-11-25
profiling: use debug=1 for goroutine profile to avoid STW

djshow832:
Refer to https://github.com/pingcap/tidb/issues/48695. With `debug=2`, there are some huge STWs, which cause latency spikes in TiDB and PD.


mornyx:
LGTM

/cc @XuHuaiyu PTAL, we'll revert the change introduced by https://github.com/pingcap/tidb-dashboard/issues/1401, so the goroutine profile with stacktrace is no longer available.

mornyx (approved):

PR #1622, opened 2023-12-06
add ticdc topology in apiserver

CharlesCheung96:
Manual test:
image


pkg/utils/topology/tidb.go
```diff
@@ -27,7 +27,7 @@ func FetchTiDBTopology(ctx context.Context, etcdClient *clientv3.Client) ([]TiDB
 
 	resp, err := etcdClient.Get(ctx2, tidbTopologyKeyPrefix, clientv3.WithPrefix())
 	if err != nil {
-		return nil, ErrEtcdRequestFailed.Wrap(err, "failed to get key %s from %s etcd", tidbTopologyKeyPrefix, distro.R().PD)
```
baurine, line 30: I think here should be the `distro.R().PD`, because it means where the etcd belongs.

baurine:
Please help fix the CI error as well.

mornyx (approved):

PR #1632, opened 2023-12-27
add more deploy info to ticdc

CharlesCheung96:
ref https://github.com/pingcap/tiflow/issues/10375

### DIFF
--- a/pkg/utils/topology/ticdc.go
+++ b/pkg/utils/topology/ticdc.go
@@ -70,9 +70,12 @@ func FetchTiCDCTopology(ctx context.Context, etcdClient *clientv3.Client) ([]TiC
 
 func parseTiCDCInfo(clusterName string, value []byte) (*TiCDCInfo, error) {
 	ds := struct {
-		ID      string `json:"id"`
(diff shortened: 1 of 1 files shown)


pkg/utils/topology/ticdc.go
```diff
@@ -85,11 +88,14 @@ func parseTiCDCInfo(clusterName string, value []byte) (*TiCDCInfo, error) {
...
-		Status:      ComponentStatusUp,
-		StatusPort:  port,
+		ClusterName:    clusterName,
+		GitHash:        ds.GitHash,
+		Version:        ds.Version,
+		IP:             hostname,
+		Port:           port,
+		DeployPath:     ds.DeployPath,
+		Status:         ComponentStatusUp,
+		StatusPort:     port,
```
mornyx, line 98: StatusPort is same to Port?

CharlesCheung96: Yes, TiCDC uses a single port for external services.

baurine (approved):

PR #1646, opened 2024-02-01
feat(slow-query): support slow query search by resource group

nolouch:
support slow query search by resource group
- If there are above 2 resource groups in the cluster, will show the `MultiSelect`
- Add relative API

Example:
image

### DIFF
--- a/ui/packages/tidb-dashboard-for-clinic-op/src/apps/SlowQuery/context.ts
+++ b/ui/packages/tidb-dashboard-for-clinic-op/src/apps/SlowQuery/context.ts
@@ -30,6 +30,17 @@ class DataSource implements ISlowQueryDataSource {
     } as any)
   }
 
+  infoListResourceGroupNames(options?: ReqConfig) {
+    // return Promise.reject(new Error('no need to implemented'))
+    return Promise.resolve({
+      data: [],
+      status: 200,
+      statusText: 'ok',
+      headers: {},
+      config: {}
+    } as any)
+  }
+
   slowQueryAvailableFieldsGet(options?: ReqConfig) {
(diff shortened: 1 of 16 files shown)


nolouch:
PTAL @baurine

baurine:
Get, in reviewing.

ui/packages/tidb-dashboard-for-clinic-op/src/apps/SlowQuery/context.ts
```diff
@@ -119,6 +131,7 @@ export const ctx: (extra: DsExtra) => ISlowQueryContext = (extra) => ({
     apiPathBase: client.getBasePath(),
     enableExport: false,
     showDBFilter: false,
-    showDigestFilter: false
+    showDigestFilter: false,
+    showResourceGroupFilter: true
```
baurine, line 135: For clinic-op, I feel `showResourceGroupFilter` is better to `false`, although in this case, `true` and `false` are the same result.

baurine (approved):

ui/packages/tidb-dashboard-for-clinic-op/src/apps/SlowQuery/context.ts
nolouch: What do clinic-op and clinic-could stand for? dedicated and serverless?

baurine: clinic-op for the OP clusters, they can collect the diagnosis data by tiup diag tool in local and upload to clinic server, while clinic-cloud for the cloud clusters.

PR #1649, opened 2024-02-28
fix: correctly display the disk mount points of tiflash

mornyx:
close: https://github.com/pingcap/tidb-dashboard/issues/1625

In the past, the data directory of tiflash was only obtained from `engine-store.path`. This PR modified it to be obtained from both `engine-store.path`, `engine-store.storage.main.dir` and `engine-store.storage.latest.dir`.

### DIFF
--- a/pkg/apiserver/clusterinfo/hostinfo/cluster_config.go
+++ b/pkg/apiserver/clusterinfo/hostinfo/cluster_config.go
@@ -3,6 +3,7 @@
 package hostinfo
 
 import (
+	"encoding/json"
 	"strings"
 
 	"gorm.io/gorm"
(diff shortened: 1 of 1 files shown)


pkg/apiserver/clusterinfo/hostinfo/cluster_config.go
```diff
@@ -37,9 +40,44 @@ func FillInstances(db *gorm.DB, m InfoMap) error {
...
+			switch row.Key {
+			case "engine-store.path":
+				items := strings.Split(row.Value, ",")
+				for _, path := range items {
+					paths = append(paths, strings.TrimSpace(path))
+				}
+			case "engine-store.storage.main.dir", "engine-store.storage.latest.dir":
+				if err := json.Unmarshal([]byte(row.Value), &paths); err != nil {
+					return err
+				}
```
JaySon-Huang, line 65: "engine-store.storage.main.dir", "engine-store.storage.latest.dir" These two configs can be set simultaneously. In this case, we need to merge the two array to `paths`

pkg/apiserver/clusterinfo/hostinfo/cluster_config.go
```diff
@@ -37,9 +40,44 @@ func FillInstances(db *gorm.DB, m InfoMap) error {
...
+			default:
+				paths = []string{row.Value}
+			}
+			for _, path := range paths {
+				mountDir := locateInstanceMountPartition(path, m[hostname].Partitions)
+				if mountDir != "" {
+					m[hostname].Instances[row.Instance].PartitionPathL = strings.ToLower(mountDir)
+					break
```
JaySon-Huang, line 73: If a TiFlash instance is deployed on multiple disks (MountPartition), can we show all mount point usage of that TiFlash instance?

baurine (approved):

PR #1661, opened 2024-03-25
chore(top_slowuqery): address feedback

baurine:
What this PR did:

1. change line chart to bar chart
2. add order by filter
3. add extra style for the table sorted column header to make it more explicit

image


shhdgit (approved):
lgtm

mornyx (approved):

PR #1672, opened 2024-04-25
fix(slow_query): fix download slowlog api, add range preview and timezone reminder

elleryasimov:
- Fix slowquery log download api.
- Add reminder for UTC-0
- Add time range preview for slow log to download.


shhdgit (approved):
lgtm.

PR #1682, opened 2024-05-23
fix(top-slowquery): fix db filter sometimes can not select item

baurine:
This PR fixes the bug that db filter can not select item after changing the time window, and clean up some codes.


shhdgit (approved):
lgtm.

PR #1695, opened 2024-06-21
fix(slow-query): fix cannot show RU in slow query detail page

nolouch:
fix https://github.com/pingcap/tidb-dashboard/issues/1693

Test:

image


baurine (approved):

nolouch:
why is the cla always not reported?

PR #1717, opened 2024-08-23
*: add share code revoke api; add a config to allow disabling custom prom addr 

mornyx:
- **add share code revoke api**: In the past, we could not revoke the authorization code obtained through "Share TiDB Dashboard Sessions" and could only wait for it to expire. I added a new API to revoke all authorization codes by resetting the key.
- **add a config to allow disabling custom prom addr**: For security reasons, we need to introduce a configuration item to disable the custom promtheus address to avoid SSRF attacks.


baurine (approved):
lgtm

PR #1737, opened 2024-09-24
fix: revert limit pnpm version

baurine:


wuhuizuo:
/hold
wait for build testing

PR #1792, opened 2025-06-11
utils: add support for keyspace prefix when fetching TiDB instances

zimulala:
close https://github.com/pingcap/tidb-dashboard/issues/1790

When we get the normal TiDB instance count of 0(fetch from tidbTopologyKeyPrefix), then let's check if there are any alive instances obtained from the path prefixed with keyspaceNameKeyPrefix.

In addition, we try our best not to affect the clusters without keyspace information

E2E manual test (we still need to update the pd dependency):

- Log in to Dashboard successfully

- In addition, the following are the situations where additional logs are added:

    截屏2025-06-11 14 09 14
    KeyspaceID 1: tidb 4000 (normal)

    KeyspaceID 101: tidb 4001 (normal)

    KeyspaceID 2: tidb 4000 (only set kv to etcd:`ETCDCTL_API=3 etcdctl --endpoints=127.0.0.1:2379 put /keyspaces/tidb/2/topology/tidb/127.0.0.1:4000/ttl 1749619996349209123` )

### DIFF
--- a/pkg/utils/topology/tidb.go
+++ b/pkg/utils/topology/tidb.go
@@ -81,7 +81,6 @@ func getAliveNodesAndInfos(ctx context.Context, etcdClient *clientv3.Client, key
 }
 
 func getAliveNodesAndInfoWithPrefix(ctx context.Context, etcdClient *clientv3.Client) (map[string]struct{}, map[string]*TiDBInfo, error) {
-	log.Warn("xxx ------------------- next-gen 00")
 	childCtx, cancel := context.WithTimeout(ctx, defaultFetchTimeout)
 	defer cancel()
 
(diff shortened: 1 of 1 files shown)


zimulala:
PTAL @mornyx

zimulala:
PTAL @baurine @breezewish @crazycs520

breezewish (approved):
Looks good. One thing I concern is that, as a serverless (keyspaced) service, the positioning of TiDB Dashboard needs to be considered -- Whether it is targeted at DBaaS Maintainers (cross Keyspace), or it is targeted at Cluster Users (within a Keyspace).

zimulala:
> Looks good. One thing I concern is that, as a serverless (keyspaced) service, the positioning of TiDB Dashboard needs to be considered -- Whether it is targeted at DBaaS Maintainers (cross Keyspace), or it is targeted at Cluster Users (within a Keyspace).

@breezewish 
This is mainly a temporary solution for next-gen. Currently, the serverless service does not use tidb-dashboard, so it won't affect the serverless offering.

PR #1875, opened 2026-03-18
Run go fix

dveeden:
This is the result of `go fix ./...`, but with the `interface{}` to `any` changes reverted.

With one extra fix for this:
```
util/jsonserde/convert.go:18:2: Consider pre-allocating `out` (prealloc)
	var out []rune
	^
```

~~And for swaggo an update is needed to support `any` which replaces `interface{}`~~


yibin87 (approved):
LGTM

PR #1877, opened 2026-03-18
Upgrade golangci-lint

dveeden:


yibin87 (commented):
One concern here: can we avoid disabling G705 globally? This PR upgrades golangci-lint, but a repo-wide exclusion will also hide future real XSS-taint findings. If this is only to address a few false positives from the new version, can we scope the exclusion to those files/paths instead of turning the rule off everywhere?

dveeden:
I have limited the G705 exclusion to _test.go files. I've also included the `misspell` linter.

PR #1916, opened 2026-07-29
test: fix slow query session attrs compatibility

dillon-zheng:
## Summary
- make slow query integration assertions require session_connect_attrs only when the running TiDB schema exposes it
- keep the all-version mock slow query tests compatible with TiDB v6.6.0

## Verification
- go test ./pkg/apiserver/slowquery -count=1
- go test ./tests/integration/slowquery -run '^$' -count=1

### DIFF
--- a/tests/integration/slowquery/mock_db_test.go
+++ b/tests/integration/slowquery/mock_db_test.go
@@ -4,6 +4,7 @@ package slowquery
 
 import (
 	"fmt"
+	"slices"
 	"testing"
 
 	"github.com/stretchr/testify/suite"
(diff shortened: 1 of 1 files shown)


jiong-nba:
I reopened #1919 with the same TiDB v6.6 session_connect_attrs compatibility fix plus the repair for this PR's current nightly failure. The nightly schema now exposes IA_remote_read_segment_size and IA_remote_read_segment_wait_time, but the mock CLUSTER_SLOW_QUERY schema here lacks those columns, which causes Error 1054 before the assertions complete. #1919 adds the columns and fixture values, keeps strict conditional assertions for both old and new schemas, and passes the slowquery integration package against real TiDB v6.6.0 and nightly v9.0.0. Maintainers can use #1919 as the combined fix.

PR #355, opened 2020-04-15
ui: Remove monorepo

breeswish:
This PR removes the monorepo to make life easier.

Fix https://github.com/pingcap-incubator/tidb-dashboard/issues/324
Fix https://github.com/pingcap-incubator/tidb-dashboard/issues/280
Fix https://github.com/pingcap-incubator/tidb-dashboard/issues/269
Fix https://github.com/pingcap-incubator/tidb-dashboard/issues/170
Fix https://github.com/pingcap-incubator/tidb-dashboard/issues/75

Also fix:

- Some styles and scripts were imported / executed multiple times
- Some package stylesheets did not have Dashboard global theme color
- Could not use `@import '~antd/es/...` in packages
- Could not import `@umijs/hooks` in packages
- KeyViz produced several lint warnings

PR #1917, opened 2026-08-07
topsql: support detailed TiKV IO dimensions

jiong-nba:
## What changed

- Expose logical_read_bytes, logical_write_bytes, and rocksdb_block_read_count in TopSQL summary responses and all list, chart, table, and plan aggregations.
- Add logical_read, logical_write, and block_read ordering options.
- Extend the TiKV I/O collection API and OP settings UI to inspect and update the optional resource-metering.enable-detailed-io-collection setting.
- Report detailed_io_enabled: true only when every TiKV node responds, contains the detailed I/O setting, and has it enabled; otherwise uniformly report false.
- Preserve the legacy logical_io field, ordering option, and existing URLs when detailed I/O collection is unavailable.
- Add Clinic Cloud context support so the embedding application can explicitly enable detailed I/O dimensions for supported deployments.

## UI behavior

- The TopSQL switch is the root control.
- Network I/O Collection is shown only when TopSQL is enabled.
- Detailed I/O Collection is shown only when Network I/O Collection is enabled.
- Disabling a parent switch hides its dependent controls and keeps the legacy behavior.
- The three detailed I/O dimensions are shown only when Network I/O Collection is enabled and detailed_io_enabled is true.

## Compatibility

- Existing clients can continue posting only enable; detailed_io_enabled remains optional.
- No separate support or multi-value fields are exposed for detailed I/O collection.
- Unsupported, disabled, mixed-version, partially enabled, and partially unreachable states uniformly degrade to detailed_io_enabled: false.
- The settings UI submits detailed_io_enabled only after the user changes that switch, so saving unrelated settings does not send the new key to older TiKV versions.
- The required enable field is validated explicitly in update requests and reflected in the generated TypeScript contract.

## Related work

- Design: https://pingcap.feishu.cn/wiki/G0SMwPsWxiPZbmk6VNVcEEp0njc
- Tracking issue: https://github.com/tikv/tikv/issues/19867
- TiKV implementation: https://github.com/tikv/tikv/pull/19953
- Clinic Dashboard reference: https://github.com/pingcap/tidb-dashboard/pull/1860
- OP reference: https://github.com/pingcap/tidb-dashboard/pull/1868

## Validation

- go test ./pkg/apiserver/topsql
- SKIP_GEN_API=1 corepack pnpm --filter @pingcap/tidb-dashboard-client build
- corepack pnpm --filter @pingcap/tidb-dashboard-lib build
- corepack pnpm --filter @pingcap/tidb-dashboard-for-op build
- corepack pnpm --filter @pingcap/tidb-dashboard-for-clinic-cloud build
- Prettier and staged ESLint checks

The Swagger specification was regenerated. TypeScript API models were synchronized from that specification and compiled successfully; full OpenAPI client regeneration was not available locally because Java was not installed.

## CI note

Docker image, backend/frontend build, and backend unit-test checks pass. The current Backend - Integration - TiDB@^6.0 job fails during make integration_test after the integration test process exits with code 2; this appears isolated to the TiDB 6.x integration environment and should be re-run for confirmation.

## Summary by CodeRabbit

* **New Features**
  * Added Top SQL sorting, charts, tables, and details for logical read bytes, logical write bytes, and RocksDB block reads.
  * Added optional detailed TiKV network I/O collection controls, including capability status and partial-state warnings.
  * Improved configuration handling across nodes, with clearer status reporting for unavailable or mixed settings.
  * Added session connection attributes to slow-query information.
* **Bug Fixes**
  * Incompatible detailed I/O selections now automatically fall back to supported metrics.


jiong-nba:
## End-to-end verification

Verified locally on August 17, 2026 with TiUP, [ng-monitoring PR #355](https://github.com/pingcap/ng-monitoring/pull/355) (`d295da59`), [TiDB Dashboard PR #1917](https://github.com/pingcap/tidb-dashboard/pull/1917) (`a70560af` source), and [TiKV PR #19953](https://github.com/tikv/tikv/pull/19953) commit `d14cd3194`.

> **TiKV HEAD blocker:** PR #19953 HEAD `00a1af385` changes kvproto to merged commit `af1ad1be`, which also introduces `ApiVersion::V3`. The current TiKV `api_version` crate has seven non-exhaustive matches and cannot build at that HEAD. The end-to-end run therefore uses `d14cd3194`, the latest buildable feature commit in the same PR, with its feature-branch kvproto dependency.

- TiKV configuration: `enable-network-io-collection: true` and `enable-detailed-io-collection: true`
- ng-monitoring health check: `{"health":true}`
- Sustained workload: six concurrent query/update patterns for five minutes, completing 4,382 statements
- The chart contains continuous per-second samples across the full five-minute window instead of a one-time spike
- Leading query attribution:
  - Logical Read: `45.85 GiB`
  - Logical Write: `241.94 MiB`
  - Read IOPS: `2.80M`
- Dashboard shows `Logical Read`, `Logical Write`, and `Read IOPS` independently. The legacy combined `Logical IO` option is hidden while detailed I/O is enabled.

### Settings and dimensions

Detailed I/O settings

Order By dimensions

### Sustained Logical Read

Sustained Logical Read workload

### Sustained Logical Write

Sustained Logical Write workload

### Sustained Read IOPS

Sustained Read IOPS workload

jiong-nba:
Dashboard OP settings screenshot. It shows the intended hierarchy: TopSQL -> Network I/O Collection -> Detailed I/O Collection. Detailed I/O is visible only when Network I/O Collection is enabled.
topsql-settings-1917

pkg/apiserver/topsql/service.go
```diff
@@ -206,15 +218,22 @@ func (s *Service) UpdateConfig(c *gin.Context) {
...
 type TikvNetworkIoCollectionConfig struct {
-	Enable       bool `json:"enable"`
-	IsMultiValue bool `json:"is_multi_value,omitempty"`
+	Enable            bool `json:"enable"`
+	IsMultiValue      bool `json:"is_multi_value,omitempty"`
+	DetailedIoEnabled bool `json:"detailed_io_enabled"`
+}
+
+type UpdateTikvNetworkIoCollectionRequest struct {
+	Enable            *bool `json:"enable" binding:"required"`
```
yibin87: `enable` used to be optional on this endpoint: an omitted value was decoded as `false` and the request proceeded. Making it `*bool` with `binding:"required"` changes that existing wire contract—`POST {}` now returns 400—and the regenerated Swagger client marks `enable` as required as well.

Since this endpoint is expected to remain backward compatible, could we keep accepting an omitted `enable` (preserving the previous `false` behavior) and leave only the newly introduced `detailed_io_enabled` as optional? If stricter validation is desired, it should be introduced in a versioned endpoint instead.

jiong-nba: Good point. Restored `enable` to a plain `bool` and removed the required binding, so omitted values continue to bind as `false` and `POST {}` remains backward compatible. `detailed_io_enabled` stays optional. I also updated the Swagger/generated TypeScript contracts and added binding regression coverage for both `{}` and a detailed-I/O-only request. Fixed in b202d79b.

yibin87: I revisited the existing Dashboard call paths and need to correct my earlier recommendation—sorry, I did not fully account for the endpoint’s implicit empty-request semantics.

The Dashboard always sends an explicit `enable` value when saving this setting, so making `enable` required does not break Dashboard compatibility. In contrast, preserving an optional plain `bool` makes `{}` (and, more importantly, a detailed-I/O-only request) indistinguishable from `{"enable": false}`, which can unexpectedly disable network I/O collection.

I now think the cleaner contract is to make `enable` required again, keep `detailed_io_enabled` optional, and reject requests that omit the parent switch. That narrows the legacy `{}` behavior, but avoids an ambiguous and potentially destructive update API.

jiong-nba: Agreed. The Dashboard save path always sends an explicit `enable`, while accepting a detailed-I/O-only request would silently turn the parent network I/O switch off. I restored `enable` as a required `*bool`, kept `detailed_io_enabled` optional, regenerated the Swagger/TypeScript contracts, and replaced the compatibility tests with validation coverage for omitted `enable`, explicit `false`, and the full request. Fixed in 44f05e91.

yibin87 (approved):
LGTM

PR #1919, opened 2026-08-18
slowquery: keep integration tests compatible across TiDB schemas

jiong-nba:
## What problem does this PR solve?

The slow-query integration tests currently break at both ends of the supported TiDB matrix:

- TiDB v6.6.0 does not expose `session_connect_attrs`, but the tests added by #1900 assumed that the field was always available.
- Current TiDB nightly exposes `IA_remote_read_segment_size` and `IA_remote_read_segment_wait_time`. `GetAvailableFields` therefore includes them in queries against the mock slow-query table, but the checked-in mock schema did not contain those columns, causing `Error 1054: Unknown column 'ia_remote_read_segment_size'`.

The second issue is the reason the nightly job still fails on #1916 after the TiUP mirror recovered.

## What is changed and how it works?

- Detect slow-query field availability with the existing `GetAvailableFields` helper.
- Validate the exact `session_connect_attrs` JSON when the running TiDB supports it, and require `nil` when it does not.
- Add the two IA remote-read columns and representative values to the mock schema and fixture.
- Validate IA values when the running TiDB exposes those fields, and require zero values on versions that do not.
- Add `Session_connect_attrs` to the detail fixture so the supported-version detail path remains covered.

This keeps the assertions strict while making the fixtures compatible with both TiDB v6.6.0 and newer/nightly schemas.

## Tests

- `go test ./pkg/apiserver/slowquery -count=1`
- `go test -race -v -count=1 ./tests/integration/slowquery` against TiDB v6.6.0
- `go test -race -v -count=1 ./tests/integration/slowquery` against TiDB nightly (v9.0.0)

Supersedes #1916 by including the same TiDB v6.6 compatibility fix plus the nightly IA fixture repair.

Related to #1900.
Unblocks #1917.

## Summary by CodeRabbit

* **Tests**
  * Improved slow-query validation across database versions with varying available columns.
  * Added coverage for remote-read metrics and session connection attributes.
  * Updated test fixtures and schema definitions to include the newly supported slow-query fields.


jiong-nba:
CI investigation update (August 18, 2026):

Both integration matrix jobs are failing before any Dashboard test is executed. The official TiUP `snapshot.json` currently references TiKV manifest version `7186`, but `https://tiup-mirrors.pingcap.com/7186.tikv.json` returns `404 Not Found` across multiple CDN edges.

This breaks both:

- `tiup playground ^6.0`
- `tiup playground nightly`

with the same error:

```text
Error: Playground bootstrapping failed: fetch /7186.tikv.json from mirror(https://tiup-mirrors.pingcap.com) failed: url https://tiup-mirrors.pingcap.com/7186.tikv.json: not found
```

The issue has been reported to TiUP maintainers: https://github.com/pingcap/tiup/issues/2728.

The slow-query assertions changed by this PR have therefore not run in the failed GitHub integration jobs. Once the mirror manifest is restored, these jobs need to be rerun.

jiong-nba:
Closing this PR as a duplicate of #1916, which addresses the same TiDB 6.6 compatibility issue in `tests/integration/slowquery/mock_db_test.go`.

One implementation difference is that this PR explicitly asserts `SessionConnectAttrs` is `nil` when `session_connect_attrs` is unavailable. That stricter assertion can be considered in #1916.

jiong-nba:
Reopening this PR because it is no longer only a duplicate of #1916. After the TiUP mirror recovered, #1916's nightly rerun reached the tests and exposed a separate schema-fixture mismatch: TiDB nightly selects the new IA remote-read fields, while the checked-in mock table does not contain them. This branch now includes both the TiDB v6.6 session_connect_attrs compatibility fix and the nightly IA fixture repair, with strict supported/unsupported assertions. The slowquery integration package passes against real TiDB v6.6.0 and nightly v9.0.0.

jiong-nba:
CI update for commit 19b50181: the fresh run is green. In particular, Backend - Integration - TiDB@^6.0 passed in 3m18s and Backend - Integration - TiDB@nightly passed in 3m27s. Backend, frontend, unit tests, and all four Docker image builds also passed. The only remaining pending status is tide, which is waiting for lgtm/approval labels.

XuHuaiyu (approved):

